PR #145, opened 2019-05-14
 Adds handling of the shutdown event

### LINKED ISSUE #114: Implement shutdown handling
## Description

When the sdk emits a shutdown event the application should handle it.

## Acceptance criteria
- 

## Tasks
- [ ]

kelsos:
Closes #114

### DIFF
--- a/raiden-dapp/src/services/raiden-service.ts
+++ b/raiden-dapp/src/services/raiden-service.ts
@@ -1,4 +1,4 @@
-import { Raiden, RaidenChannel } from 'raiden';
+import { Raiden, RaidenActionType, RaidenChannel } from 'raiden';
 import { Store } from 'vuex';
 import { RootState } from '@/types';
 import { Web3Provider } from '@/services/web3-provider';
@@ -52,9 +52,18 @@ export default class RaidenService {
         this.store.commit('account', await this.getAccount());
         this.store.commit('balance', await this.getBalance());
 
-        this.subscription = raiden.channels$.subscribe(value => {
-          this.store.commit('updateChannels', value);
+        this.subscription = raiden.events$.subscribe(value => {
+          if (value.type === RaidenActionType.SHUTDOWN) {
+            this.store.commit('reset');
+            this.subscription!.unsubscribe();
+          }
         });
+
+        this.subscription.add(
+          raiden.channels$.subscribe(value => {
+            this.store.commit('updateChannels', value);
+          })
+        );
 
         this.store.commit('network', raiden.network);
       }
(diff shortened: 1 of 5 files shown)


raiden-dapp/src/services/raiden-service.ts
```diff
@@ -52,10 +52,20 @@ export default class RaidenService {
         this.store.commit('account', await this.getAccount());
         this.store.commit('balance', await this.getBalance());
 
-        this.subscription = raiden.channels$.subscribe(value => {
-          this.store.commit('updateChannels', value);
+        this.subscription = raiden.events$.subscribe(value => {
+          if (value.type === RaidenActionType.SHUTDOWN) {
+            this.store.commit('reset');
+            raiden.stop();
```
andrevmatos: No need to `stop` after `RaidenActionType.SHUTDOWN`, as the former is also a shutdown event, and everything that happens after shutdown is ignored

kelsos: That is good to know thank you :)

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -52,10 +52,19 @@ export default class RaidenService {
         this.store.commit('account', await this.getAccount());
         this.store.commit('balance', await this.getBalance());
 
-        this.subscription = raiden.channels$.subscribe(value => {
-          this.store.commit('updateChannels', value);
+        this.subscription = raiden.events$.subscribe(value => {
```
andrevmatos: You shouldn't need to unsubscribe manually, the `events$` stream completes upon shutdown, and therefore all subscriptions are dropped automatically. Also, I think it'd be more elegant to use a filter instead of a if, something like:
```
raiden.events$
  .pipe(filter(value => value.type === RaidenActionType.SHUTDOWN))
  .subscribe(() => this.store.commit('reset'));
```

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -52,10 +52,19 @@ export default class RaidenService {
...
+          if (value.type === RaidenActionType.SHUTDOWN) {
+            this.store.commit('reset');
+            this.subscription!.unsubscribe();
+          }
         });
 
+        this.subscription.add(
```
andrevmatos: `channels$` also complete upon shutdown

raiden-dapp/src/services/raiden-service.ts
kelsos: Ok it makes sense. I will go an remove them and then merge the PR.

PR #164, opened 2019-05-27
Implement new channel deposit design

### LINKED ISSUE #126: Design and implement channel deposit
## Description
Design and implement the channel deposit

## Acceptance criteria
- It's in the channel list
- Should be a window overlay
- Token name should be visible (on top of overlay)
- Needs to have a 'Cancel' and 'Confirm' button

## Tasks
- [x] Design
- [x] Implementation

kelsos:
- Removes Deposit dialog
- Replaces Deposit dialog with inline component
- Changes layout for Amount input and makes it work on both light/dark background 

Closes #126


kelsos:
Yes, probably the dark one is not needed since it is already applied to the whole app. I will remove it and see if that works or breaks something. If it doesn't break anything I will push a fix on top and merge.

kelsos:
@andrevmatos I also named my classes light and dark and my custom property light in the ChannelDeposit.vue component. Do you think that this is confusing? or overlapping somehow with vuetify's own properties?

PR #170, opened 2019-06-04
Matrix room create, invite, be invited, leave, clean, send, receive messages

### LINKED ISSUE #91: I want to be able to check, if a specific node receives a message (incl room creation)
## User Story

As Chris, I want to be able to check, if a specific node receives a message so that I can decide, if I want to send a payment or not.

## Acceptance criteria
- A private matrix room has been created for the channel
- I am able to send a valid message (can be hardcoded)
- I receive a delivered message
- Feature does not need to be designed. It's purpose is to have a first spike solution for the sending and receiving messages.
- No API

## Tasks
- [x] Create private matrix room with partners
- [x] Send a message
- [x] Ensure, that a single room is created (Concurrency issue)
- [x] Receive a message
- [ ] Add tests

andrevmatos:
- [x] `matrix.startClient` is called lazily, to give time for callback hooks to be in place
- [x] When sendMessage to an address we don't know how to talk, create a private room and invite them
- [x] Even if user is offline or roam of server, they can be re-invited to this room
- [x] If it happens they invite us instead, join and start talking on their room
- [x] If there's more than 3 rooms for a user, leave the >3 ones
- [x] Leave/cleanup rooms we don't know about (global/discovery, peers)
- [x] Clean a room if somehow we're kicked out of it
- [x] Send message: when required to send a message, wait for a valid user presence for address, a room to be created for them, they joining and being online, and only then send (once, for now)
- [x] On any relevant room (with users of interest, which we started monitoring), listen for messages and output actions for them
- [x] If we receive a message for an user on a room is not our first option, it means they want to talk on that room, so put it in front of the queue and start sending messages on it
- [ ] ~Wrap JSON messages and `message_id`s~ (on next PR)
- [ ] ~Retry JSON messages until some condition, e.g. `Delivered`~ (on next PR)
- [x] Tests

Fix #91

### DIFF
--- a/raiden/src/store/epics/matrix.ts
+++ b/raiden/src/store/epics/matrix.ts
@@ -1,22 +1,28 @@
 import { ofType } from 'redux-observable';
-import { Observable, from, of, EMPTY, fromEvent } from 'rxjs';
+import { Observable, combineLatest, from, of, EMPTY, fromEvent, timer, ReplaySubject } from 'rxjs';
 import {
   catchError,
+  concatMap,
+  distinctUntilChanged,
   filter,
+  groupBy,
   ignoreElements,
   map,
   mergeMap,
+  multicast,
   withLatestFrom,
   scan,
   startWith,
   switchMap,
+  take,
+  takeUntil,
   tap,
   toArray,
 } from 'rxjs/operators';
-import { minBy } from 'lodash';
+import { get, find, minBy } from 'lodash';
 
 import { getAddress, verifyMessage } from 'ethers/utils';
-import { MatrixEvent, User } from 'matrix-js-sdk';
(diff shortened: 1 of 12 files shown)


raiden/src/store/epics/matrix.ts
```diff
@@ -263,3 +305,416 @@ export const matrixPresenceUpdateEpic = (
...
+          if (!member) return from(matrix.invite(roomId, action.userId));
+        }
+      }
+      return EMPTY;
+    }),
+    ignoreElements(),
+  );
+
+/**
+ * Handle invites sent to us and accepts them iff sent by a monitored user
```
kelsos: nitpick: iff/if

andrevmatos: @kelsos `iff` is an abbreviation for `if and only if`, more commonly used on math, but also valid on algorithms context, like this one

Issue #143, opened 2019-05-14
Architecture of token transfers in sdk


christianbrb:
## Description

Before starting to implement the token transfers we should design the architecture

## Acceptance criteria
- Not mediating transfers
- No refunds off-chain
- Description how to create a balance proof
- Describe how to populate the pending locks merkle tree
- Description how a balance proof is signed
- The data model is modeled
- When and how to expire a lock
- Christian, Kelsos, and Stefan understand it
- Documentation is improved to enable other projects to develop a Raiden client
- Presented to core devs
- Just focus on the v0.4 contracts (Red Eyes)

## Tasks
- [x] Reading smart contracts specification v0.4 (@christianbrb and @kelsos)
- [ ] Update the Raiden specification @kelsos @Dominik1999 @christianbrb 
    - [x] How do we interact on it? Google Docs or PR?
- [ ] Create a data model @andrevmatos 
    - [ ] Channel state data model @andrevmatos 
    - [ ] How the state data is getting updated @andrevmatos 
- [ ] Write down the state and actions (state changes) @andrevmatos  

## Additional information
- [Raiden Messages Specification](https://raiden-network-specification.readthedocs.io/en/latest/messaging.html)
- ~~[Description of the Raiden State Machines](https://raiden-network.readthedocs.io/en/latest/state_machines.html)~~ outdated
- [Raiden Smart Contracts Specification](https://raiden-network-specification.readthedocs.io/en/latest/smart_contracts.html)
- [Raiden developer onboarding guide](https://raiden-network.readthedocs.io/en/latest/onboarding.html)
- [Raiden transfer directory](https://github.com/raiden-network/raiden/tree/master/raiden/transfer)
- [Picture regarding the message flow from the light client kick-off workshop](https://user-images.githubusercontent.com/35398162/58076303-29f3a180-7baa-11e9-9f30-3bda0336c960.png)

### 1. Send a Locked Transfer

[Data Structure](https://raiden-network-specification.readthedocs.io/en/latest/messaging.html#id3)
- [ ] What needs to be done to prepare the data?

["Invariant"](https://raiden-network-specification.readthedocs.io/en/latest/messaging.html#id2)
- [ ] Remove what is not needed, if we don't mediate a message

[Python Code](https://github.com/raiden-network/raiden/blob/38971b372dafb3205cbd3df8cfc3306922a55eac/raiden/messages.py#L1124)
- [ ] Document Methods

### 2. Receive a Secret Request

[...]

PR #171, opened 2019-06-05
Transfers data model

andrevmatos:
- [x] Messages codecs
- [x] Transfers/channel state stored data codecs
- [ ] ~~Actions/state changes~~ (to be done as #180)

Part of #143


andrevmatos:
Let's push a partial messages implentation, as they're not yet used, and get the refactors out before. The changes implemented in this PR are alreayd out of scope, and increasing, so we can merge as is and continue on next PRs

PR #185, opened 2019-06-18
Use typesafe-actions

andrevmatos:
Replace hand-made actions with [typesafe-actions](https://github.com/piotrwitek/typesafe-actions), which are more explicit, have better typechecking capabilities, allow defining actions in a single declaration therefore easing splitting actions between domain driven modules as well as easing complying with [flux-standard-action](https://github.com/redux-utilities/flux-standard-action) pattern
- [x] implement `typesafe-actions`
- [x] adapt tests
- [x] adapt dApp events

### DIFF
--- a/raiden/src/store/actions.ts
+++ b/raiden/src/store/actions.ts
@@ -1,526 +1,178 @@
-import { AnyAction } from 'redux';
 import { BigNumber } from 'ethers/utils';
+import { createStandardAction } from 'typesafe-actions';
+
 import { RaidenMatrixSetup } from './state';
+import { ChannelId } from './types';
+import { ShutdownReason } from '../constants';
 
-export enum RaidenActionType {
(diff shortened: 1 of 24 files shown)


raiden/src/store/actions.ts
```diff
@@ -1,526 +1,178 @@
...
-  type: RaidenActionType.CHANNEL_SETTLE_FAILED;
-  tokenNetwork: string;
-  partner: string;
-}
-
-export interface MatrixSetupAction extends RaidenAction {
-  type: RaidenActionType.MATRIX_SETUP;
+}>();
+
+export const newBlock = createStandardAction('newBlock')<{ blockNumber: number }>();
```
andrevmatos: Single-property action payloads could be the payload-type itself, but for always having an object there is good for consistency as well as using the property name also as a self-documenting constraint

PR #191, opened 2019-06-24
Implements message packing functionality

### LINKED ISSUE #179: Implement packing of messages
## Description

Implement packing for messages
- [x] Delivered
- [x] Processed
- [x] Secret Request 
- [ ] Reveal Secret 
- [x] Locked Transfer
- [x] Refund Transfer 
- [x] Unlock 
- [x] Lock Expired 

## Acceptance criteria
- All messages can be encoded and packed
- Tests are included
- Output of packed message is the same, that Raiden uses to sign

## Tasks
- [ ]

kelsos:
Closes #179

- All the packed hexes are taken from the python client for the same data.

### DIFF
--- a/raiden/tests/unit/pack.spec.ts
+++ b/raiden/tests/unit/pack.spec.ts
@@ -1,39 +1,71 @@
-import { LockedTransfer, packMessage } from 'raiden/messages';
+import { LockedTransfer, MessageType, packMessage, RefundTransfer } from 'raiden/messages';
+import { PositiveInt } from 'raiden/utils/types';
+import { BigNumber } from 'ethers/utils';
 
+/* eslint-disable @typescript-eslint/camelcase */
 describe('packMessage', () => {
   test('LockedTransfer', () => {
-    /* eslint-disable @typescript-eslint/camelcase */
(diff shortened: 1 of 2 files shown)


raiden/tests/unit/pack.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { LockedTransfer, packMessage } from 'raiden/messages';
+
+describe('packMessage', () => {
+  test('LockedTransfer', () => {
+    /* eslint-disable @typescript-eslint/camelcase */
+    const message = {
```
andrevmatos: Maybe it's useful to anotate the type `message: LockedTransfer`, so we get a compile-type warning if structure changes, otherwise for tsc it's just an `object` which can have any schema, and the error will be captured only at test-time.

kelsos: I actually tried to do this but tsc would show errors be because it expected a `PositiveNumber` instead of a `number`. My guess would be that I would have to do something similar to the fold for every field. If you have any suggestion on how I would avoid doing this I would gladly accept it.

PR #193, opened 2019-06-25
Use branded types

andrevmatos:
Branded types allows us to have specific, refined types at compile time, while relying on the usual primitives at runtime, giving us even more type safety without hurting performance. This way, a variable of type `Address = HexString(20)` can't be assigned to a variable or function argument of type `Hash = HexString(32)`, even though both are `typeof string` at runtime, and the compile-time error would even give specific messages, saying e.g. the argument requires the size to be 32 bytes, and Address only has 20 bytes.
This pattern is very powerful and adds a lot of safety to the system. The drawback is the need to either validate or cast explicitly at the edges (receiving data from matrix, casting return types from 3rd-party libs, etc), but this makes it even more safe, as we then are required to acknowledge and accept data types coming from specific entrypoints.
Also, updated dependencies.


raiden/src/raiden.ts
```diff
@@ -375,11 +372,13 @@ export class Raiden {
    * @param address  Optional target address. If omitted, gets own balance
    * @returns  BigNumber containing address's token balance
    */
-  public async getTokenBalance(token: Address, address?: Address): Promise<BigNumber> {
+  public async getTokenBalance(token: string, address?: string): Promise<BigNumber> {
+    address = address || this.address;
+    if (!Address.is(address) || !Address.is(token)) throw new Error('Invalid address');
```
kelsos: nitpick: My only sligh objection here is that it is not too clear which address is invalid. Maybe in the future we can improve it to be more specific.

andrevmatos: It's not too much here to actually be used, the user is expected to handle a proper address every time. This line is just to ensure typescript knows the addresses are `Address` type/brand after this line. But yeah, I agree, this can improve later.

PR #194, opened 2019-06-27
Lossless JSON

andrevmatos:
Implements Lossless JSON encoding/decoding.
For now, used only in RaidenState/persistency, but we need to keep consistency with messages encoding, to avoid needing multiple codecs for each type to support different serialization methods.
Based on top of #193

### DIFF
--- a/raiden/src/utils/types.ts
+++ b/raiden/src/utils/types.ts
@@ -1,5 +1,6 @@
 import * as t from 'io-ts';
 import { BigNumber, bigNumberify, getAddress } from 'ethers/utils';
+import { LosslessNumber } from 'lossless-json';
 import { set } from 'lodash';
 
 /* A Subset of DOM's Storage/localStorage interface which supports async/await */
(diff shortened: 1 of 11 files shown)


andrevmatos:
TODO: remove specific types after https://github.com/josdejong/lossless-json/pull/10 is merged and released

raiden/src/utils/types.ts
```diff
@@ -9,25 +10,32 @@ export interface Storage {
...
+  u !== null && u !== undefined && typeof (u as any)['toString'] === 'function';
 const isBigNumber = (u: unknown): u is BigNumber => u instanceof BigNumber;
 /**
- * Codec of ethers.utils.BigNumber objects, to/from Decimal string
+ * Codec of ethers.utils.BigNumber objects
+ * Input can be anything bigNumberify-able: number, string, LosslessNumber or BigNumber
+ * Output is LosslessNumber, so we can JSON-serialize with 'number' types bigger than JS VM limits
+ * of ~53bits, as Raiden full-client/python stdlib json encode/decode longs as json number
+ * (which is valid json despite not javascript-friendly)
+ * TODO: get Raiden to string-serialize long ints, so we can drop lossless-json and use standard
```
kelsos: comment: As we discussed, while I am always cautious about introducing new dependencies and I would prefer to avoid when possible, in this case, I would prefer to not string-serialize long ints and keep the library. Since this is not a protocol/JSON limitation, but a language one, we should not force such a change to all other implementations of the protocol. I had to deal on Android with a backend that would sometimes return string-serialized numbers and others it would return numbers and it was introducing unnecessary bugs and extra special handling.

PR #201, opened 2019-06-28
Introduces UInt64 and UInt256 types replaces PositiveInt

### LINKED ISSUE #199: Use UInt in messages
## Description

In #193 , we introduced the `UInt` type, a brand of `BigNumber` with `Sized`, which allows to specify the size of an unsigned integer BigNumber, and validates range. We should use it in messages data model to have strict validation of intervals there, and thus prevent accepting messages with numbers outside valid ranges

`UInt`s can be used like:
```typescript
const model = t.type({
  nonce: UInt(8), // parenthesis == codec object, 8 bytes == 64bits
})
const nonce: UInt<8>; // generics notation == branded type
const siz = sizeOf(model.props.nonce); // type of siz is literal `8`, and value is 8
if (UInt(8).is(num)) {
  // num is narrowd to type UInt<8>, which is a BigNumber and is branded with size 8
}
```

## Tasks
- [x] Replace `PositiveInt` and `BigNumber[C]` with branded UInt in messages and other modules
- [x] Remove `Positive` and `PositiveInt` brands, they aren't useful anymore
- [x] Adjust tests

kelsos:
Closes #199

### DIFF
--- a/raiden/src/utils/types.ts
+++ b/raiden/src/utils/types.ts
@@ -55,21 +55,6 @@ export class EnumType<A> extends t.Type<A> {
   }
 }
 
-// Positive & PositiveInt taken from io-ts Readme
-export interface PositiveBrand {
-  readonly Positive: unique symbol;
-}
-
-export const Positive = t.brand(
-  t.number,
-  (n): n is t.Branded<number, PositiveBrand> => n >= 0, // type guard for branded values
-  'Positive', // the name must match the readonly field in the brand
-);
-export type Positive = t.TypeOf<typeof Positive>;
-
-export const PositiveInt = t.intersection([t.Int, Positive]);
-export type PositiveInt = t.TypeOf<typeof PositiveInt>;
-
 // sized brands interfaces must derive from this interface
 export interface SizedB<S extends number> {
   readonly size: S;
--- a/raiden/src/utils/types.ts
+++ b/raiden/src/utils/types.ts
@@ -1,5 +1,6 @@
 import * as t from 'io-ts';
 import { BigNumber, bigNumberify, getAddress } from 'ethers/utils';
+import { Two } from 'ethers/constants';
 import { LosslessNumber } from 'lossless-json';
 import { set } from 'lodash';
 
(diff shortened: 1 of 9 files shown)


raiden/src/utils/types.ts
```diff
@@ -192,3 +181,9 @@ set(sizedCodecCache, ['20', 'Address'], Address); // register on sized cache, fo
 
 export const Byte = UInt(1);
 export type Byte = t.TypeOf<typeof Byte>;
+
+export const UInt64 = UInt(8);
```
andrevmatos: The `UInt` factory function is cached, you don't need to define and export the trivial types. Please, remove these exports  and use direct call (`UInt(8)`) and type (`UInt<8>`) where they're needed. Declaring the types is only useful when you want to _semantically_ make a point of the layout of a given type, like `Address` being 20 bytes.

raiden/src/utils/types.ts
```diff
@@ -143,11 +128,15 @@ export function UInt<S extends number = number>(size?: S) {
...
+  const max = siz
+    ? bigNumberify(2)
+        .pow(siz * 8)
+        .sub(1)
+    : undefined;
   return (sizedCodecCache[siz][name] = t.brand(
     BigNumberC,
     (n): n is t.Branded<BigNumber, UIntB<S>> =>
-      BigNumberC.is(n) && n.gte(0) && (max === undefined || n.lt(max)),
+      BigNumberC.is(n) && n.gte(0) && (max === undefined || n.lte(max)),
```
andrevmatos: Why this change? Isn't both the same, as the numbers are always integers anyway? Also, when dealing with ranges due to bits limitation, is more common intervals of type `[a, b[`, therefore `0 <= n < max` is more common.

kelsos: When size is 32 the max will overflow BigNumber and the check will fail.

raiden/src/utils/types.ts
kelsos: My reasoning was mostly for consistency with the rest of the code. If there is no benefit I will remove it no worries.

raiden/src/utils/types.ts
andrevmatos: Oh, right, understood your thinking.
Well, it shouldn't happen here, BigNumbers only overflow if you try to construct it with a number which lost precision, otherwise they're kept as strings and BigNumber objects can actually be as large as the memory.
So, these `size` checks are just a compile time validation. At runtime they're both BigNumbers which shouldn't care for size and always test correctly.

raiden/src/utils/types.ts
andrevmatos: I think it's important to think how **semantically** a given type will pass a new idea. E.g. a `Hash` is well known, and people will understand what guarantees are involved there (in this case, only that it's 32 bytes long, and as so, at compile time is translated directly as `string & HexStringB<32>`, being equivalent to other 32bytes strings).
Maybe e.g. a `Nonce` type would make more sense, so we can specify in a centralized point all the type guarantees, but I don't think the size-only justify specifying a new type. And anyway, `UInt256` and `UInt(256)` have the same semantic value, so it doesn't justify this new type. Instead, using the `UInt` directly in the data model definitions pass exactly the idea we want, and have the same meaning at both compile and runtime.

PR #207, opened 2019-07-05
Update ui

### LINKED ISSUE #162: Checking current design and implementation - smaller details
## Description

Checking current design and implementation - smaller details

## Acceptance criteria
- Copy text has been reviewed (@christianbrb and @franzihei) and changed
- The tone has been decided about
- Colour has been decided on and implemented (eg. for error messages)
- Minor fixes are designed and implemented
- Nothing new
- Doesn't need to be perfect
- No text for the error messages
- Usability feedback is going to be implemented later

## Tasks
- [x] Check copy text with @christianbrb 
- [x] Design
- [ ] Implement

kelsos:
Closes #162


raiden-dapp/src/components/AppHeader.vue
```diff
@@ -20,31 +24,50 @@
...
-        <div class="address text-xs-left">
+    <v-layout class="app-header__bottom" align-center>
+      <v-flex xs6>
+        <div class="app-header__bottom__address text-xs-left">
           <v-tooltip bottom>
-            <template v-slot:activator="{ on }">
-              <span v-on="on"> {{ defaultAccount | truncate }}</span>
+            <template #activator="{ on }">
+              <span v-on="on">
+                {{ defaultAccount | truncate(8) }}
```
andrevmatos: Would it be useful to truncate only if width is smaller than a threshold? I don't think the address is worth it to be truncated if there's enough space

kelsos: It would expand way too much compared to the ETH amount on the other size though. 
now it looks close to the size of the eth amount on the other size.

kelsos: Also, the max size of the card is around 620px so the full address and the copy button will cross the middle of the card and get close to the amount.
image

raiden-dapp/src/views/SelectToken.vue
```diff
@@ -26,9 +22,14 @@
                   {{ token.symbol }} | {{ token.name }}
                 </v-list-tile-title>
                 <v-list-tile-sub-title class="token-address">
-                  {{ token.address }}
+                  {{ token.address | truncate }}
```
andrevmatos: There need to be ways to see the full address here.. again, you could truncate only if needed, or maybe put the address in a tooltip

kelsos:
@andrevmatos I was looking at [bem](http://getbem.com/) lately because CSS was messy and it reusing the same names made it hard to avoid duplication or overrides.

raiden-dapp/src/views/SelectToken.vue
kelsos: I will add a tooltip, thanks for pointing out.

PR #209, opened 2019-07-12
Send transfer

### LINKED ISSUE #23: I want to be able to enter the payment target address
## User Story
As Chris I want to be able to enter the payment target address so that I can make payments to a specific client.

## Tasks
- [x] Design
- [ ] Implementation

## Story Points

### LINKED ISSUE #24: I want to be able to enter a payment amount
## User Story
As Chris, I want to be able to enter a payment amount so that I can transfer a specific amount to another node.

## Tasks
- [x] Design
- [ ] Implementation

kelsos:
This PR depends on #204 

Fixes #24 
Fixes #23


raiden-dapp/src/services/raiden-service.ts
```diff
@@ -239,6 +239,15 @@ export default class RaidenService {
       this.store.commit('updateTokens', cache);
     }
   }
+
+  async transfer(token: string, target: string, amount: BigNumber) {
+    try {
+      await this.raiden.getAvailability(target);
```
andrevmatos: Can be done later, in another PR, but it's interesting to check `getAvailability` as soon as the address is input, and require/validate that the `available` is `true`. No need to cache the return of this function, as it's cached internally by the actions system, but be sure to only pass addresses which were already validated to it.

raiden-dapp/src/views/Transfer.vue
```diff
@@ -0,0 +1,138 @@
...
+
+@Component({
+  components: {
+    AddressInput,
+    AmountInput,
+    ProgressOverlay,
+    ErrorScreen
+  }
+})
+export default class Send extends Vue {
```
andrevmatos: Should the `Send` also be named `Transfer` for consistency?

kelsos: Yes, somehow I missed it.

raiden-dapp/src/services/raiden-service.ts
kelsos: Yes let's open an issue about this

kelsos: Opened the issue https://github.com/raiden-network/light-client/issues/218

Issue #181, opened 2019-06-13
Define text for error messages and implement


christianbrb:
## Description
Define text for error messages and implement

## Acceptance criteria
- Error messages are gathered at one place
- Errors are reviewed
- Messages fit the space we have (it depends)
- Messages avoid tech jargon

## Tasks
- [x] Start searching for a known way for internationalization for Vue apps
- [x] "Translate" the errors
- [x] Proof check that error messages fit within the layout

## Supporting

See https://uxplanet.org/how-to-write-good-error-messages-858e4551cd4

christianbrb:
Closed via #241

PR #227, opened 2019-07-26
Extract text resources to locale json

kelsos:
Introduces i18n and extracts all text resources to `en.json` locale file in order to make #181 easier.


kelsos:
@andrevmatos I don't have any strong feelings about this, to be honest, the examples I checked on vue i18n were using this format and structure so I followed it.

PR #236, opened 2019-08-05
Updates to vuetify 2

kelsos:
Also removes cypress to avoid downloading it on CI since it is not used.


raiden-dapp/src/components/AddressInput.vue
```diff
@@ -197,11 +197,13 @@ export default class AddressInput extends Mixins(BlockieMixin) {
   }
 }
 
-.address-input /deep/ .v-text-field__details {
+//noinspection CssInvalidPseudoSelector
```
andrevmatos: Any way to disable this globally? eslint and friends usually support tweaking individual configs

andrevmatos: From [vue's docs](https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors), it seems both syntax are supported

kelsos: So as soon as I switched from `node-sass` to `sass` as per the vuetify upgrade notes the `/deep/` operator started failing. I tried the `>>>` but it seems to not work at all. The documentation actually says that the `sass` preprocessor might have issues with `>>>`

kelsos: Ok it seems that this error was an IDE error, I removed the comment and reduced the severity to warning.

kelsos:
@andrevmatos I actually went through at least once everything. Hopefully I didn't miss anything but I will keep an eye

PR #241, opened 2019-08-06
Updated localized texts

nephix:
A couple of changes to localized texts
Relates to #181


raiden-dapp/src/locales/en.json
```diff
@@ -1,28 +1,28 @@
...
     "blockie-alt": "Selected token address blockie",
     "error": {
-      "invalid-address": "The input doesn't seem like a valid address or ens name",
-      "no-checksum": "The address is not in checksum format",
-      "empty": "The address cannot be empty",
-      "ens-resolve-failed": "Could not resolve an address for {url}"
+      "invalid-address": "Please enter a valid address or ENS name",
+      "no-checksum": "Please enter an address in checksum format. You can convert your address on https://ethsum.netlify.com",
+      "empty": "Please enter an address or ENS name",
+      "ens-resolve-failed": "Please enter a resolvable ENS name"
```
kelsos: If we are gonna remove a parameter from a resource, we should also remove it from the template.

raiden-dapp/src/locales/en.json
```diff
@@ -1,28 +1,28 @@
...
   "amount-input": {
     "input": {
-      "hint": "Please enter the amount you wish to deposit into this channel."
+      "hint": "Please enter an amount"
     },
     "error": {
-      "empty": "The amount cannot be empty",
-      "too-many-decimals": "The token supports only up to {decimals} decimals",
+      "empty": "Please enter an amount",
+      "too-many-decimals": "Please use only up to {decimals} decimals for this token",
```
kelsos: Starting with `Please use...` sounds more like a suggestion, that you could somehow avoid. Here we have a hard limitation, isn't there a better way to emphasize that fact?

raiden-dapp/src/locales/en.json
nephix: Sorry which param are you referring to?

raiden-dapp/src/locales/en.json
nephix: I think the `only` makes this a hard limit

raiden-dapp/src/locales/en.json
kelsos: The `ens-resolve-failed` originally had a `{url}` parameter. This parameter is passed to the localization function in the template.

nephix: Ah thanks good catch, will fix. Although I think they're just ignored, if no placeholders are present

raiden-dapp/src/locales/en.json
kelsos: Not really, if I say to someone "Please use only up to x of y" it is up to their discretion to use up to `x`. It might mean that there is a limited supply in total, but it doesn't make it a per person hard limit.

nephix: Which wording do you suggest then?

kelsos: I cannot think of a proper wording at the moment.

kelsos: Let's ask @christianbrb if he has a better idea

PR #243, opened 2019-08-06
Add capacity to channel$ event stream

andrevmatos:


raiden/src/types.ts
```diff
@@ -50,7 +50,13 @@ export interface RaidenChannel {
   state: ChannelState;
   ownDeposit: BigNumber;
   partnerDeposit: BigNumber;
+  // balance is difference between partner's sent tokens minus own sent tokens
+  // as of now we can only send, balance usually is a negative number, as once you send
+  // X tokens, you have X less tokens than before, and initial balance is zero
   balance: BigNumber;
+  // "distributable" capacity of channel, sum of own total deposit and balance (which as usually
+  // negative, decreases capacity)
+  capacity: BigNumber;
```
nephix: Thanks for these explanations

raiden/src/raiden.ts
```diff
@@ -168,7 +159,19 @@ export class Raiden {
                         )
                       : Zero,
                   ),
-                }),
+                  capacity = channel.own.deposit.add(balance);
+                return (partner2raidenChannel[partner] = {
```
nephix: Might be difficult to understand for contributors, could be prevented by eslint's `no-return-assign`

raiden/src/raiden.ts
```diff
@@ -168,7 +159,19 @@ export class Raiden {
                         )
                       : Zero,
                   ),
-                }),
+                  capacity = channel.own.deposit.add(balance);
```
nephix: Quite difficult to read with two ternary operators right above. Could be prevented via eslint's `one-var: ["always"]`

raiden/src/raiden.ts
andrevmatos: Hmm, I may split it in a new var, but it'd be pointless as it'd be used only once on the return below. I'll check if `transform` requires the return to be in place.

raiden/src/raiden.ts
andrevmatos: Please, no... `one-var` is horrendous, it'd make our code a spam of pointless `const`s and `let`s, while a properly configured IDE should highlight the assignment operations.
I know the multiple ternary operators are a little bit hard to follow sometimes, but that's due to TypeScript's hard requirements on type safety (as these properties aren't always available).. This will get improved on TS 3.7, which should ship with [optional chaining](https://github.com/tc39/proposal-optional-chaining) and [nullish coalescing](https://github.com/tc39/proposal-nullish-coalescing) operators. We may improve that here as well by making the `balanceProof` and `locks` properties set to defaults earlier, so the type narrowing on `state` can infere their presence.

PR #259, opened 2019-08-14
Sign message via personal_sign

### LINKED ISSUE #223: Ensure that the dApp works on wallets with dApp browser
## Description

Out of curiosity yesterday I tried to check how the dApp runs in status.im. Apparently, it doesn't after spending a bit investigating I found out that a `raidenShutdown` is fired immediately after `raidenInit`.

```
{
  meta: undefined
  payload: {reason: Error: The method eth_sign does not exist/is not available at Object.eval [as   callback] (webpac…}
  type: "raidenShutdown"
}
```

https://medium.com/metamask/the-new-secure-way-to-sign-data-in-your-browser-6af9dd2a1527

## Acceptance criteria
- Works in Status and imToken
- Needs to be tested with https

## Tasks
- [ ]

nephix:
Fixes #223

### DIFF
--- a/raiden/src/raiden.ts
+++ b/raiden/src/raiden.ts
@@ -217,9 +217,6 @@ export class Raiden {
 
     // initialize epics, this will start monitoring previous token networks and open channels
     this.store.dispatch(raidenInit());
-    
-    // Patch provider's sign method (https://github.com/raiden-network/light-client/issues/223)
-    patchSignSend(this.provider);
   }
 
   /**
@@ -257,6 +254,10 @@ export class Raiden {
     } else {
       provider = new Web3Provider(connection);
     }
+
+    // Patch provider's sign method (https://github.com/raiden-network/light-client/issues/223)
+    patchSignSend(provider);
+
     const network = await provider.getNetwork();
 
     // if no ContractsInfo, try to populate from defaults
--- a/raiden/src/raiden.ts
+++ b/raiden/src/raiden.ts
@@ -70,6 +70,7 @@ import {
 } from './transport/actions';
 import { transfer, transferred, transferFailed } from './transfers/actions';
 import { makeSecret } from './transfers/utils';
+import { patchSignSend } from './utils/ethers';
 
 export class Raiden {
   private readonly provider: JsonRpcProvider;
@@ -216,6 +217,9 @@ export class Raiden {
 
     // initialize epics, this will start monitoring previous token networks and open channels
     this.store.dispatch(raidenInit());
+    
+    // Patch provider's signMessage method (https://github.com/raiden-network/light-client/issues/223)
+    patchSignSend(this.provider);
   }
 
   /**
--- a/raiden/src/raiden.ts
+++ b/raiden/src/raiden.ts
@@ -218,7 +218,7 @@ export class Raiden {
     // initialize epics, this will start monitoring previous token networks and open channels
     this.store.dispatch(raidenInit());
     
-    // Patch provider's signMessage method (https://github.com/raiden-network/light-client/issues/223)
+    // Patch provider's sign method (https://github.com/raiden-network/light-client/issues/223)
     patchSignSend(this.provider);
   }
 
--- a/raiden/src/utils/ethers.ts
+++ b/raiden/src/utils/ethers.ts
@@ -20,7 +20,7 @@ import { filter, first, map, mergeAll, switchMap, withLatestFrom, mergeMap } fro
 export function fromEthersEvent<T>(
   target: Provider | Contract,
   event: EventFilter | string,
-  resultSelector?: (...args: any[]) => T,  // eslint-disable-line
+  resultSelector?: (...args: any[]) => T, // eslint-disable-line
 ): Observable<T> {
   return fromEventPattern<T>(
     (handler: Listener) => target.on(event, handler),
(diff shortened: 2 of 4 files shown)


raiden/src/utils/ethers.ts
```diff
@@ -107,3 +107,33 @@ export function getEventsStream<T extends any[]>(
...
+    if (method === 'eth_sign') {
+      // try 'personal_sign' by default instead of 'eth_sign'
+      return origSend
+        .bind(this)('personal_sign', [params[1], params[0]])
+        .catch(err => {
+          // on first error, if personal_sign isn't available
+          if (
+            err instanceof Error &&
+            (err.message.includes('The method personal_sign does not exist') ||
+              err.message.includes('Method personal_sign not supported'))
```
nephix: This 2nd error messages was thrown by `ganache-cli` and caused the e2e tests to fail, so I added it

raiden/src/raiden.ts
```diff
@@ -216,6 +217,9 @@ export class Raiden {
 
     // initialize epics, this will start monitoring previous token networks and open channels
     this.store.dispatch(raidenInit());
+    
+    // Patch provider's sign method (https://github.com/raiden-network/light-client/issues/223)
+    patchSignSend(this.provider);
```
andrevmatos: Please, patch in the `create` static method, right below provider authentication, instead of here. constructor is expected to receive a _compatible_ provider, and this is a workaround is for ethers specific cases. If a user can already provide the compatible provider by theirselves, it's supported for them to bypass this factory static method and instantiate directly.

raiden/src/utils/ethers.ts
andrevmatos: No problem, I suppose we'll need to add other messages here in the future. Error messages aren't very standardized between clients unfortunatelly, and we need to specify each supported cases.

raiden/src/utils/ethers.ts
```diff
@@ -107,3 +107,33 @@ export function getEventsStream<T extends any[]>(
...
+ *
+ * @param provider  A JsonRpcProvider instance to patch
+ */
+export function patchSignSend(provider: JsonRpcProvider): void {
+  const origSend: (method: string, params: any) => Promise<any> = provider.send;
+  provider.send = async function(method: string, params: any): Promise<any> {
+    if (method === 'eth_sign') {
+      // try 'personal_sign' by default instead of 'eth_sign'
+      return origSend
+        .bind(this)('personal_sign', [params[1], params[0]])
```
andrevmatos: Also, I realized `bind` isn't appropriate for this, as it will bind the method indefinitelly. Please, replace it with:
```suggestion
        .apply(this, ['personal_sign', [params[1], params[0]]])
```

raiden/src/utils/ethers.ts
```diff
@@ -107,3 +107,33 @@ export function getEventsStream<T extends any[]>(
...
+            (err.message.includes('The method personal_sign does not exist') ||
+              err.message.includes('Method personal_sign not supported'))
+          ) {
+            provider.send = origSend; // un-patch
+            return provider.send(method, params); // and retry with eth_sign
+          }
+          throw err; // else, re-raise
+        });
+    }
+    return origSend.bind(this)(method, params);
```
andrevmatos: Same
```suggestion
    return apply(this, [method, params]);
```

raiden/src/utils/ethers.ts
nephix: Good catch!

andrevmatos:
Nice, thank you. Just a small note about or usual Git workflow: while the PR is under review, you can and should rebase, reset and squash whenever needed, and forcepush it to the PR branch. The PRs don't need to be a single commit, but usually we prefer to keep relevant commits only, which make sense on the history and can help understand what was done on it. So, e.g. the `revert`s on this PR which deal with commits which were also added on the PR, could just be dropped on a rebase, or changes in behavior which were added on this PR, could be amended instead of a new commit. I'm not sure if this is in the contributing guide, but if not, could be a nice addition to formalize our git workflow there.

nephix:
> Nice, thank you. Just a small note about or usual Git workflow: while the PR is under review, you can and should rebase, reset and squash whenever needed, and forcepush it to the PR branch. The PRs don't need to be a single commit, but usually we prefer to keep relevant commits only, which make sense on the history and can help understand what was done on it. So, e.g. the `revert`s on this PR which deal with commits which were also added on the PR, could just be dropped on a rebase, or changes in behavior which were added on this PR, could be amended instead of a new commit. I'm not sure if this is in the contributing guide, but if not, could be a nice addition to formalize our git workflow there.

I think it's enough to squash the commits when merging the branch. This keeps unnecessary overhead to a minimum and the branch in the forked repo can be deleted afterwards. Or did I miss any downsides @andrevmatos ?

andrevmatos:
Depends on the size of the PR. For small, self-contained ones which add one change or fix, squashing everything is ok. But for anything bigger, it's preferably to have the changes that were made semantically listed on the history, but clean on the steps that may have led to it through review.

nephix:
I would prefer to have multple PRs for such cases. If I need to understand the history on top of the code changes that usually means that the PR or ticket is already too big and should have been split up.

andrevmatos:
This is good for a guideline, but in practice, often some PRs can't be split because a feature implementation or bugfix is _atomic_, and either is done whole and merged as so in master, or leaves master broken during intermediary steps, even worse when there're architectural decision changes for it and something needs to be changed for code already commited for this feature.
As much as possible, we should try to create small PRs, but these guidelines arised on Raiden Core exactly because this kind of feature PR happened from time to time. It's not expected one needs to read the history to understand the PR, but history also works as diff tracking/blame tool to understand who, why and the context something took place from the past.

christianbrb:
@nephix @andrevmatos Great PR discussion, that we should follow up in person

PR #271, opened 2019-08-21
Transfers history, public notifications

### LINKED ISSUE #263: Expose pending transfers info in public API
## Description

Sdk public API (`Raiden`) exposes the `transfer` method, but besides its `Promise` return, there's no way yet for users to query pending transfers status and info.

It may be useful to enable users to get it through a reactive property (an Observable, maybe `transfers$`, like `channels$`), so they can both know current status and get notified when it changes. 

A first usecase is providing a counter of pending transfers to the user in the dApp UI, and prevent/warn users trying to leave/close tab while there're pending transfers: `Promise` returned value for `transfer` method just informs user when a transfer succeeded or failed irreversibly, but on the fail case, some actions for the complete transfer protocol may still be pending even after we know/notifier about transfer failure; although raiden protocol/sdk is able to recover if leaving (hopefully at any point in protocol) and coming back online later, it's optimal to wait for the transfer task to complete (e.g. expiring the lock).

## Acceptance criteria
- There's a `transfers$` public property in `Raiden`
- User can query and get notified about current pending transfers

## Tasks
- [ ]

andrevmatos:
Fix #263 
channel state `history` property is gone. Now, success, failed, completed, channel-closed transfers are kept indefinitely in `state.sent` mapping (a clear history option will be added later). 
`Raiden.transfer` returned Promise resolves to `secrethash` as soon as the transfer is signed (or rejects if cancelled), and users may subscribe to `Raiden.transfers$` observable to get notified about update for this specific transfer, from the pending state until it's completed. History is kept, and at startup, all transfers history is emitted (as completed transfers), and may be stored to be presented to the user at any time, as well as pending transfers on restart, for example.

- [x] `RaidenState.sent` keeps history (isn't cleared upon transfer completion), timestamp of every message/event is saved as well, `ChannelEnd.history` gone as it was duplicating this info
- [x] `Raiden.transfers$` observable notifies about historic and current pending transfers
- [x] `Raiden.transfer` async method resolves to `secrethash` (transfer unique identifier) as soon as transfer is signed and persisted
- [x] Unit tests, and also first e2e tests for transfers!

### DIFF
--- a/raiden/src/raiden.ts
+++ b/raiden/src/raiden.ts
@@ -675,21 +675,16 @@ export class Raiden {
    * queried with this id on this.transfers$ observable, which will just have emitted the 'pending'
    * transfer. Any following transfer state change will be notified through this observable.
    *
-   * Coverage ignored until we handle the full transfer lifecycle
-   * TODO: remove istanbul ignore when covered
-   *
    * @param token  Token address on currently configured token network registry
    * @param target  Target address (must be getAvailability before)
    * @param amount  Amount to try to transfer
-   * @param opts.paymentId  Optionally specify a paymentId to use for this transfer
-   * @param opts.secret  Optionally specify a secret to use on this transfer
-   *    (in which case, it'll be registered and revealed to target)
-   * @param opts.secrethash  Optionally specify a secrethash to use. If secret is provided,
-   *    secrethash must be the keccak256 hash of the secret. If no secret is provided, the target
-   *    must be informed of it by other means/externally.
+   * @param opts  Optional parameters for transfer:
+   *                - paymentId  payment identifier, a random one will be generated if missing
+   *                - secret  Secret to register, a random one will be generated if missing
+   *                - secrethash  Must match secret, if both provided, or else, secret must be
+   *                              informed to target by other means, and reveal can't be performed
    * @returns A promise to transfer's secrethash (unique id) when it's accepted
    */
-  /* istanbul ignore next */
   public async transfer(
     token: string,
     target: string,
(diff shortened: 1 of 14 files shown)


raiden/src/raiden.ts
```diff
@@ -175,6 +184,22 @@ export class Raiden {
...
     );
 
+    this.transfers$ = state$.pipe(
+      map(state => state.sent),
+      distinctUntilChanged(),
+      concatMap(sent => from(Object.entries(sent))),
+      /* this scan stores a reference to each [key,value] in 'acc', and emit as 'changed' iff it
+       * changes from last time seen. It relies on value references changing only if needed */
+      scan<[string, SentTransfer], { acc: SentTransfers; changed?: SentTransfer }>(
+        ({ acc }, [k, v]) => (acc[k] === v ? { acc } : { acc: { ...acc, [k]: v }, changed: v }),
```
kelsos: While I understand exactly what this line does. I had to pause for a minute or two to read it :smile:

andrevmatos: Haha, sorry, I'll try to make it more explicit.. is just that those mappers/reducers fit nicely on a single line, but sometimes can be harder to understand.

raiden/src/raiden.ts
```diff
@@ -191,7 +191,12 @@ export class Raiden {
       /* this scan stores a reference to each [key,value] in 'acc', and emit as 'changed' iff it
        * changes from last time seen. It relies on value references changing only if needed */
       scan<[string, SentTransfer], { acc: SentTransfers; changed?: SentTransfer }>(
-        ({ acc }, [k, v]) => (acc[k] === v ? { acc } : { acc: { ...acc, [k]: v }, changed: v }),
+        ({ acc }, [secrethash, sent]) => {
+          // if ref didn't change, emit previous accumulator, without 'changed' value
+          if (acc[secrethash] === sent) return { acc };
+          // else, update ref in 'acc' and emit value in 'changed' prop
+          else return { acc: { ...acc, [secrethash]: sent }, changed: sent };
```
kelsos: Thanks for the naming change now it is more clear :)

PR #295, opened 2019-08-29
Adds new messages, updates message tests

### LINKED ISSUE #291: Upgrade messages to be compatible with Alderaan clients
## Description

After #290, we need to ensure the messages we send, receive (validate) and reply are made compatible with the current Raiden Python Client for Alderaan. This includes both adapting the existing messages formats and validation, adjusting the logic for anything that may have changed on them, and adding new messages and replies which may come and be part of the protocol, even if we don't support the feature yet.

Part of #244 

## Acceptance criteria
- the SDK is able to send direct (non-mediated transfers) to a connected partner/neighbor/hub
- mediated transfers requests fails immediately
- the SDK tries to not get out-of-sync with partner even if receiving messages for non-supported features (withdraw)
- All tests should pass
- Work is done on `next` branch

## Tasks
- [x] Adapt messages to current specification (e.g. route included in a LockedTransfer)
- [x] Fails immediately if transfer request target isn't recipient
- [x] In case there're multiple channels, prioritize direct route
- [x] Add withdraw messages schemas and validation
- [x] ~~Check (and implement, if possible) if we can just reply withdraw requests with `Processed` (to avoid getting spammed by the retries), and let it expire (like we do for receiving transfers)~~ **Withdraw requests:** Implement handling of `WithdrawRequest`, sends `WithdrawConfirmation`; reply `WithdrawExpiration` with `Processed` (as it may not be required, see https://github.com/raiden-network/raiden/issues/4721), listen to `ChannelWithdraw` events and store it in `ChannelEnd` (done in #304)

kelsos:
Closes #291

### DIFF
--- a/raiden-ts/src/messages/utils.ts
+++ b/raiden-ts/src/messages/utils.ts
@@ -82,7 +82,7 @@ export function createMessageHash(message: EnvelopeMessage): Hash {
     case MessageType.LOCKED_TRANSFER:
     case MessageType.REFUND_TRANSFER:
       // hash of packed representation of the whole message
-      const packed = concat([
+      let packed = concat([
         encode(CMDIDs[message.type], 1),
         encode(message.message_identifier, 8),
         encode(message.payment_identifier, 8),
@@ -95,11 +95,10 @@ export function createMessageHash(message: EnvelopeMessage): Hash {
         encode(message.lock.amount, 32),
         encode(message.fee, 32),
       ]);
-      const hashable =
-        message.type === MessageType.LOCKED_TRANSFER
-          ? concat([packed, createMetadataHash(message.metadata)])
-          : packed;
-      return keccak256(hashable) as Hash;
+
+      if (message.type === MessageType.LOCKED_TRANSFER)
+        packed = concat([packed, createMetadataHash(message.metadata)]);
+      return keccak256(packed) as Hash;
     case MessageType.UNLOCK:
       return keccak256(
         concat([
@@ -199,7 +198,7 @@ export function packMessage(message: Message) {
           encode(0, 3),
           encode(message.message_identifier, 8),
         ]),
-      );
+      ) as HexString<12>;
     case MessageType.WITHDRAW_REQUEST:
     case MessageType.WITHDRAW_EXPIRED:
     case MessageType.WITHDRAW_CONFIRMATION:
@@ -213,7 +212,7 @@ export function packMessage(message: Message) {
           encode(message.total_withdraw, 32),
           encode(message.expiration, 32),
         ]),
-      );
+      ) as HexString<200>;
   }
 }
 
--- a/raiden-ts/src/messages/utils.ts
+++ b/raiden-ts/src/messages/utils.ts
@@ -2,15 +2,22 @@
 import { ThrowReporter } from 'io-ts/lib/ThrowReporter';
 import { isLeft } from 'fp-ts/lib/Either';
 
-import { Signer } from 'ethers';
+import { ethers, Signer } from 'ethers';
 import { keccak256, verifyMessage } from 'ethers/utils';
-import { concat, hexlify, arrayify } from 'ethers/utils/bytes';
+import { arrayify, concat, hexlify } from 'ethers/utils/bytes';
 import { HashZero } from 'ethers/constants';
 
 import { Address, Hash, HexString, Signature, UInt } from '../utils/types';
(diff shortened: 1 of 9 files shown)


raiden-ts/src/messages/utils.ts
```diff
@@ -2,15 +2,22 @@
 import { ThrowReporter } from 'io-ts/lib/ThrowReporter';
 import { isLeft } from 'fp-ts/lib/Either';
 
-import { Signer } from 'ethers';
+import { ethers, Signer } from 'ethers';
```
andrevmatos: You could import `RLP` directly from `ethers/utils`. `Importing the whole `ethers` module again only for `ethers.utils.RLP.encode` isn't good.

kelsos: I completely missed that thank you, I will fix it :)

raiden-ts/src/messages/utils.ts
```diff
@@ -54,61 +84,40 @@ export function createMessageHash(message: EnvelopeMessage): Hash {
...
+        encode(message.lock.expiration, 32),
+        encode(message.token, 20),
+        encode(message.recipient, 20),
+        encode(message.target, 20),
+        encode(message.initiator, 20),
+        encode(message.lock.secrethash, 32),
+        encode(message.lock.amount, 32),
+        encode(message.fee, 32),
+      ]);
+      const hashable =
```
andrevmatos: Nitpick: wouldn't it be clearer to `let packed`, and then 
```
if (message.type === MessageType.LOCKED_TRANSFER)
  packed = concat([packed, createMetadataHash(message.metadata)])`;
return keccak256(packed) as Hash;
```
Like, to tell: return for both is keccak of packed, but for the locked transfer, the metadata hash is appended.
This is just a suggestion in order to prevent the need for a new var (`hashable`) which has the same meaning as `packed` before hashing (the packed data which will be hashed), so you can ignore if you think it's better the way it is

raiden-ts/src/messages/utils.ts
```diff
@@ -54,61 +84,40 @@ export function createMessageHash(message: EnvelopeMessage): Hash {
...
+        message.type === MessageType.LOCKED_TRANSFER
+          ? concat([packed, createMetadataHash(message.metadata)])
+          : packed;
+      return keccak256(hashable) as Hash;
     case MessageType.UNLOCK:
       return keccak256(
         concat([
           encode(CMDIDs[message.type], 1),
-          encode(0, 3),
```
andrevmatos: Whoah, they really stripped the `additional_hash` for these messages, didn't they? I hope it doesn't miss important data for the integrity check. This doesn't affect the smartcontract at all, as soon as both clients compose this `additional_hash` exactly the same, so tests which compare our packing with output of python client should be enough. Thank you for going over it

raiden-ts/src/messages/utils.ts
```diff
@@ -21,8 +28,31 @@ const CMDIDs: { readonly [T in MessageType]: number } = {
...
   [MessageType.LOCK_EXPIRED]: 13,
+  [MessageType.TO_DEVICE]: 14,
+  [MessageType.WITHDRAW_REQUEST]: 15,
+  [MessageType.WITHDRAW_CONFIRMATION]: 16,
+  [MessageType.WITHDRAW_EXPIRED]: 17,
 };
 
+// raiden_contracts.constants.MessageTypeId
+enum MessageTypeId {
+  BALANCE_PROOF = 1,
```
andrevmatos: Nice

raiden-ts/src/messages/utils.ts
```diff
@@ -185,6 +194,28 @@ export function packMessage(message: Message) {
...
+        concat([
+          encode(message.token_network_address, 20),
+          encode(message.chain_id, 32),
+          encode(MessageTypeId.WITHDRAW, 32),
+          encode(message.channel_identifier, 32),
+          encode(message.participant, 20),
+          encode(message.total_withdraw, 32),
+          encode(message.expiration, 32),
+        ]),
+      );
```
andrevmatos: Just for completeness, could you add the size of the packed withdraw message?
```suggestion
      ) as HexString<200>;
```

raiden-ts/src/messages/utils.ts
kelsos: They removed quite a few properties from the hashing yes.

PR #304, opened 2019-09-03
Implements Withdraw handling

### LINKED ISSUE #248: I want to support a hub to withdraw tokens from the channel
## User Story
As Chris I want to support a hub to withdraw tokens from the channel so that he get's liquidity out of the channel

andrevmatos:
Implements withdraw request handling for requests and dummy `Processed` for `WithdrawExpired`.
Also, listen and store `ChannelWithdraw` events.
Fix #248 , solves the withdraw part of #291

### DIFF
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -545,7 +545,7 @@ export class Raiden {
     const tokenNetwork = state.tokens[token];
     if (!tokenNetwork) throw new Error('Unknown token network');
     deposit = bigNumberify(deposit);
-    if (!UInt(32).is(deposit)) throw new Error('invalid deposit');
+    if (!UInt(32).is(deposit)) throw new Error('invalid deposit: must be 0 < amount < 2^256');
     const promise = this.action$
       .pipe(
         filter(isActionOf([channelDeposited, channelDepositFailed])),
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -544,6 +544,8 @@ export class Raiden {
     const state = this.state;
     const tokenNetwork = state.tokens[token];
     if (!tokenNetwork) throw new Error('Unknown token network');
+    deposit = bigNumberify(deposit);
+    if (!UInt(32).is(deposit)) throw new Error('invalid deposit');
     const promise = this.action$
       .pipe(
         filter(isActionOf([channelDeposited, channelDepositFailed])),
@@ -557,9 +559,7 @@ export class Raiden {
         }),
       )
       .toPromise();
-    this.store.dispatch(
-      channelDeposit({ deposit: bigNumberify(deposit) }, { tokenNetwork, partner }),
-    );
+    this.store.dispatch(channelDeposit({ deposit }, { tokenNetwork, partner }));
     return promise;
   }
 
--- a/raiden-ts/src/transfers/epics.ts
+++ b/raiden-ts/src/transfers/epics.ts
@@ -377,13 +377,13 @@ function makeAndSignLockExpired(
 }
 
 /**
- * Create an observable to compose and sign a WithdrawConfirmation message
+ * Create an observable to compose and sign a [[WithdrawConfirmation]] message
  *
  * Validate we're inside expiration timeout, channel exists and is open, and that total_withdraw is
  * less than or equal withdrawable amount (while we don't receive, partner.deposit +
  * own.transferredAmount).
- * We need it inside transferGenerateAndSignEnvelopeMessageEpic concatMap/lock because we read and
- * change the 'nonce', even though WithdrawConfirmation doesn't carry a full balanceProof.
+ * We need it inside [[transferGenerateAndSignEnvelopeMessageEpic]] concatMap/lock because we read
+ * and change the 'nonce', even though WithdrawConfirmation doesn't carry a full balanceProof.
  * Also, instead of storing the messages in state and retrying, we just cache it and send cached
  * signed message on each received request.
  *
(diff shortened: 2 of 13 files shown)


raiden-ts/src/transfers/epics.ts
```diff
@@ -974,16 +1092,53 @@ export const transferChannelClosedEpic = (
...
+  action$: Observable<RaidenAction>,
+  state$: Observable<RaidenState>,
+): Observable<ActionType<typeof transferRefunded | typeof transferFailed>> =>
+  action$.pipe(
+    filter(isActionOf(messageReceived)),
+    withLatestFrom(state$),
+    mergeMap(function*([action, state]) {
+      const message = action.payload.message;
+      if (!message || !Signed(RefundTransfer).is(message)) return;
+      const secrethash = message.lock.secrethash;
```
nephix: We could destructure here 

- `action.payload.message`
- `message.lock.secrethash`
- `state.sent[secrethash].transfer[1]`

might improve readability of the if condition

raiden-ts/src/transfers/epics.ts
```diff
@@ -1025,36 +1181,48 @@ export const transferReceivedReplyProcessedEpic = (
 };
 
 /**
- * Receiving RefundTransfer for pending transfer fails it
+ * When receiving a WithdrawRequest message, create the respective withdrawReceiveRequest action
```
nephix: `[[WithdrawRequest]]` although not sure if our typedoc process links this correctly

raiden-ts/src/transport/epics.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/camelcase */
```
nephix: Because of `user_id`? We should maybe add exceptions in our eslint config a la 
```
"@typescript-eslint/camelcase": ["error", { allow: ["user_Id"] }]
```
for variable names we don't have any control over

raiden-ts/tests/unit/reducers.spec.ts
```diff
@@ -230,6 +233,105 @@ describe('raidenReducer', () => {
...
+            participant: partner,
+            totalDeposit: deposit,
+            txHash,
+          },
+          { tokenNetwork, partner },
+        ),
+      ].reduce(raidenReducer, state);
+    });
+
+    test('channel not in open state', () => {
```
nephix: It's a bit hard to follow what the expected result here is

raiden-ts/tests/unit/reducers.spec.ts
```diff
@@ -230,6 +233,105 @@ describe('raidenReducer', () => {
...
+            txHash,
+          },
+          { tokenNetwork, partner },
+        ),
+      );
+      expect(newState.channels).toMatchObject({
+        [tokenNetwork]: {
+          [partner]: {
+            state: ChannelState.open,
+            own: { deposit, withdraw }, // our totalWithdraw was updated
```
nephix: Comments like these definitely help to understand what the test case is about 👍

raiden-ts/src/transport/epics.ts
andrevmatos: I thought about it, but I'd rather not have that for the whole projects, as only some files need to deal with snake case names (mainly matrix and messages, as these communicate with the external world with functions and objects externally defined to be on snake case). Otherwise, I'd prefer to have camel enforced on the rest of the project.

raiden-ts/src/transfers/epics.ts
andrevmatos: This would require splitting the `if`, because e.g. `sent[secrethash]` isn't defined if `!(secrethash in state.sent)`, and so on. Can you suggest code on how you think it'd be better, please?

raiden-ts/tests/unit/reducers.spec.ts
andrevmatos: I'll try to make it clearer

raiden-ts/src/raiden.ts
```diff
@@ -544,6 +544,8 @@ export class Raiden {
     const state = this.state;
     const tokenNetwork = state.tokens[token];
     if (!tokenNetwork) throw new Error('Unknown token network');
+    deposit = bigNumberify(deposit);
+    if (!UInt(32).is(deposit)) throw new Error('invalid deposit');
```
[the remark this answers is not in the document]
andrevmatos: It's a very edgy case, user providing a negative or larger than 256bit number, so I thought the straightforward error would be enough, but I'll write something more informative then

raiden-ts/src/transfers/epics.ts
nephix: I think splitting up the if might actually be good as we could do a short comment on why we return there (besides that we already have another if above)

I thought about something like this:
```
const sentTransfer = state.sent[secrethash];
if (!sentTransfer) return;

const [_, firstTransfer] = sentTransfer.transfer,
  {
    recipient,
    payment_identifier,
    lock: { amount, expiration },
  } = firstTransfer;

if (
  message.initiator !== recipient ||
  !message.payment_identifier.eq(payment_identifier) ||
  !message.lock.amount.eq(amount) ||
  !message.lock.expiration.eq(expiration) ||
  sentTransfer.unlock || // already unlocked
  sentTransfer.lockExpired || // already expired
  sentTransfer.channelClosed || // channel closed
  message.lock.expiration.lte(state.blockNumber) // lock expired but transfer didn't yet
)
  return;
```

But it's mostly about readability so no big deal

andrevmatos: Did almost what you suggested, just not the destructuring of the first transfer. Thank you

PR #308, opened 2019-09-04
Added better payment feedback handling

### LINKED ISSUE #278: I want to get feedback on the UI, if a payment has been successful
## User Story
As Chris, I want to get feedback on the UI, if a payment has been successful so that I don't have to look at the changing capacity to be sure, that the transfer was successful.

## Acceptance criteria
- It is successful when a transfer has been completed
- It fails, when the transfer has failed 
- The screen is shown (Transfering -> Done)
- Promise works, when the transfer is complete (transfer observable)

nephix:
Resolves #278

### DIFF
--- a/raiden-dapp/src/services/raiden-service.ts
+++ b/raiden-dapp/src/services/raiden-service.ts
@@ -1,4 +1,4 @@
-import { Raiden, RaidenChannel } from 'raiden-ts';
+import { Raiden, RaidenChannel, RaidenSentTransfer } from 'raiden-ts';
 import { Store } from 'vuex';
 import { RootState } from '@/types';
 import { Web3Provider } from '@/services/web3-provider';
@@ -11,7 +11,7 @@ import {
 } from '@/model/types';
 import { BigNumber } from 'ethers/utils';
 import { Zero } from 'ethers/constants';
-import { filter } from 'rxjs/internal/operators';
+import { filter, first } from 'rxjs/internal/operators';
 
 export default class RaidenService {
   private _raiden?: Raiden;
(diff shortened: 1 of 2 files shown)


raiden-dapp/src/services/raiden-service.ts
```diff
@@ -259,7 +259,17 @@ export default class RaidenService {
   async transfer(token: string, target: string, amount: BigNumber) {
     try {
       await this.raiden.getAvailability(target);
-      await this.raiden.transfer(token, target, amount);
+      const secretHash = await this.raiden.transfer(token, target, amount);
+
+      // Wait for transaction to be completed
+      await this.raiden.transfers$
```
nephix: @andrevmatos it might make sense to have some kind of follow up ticket that implements timeout handling here, e.g. if the user declines transactions in MetaMask. What do you think?

andrevmatos: @nephix I would not suggest the timeout, or else it may give incomplete results. The transfer will ALWAYS complete eventually, in the worst case after `LockExpired` `Processed` (which may take several minutes, but the idea is still to give the proper feedback for the user here)

andrevmatos: If the user declines the first prompt on Metamask, the `transfer` promise will reject right alway. If they reject any of the following ones, it'll eventually be asked again (upon messages retries) until transfer expires or they accept it, so no exception case should be fatal or leave a transfer pending forever. Even closing the channel mid-transfer completes it.

nephix: Thanks for the clarification @andrevmatos

raiden-dapp/tests/unit/raiden-service.spec.ts
```diff
@@ -659,4 +662,16 @@ describe('RaidenService', () => {
       expect(transfer).toHaveBeenCalledWith('0xtoken', '0xpartner', One);
     });
   });
+
+  it('resolve ens addresses', async () => {
```
nephix: Added this test to increase overall code coverage because codecov complains that https://github.com/raiden-network/light-client/pull/308/files#diff-ddfc85a29ab30ce1822e4b860e4305d2R268 isn't hit during tests and I didn't manage to get it hit.

andrevmatos: Ok, this is not ideal, but I don't see big deal with it right now, this part of the code will need to be reworked when implementing pending/parallel transfers badge/dropdown, so then we can add proper coverage on this

PR #322, opened 2019-09-06
Details on how to try the Raiden Demo dApp

taleldayekh:
An updated readme with more detailed instructions in a step-by-step guide on how to try out the Raiden Demo dApp.

What remains to be improved is adding some screenshots (need to decide which are best suited) and updating the hub to a 24/7 Raiden node.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 </p>
 
 <p align="center">
-  <a href="https://lightclient.raiden.network/">Play with the Raiden Demo dApp</a>
+  <a href="https://lightclient.raiden.network/">Try Out the Raiden Demo dApp</a>
 </p>
 
 <p align="center">
--- a/README.md
+++ b/README.md
@@ -46,20 +46,21 @@ The [Raiden dApp](#raiden-dapp) is a reference implementation of the Raiden Ligh
 > **INFO:** The Light Client SDK and dApp are **work in progress** and can only be used on the Ethereum **Testnets**.
 
 ## Table of Contents
+- [Table of Contents](#table-of-contents)
 - [About The Project](#about-the-project)
 - [Architecture](#architecture)
-  * [Raiden Light Client SDK](#raiden-light-client-sdk)
-  * [Raiden dApp](#raiden-dapp)
-  * [Architecture diagram](#architecture-diagram)
+  - [Raiden Light Client SDK](#raiden-light-client-sdk)
+  - [Raiden dApp](#raiden-dapp)
+  - [Architecture diagram](#architecture-diagram)
 - [Getting Started](#getting-started)
-  * [Learn about Raiden](#learn-about-raiden)
-  * [Play with the Raiden Demo dApp](#play-with-the-raiden-demo-dapp)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -139,9 +140,47 @@ If you didn't use Raiden before, you can
...
+You need to have MetaMask installed for your browser.
+
+1. Visit the [MetaMask website](https://metamask.io/) to download and install MetaMask.
+2. Select the Ropsten Test Network when you've logged in to your MetaMask.
+
+
+__Step 1: Acquire and Wrap ETH__
+
+1. Visit the Ropsten faucet at [https://faucet.ropsten.be](https://faucet.ropsten.be).
+2. Enter your keystore address associated with the Ropsten Test Network.
```
nephix: I guess `wallet address` instead of `keystore address`

README.md
```diff
@@ -139,9 +140,47 @@ If you didn't use Raiden before, you can
...
+
+__Step 2: Connect to the Raiden Demo dApp__
+
+1. Visit the Raiden Demo dApp at [https://lightclient.raiden.network/](https://lightclient.raiden.network/).
+2. Click __Connect__ to connect the dApp to your MetaMask.
+3. Select your newly wrapped token in the list of available tokens.
+
+__Step 3: Select a Hub and Open a Channel__
+
+1. Enter ```0x5257964Ef9b81fba7276AF2a97c111AaD7B840D6``` as your hub for testing.
```
nephix: More of a question, should we explain what `0x5257964Ef9b81fba7276AF2a97c111AaD7B840D6` is?

README.md
```diff
@@ -139,9 +140,47 @@ If you didn't use Raiden before, you can
...
+2. Enter the amount of wrapped ETH (WETH) you want to deposit when opening a channel.
+3. Sign the deposit with your MetaMask.
+4. Click __Open Channel__.
+5. Sign again with your MetaMask when prompted for "Open Channel", "Approve" and "Set Total Deposit".
+
+You can now start making payments. To do so, simply:
+
+1. Enter the address of the receiver of your payment.
```
nephix: Again, more of a question: Do we have addresses available to send payments to?

taleldayekh:
> Good job @taleldayekh looks good. I added a couple of questions I'm not sure about.

Thank you @nephix, and thanks for having a look!

README.md
taleldayekh: Yes, I was thinking about wallet but was a bit unsure whether to use that term since I've only used "keystore address" previously. But I guess this confirms that wallet is a valid term. I will go ahead and change to wallet. :)

README.md
taleldayekh: I'm adding a short explanation about it beeing a Raiden test node. I believe that the address will change soon, I'm not sure how temporary this one is. I'm supposed to get access to a 24/7 node from Ulo and will update the address and add additional details as soon as that is solved.

README.md
taleldayekh: Not what I am aware of. But I can look into it and if we do or if we set up addresses I can update it later on.

PR #323, opened 2019-09-06
Added route guard for routes with a token network address

### LINKED ISSUE #275: Guard routes that contain token addresses
## Description

At the moment there is no guarding at the routes so if you copy a route from the channels of a token in kovan, switch to rinkeby and use the same route the dApp will fail silently.

There should be a guard that disallows the navigation to invalid tokens for the network.

## Acceptance criteria
- Should only go to the tokens available in the same network

## Tasks
- [x] Research how routing works in Vue.js

nephix:
Resolves #275

### DIFF
--- a/raiden-dapp/src/router.ts
+++ b/raiden-dapp/src/router.ts
@@ -1,12 +1,45 @@
 /* istanbul ignore file */
 import Vue from 'vue';
-import Router from 'vue-router';
+import Router, { Route } from 'vue-router';
+import { Dictionary } from 'vue-router/types/router';
+
 import Home from './views/Home.vue';
 import { RouteNames } from '@/route-names';
+import store from './store';
+import { Tokens } from './types';
 
 Vue.use(Router);
 
-export default new Router({
+/**
+ * Checks whether a token parameter is present and whether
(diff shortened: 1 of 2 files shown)


raiden-dapp/src/router.ts
```diff
@@ -64,3 +66,38 @@ export default new Router({
...
+ * Checks whether token network route is accessed and whether
+ * requested token network address is present in store.
+ */
+export const checkTokenNetworkRoute = (
+  name: string,
+  path: string,
+  next: (to?: string | void) => void,
+  tokens: Tokens
+): void => {
+  [RouteNames.SELECT_HUB as String, RouteNames.OPEN_CHANNEL as String].includes(
```
andrevmatos: I believe almost all other routes should be checked. Only `HOME` and `SELECT_TOKEN` don't have a `:token` param. Also, it's not a good practice to test if you're in one (or some) specific route on a global guard, it'd be clearer if you did a guard as a separate function (to reuse) and applied it as per-route guards, or doing a global guard which would redirect if a `token` param is present but not in the tokens list (independently of which route it is).

PR #325, opened 2019-09-08
Refactor contracts usage in sdk

### LINKED ISSUE #324: Refactor contracts usage in the SDK
## Description

This is a wishlist of small refactorings which would make it clearer how the SDK interacts with the contracts, clear the types and improve performance and readability

## Tasks
- [x] `src/types::RaidenContracts` is used only as the type of a cache for the contracts (registry, token, token network). Get rid of it and the private methods for creating these contracts instances and use memoized functions instead, for clarity and performance.
- [x] `src/types::TokenInfo` is used internally only for the token information cache as well. Get rid of it, use a memoized function member. 
- [x] `getPresences$` observable is actually the same for a given `action$` observable, but calling it on each epic actually create separate/equal mappings for the presences. Use a memoized function which returns a shared observable.
- [x] action `tokenMonitored` `first: boolean` member should be replaced by a `fromBlock?: number` of which block the TokenNetwork was deployed, so we scan only from that point instead of since registry deployment. This also makes it consistent with `channelMonitored`.
- [x] Move `raiden-ts/contracts` to `raiden-ts/src/contracts`, for consistency with `abi` and `deployment`, and scoped access (avoid `..` importing things relativety outside `src`). `tsc` [won't copy the contracts interfaces to outDir](https://stackoverflow.com/questions/56018167/typescript-does-not-copy-d-ts-files-to-build), so a postbuild script will be needed.
- [x] Add `chainId` and `registry` address to state, and validate it, to prevent one passing state from other networks to Raiden. It'll be more relevant when we allow one to download and re-upload a database (state, JSON).

andrevmatos:
Closes #324


raiden-ts/package.json
```diff
@@ -9,13 +9,14 @@
...
-    "build": "npm run contracts && tsc -p src/tsconfig.json && tsc -p src/tsconfig.cjs.json",
+    "build": "npm run contracts && tsc -p src/tsconfig.json && tsc -p src/tsconfig.cjs.json && npm run contracts:post",
     "test": "npm run lint && NODE_ENV=development jest --coverage=true",
     "lint": "eslint src/**/*.ts",
-    "clean": "rimraf dist/ dist:cjs/ .coverage/ contracts/ src/abi/ src/deployment/",
-    "contracts:copy": "node ./scripts/copyContracts.js",
-    "contracts:build": "typechain --target=ethers 'src/abi/*.json' --outDir contracts",
-    "contracts": "npm run contracts:copy && npm run contracts:build",
+    "clean": "rimraf dist/ dist:cjs/ .coverage/ src/contracts/ src/abi/ src/deployment/",
+    "contracts:pre": "node ./scripts/copyContracts.js prebuild",
```
nephix: `prebuild` is unecessary because we don't check for that specific argument

raiden-ts/src/raiden.ts
```diff
@@ -89,17 +86,35 @@ export class Raiden {
...
   /**
    * Expose ether's Provider.resolveName for ENS support
    */
   public readonly resolveName: (name: string) => Promise<Address>;
 
   /**
-   * Observable of completed and pending transfers
-   * Every time a transfer state is updated, it's emitted here. 'secrethash' property is unique and
-   * may be used as identifier to know which transfer got updated.
+   * Get token information: totalSupply, decimals, name and symbol
```
nephix: This will be generated by typedoc so no need to include in comments

raiden-ts/src/raiden.ts
```diff
@@ -89,17 +86,35 @@ export class Raiden {
...
+   * name and symbol may be undefined, as they aren't actually part of ERC20 standard, although
+   * very common and defined on most token contracts.
+   *
+   * @param token  address to fetch info from
+   * @returns  token info
    */
-  public readonly transfers$: Observable<RaidenSentTransfer>;
+  public getTokenInfo: (
+    token: string,
+  ) => Promise<{
```
nephix: In `raiden-ts/src/types.ts` we have `TokenInfo` for this

raiden-ts/src/raiden.ts
```diff
@@ -194,6 +199,19 @@ export class Raiden {
 
     this.events$ = action$.pipe(filter(isActionOf(Object.values(RaidenEvents))));
 
+    this.getTokenInfo = memoize(async (token: string) => {
```
nephix: Return-type missing, should probably be `TokenInfo`?

raiden-ts/package.json
andrevmatos: I know. But as it's also accepted, it's here as "documentation", it's clear when it's run. The fact that we don't check it is just an implementation detail, and we could do.

raiden-ts/src/raiden.ts
andrevmatos: Ok, I'll remove and rewrite

raiden-ts/src/raiden.ts
andrevmatos: Trivial/simple type declarations, which are used only once, don't need to have its own export. There's plenty of inline types throughout the code. As the type itself is self-describing, removing it is cleaning the type namespace, the public interface and one explicit schema for us to maintain.

raiden-ts/src/raiden.ts
andrevmatos: We DO have type support. It's just implicit (actually explicit by the declaration of the member above, just not here). `memoize` accepts a generic type, which is a function `(token: string) => Promise<{ ...members }>`, and return this type intersected with a `MemoizedFunction` (which is just an expansion to add the `cache` member and other memoized function additional attributes), which is also of the first type and then assignable to the member. If you e.g. try to rename one of the members of the returned object, you'll see `tsc` will complain the types are incompatible, so we have stricit/full type support here.

raiden-ts/package.json
andrevmatos: Fixed by making it required in `copyContracts.js`

Issue #326, opened 2019-09-10
SDK Configuration


andrevmatos:
## Description

As part of #245, we need to provide a way to pass settings/configuration parameters to the SDK, dynamically.This config object must be made **reactive** to the SDK internals, and where it makes sense, the reactive/up-to-date values should be used and changes on it reflected immediately at runtime, with public methods to updating it. It should be optional, and sane defaults choosen for any missing fields. It should be serializable/validated through `io-ts` codecs. We should put any relevant internal parameter or constants as properties on it, and document them.

## Acceptance criteria
- There's an exposed `RaidenConfig` **codec** and its **type**.
- Raiden's create/constructor creates a default config, but accepts a `Partial<RaidenConfig>` to overwrite fields/members/params on it.
- There's some public method able to receive again a `Partial<RaidenConfig>` and re-set some parameters on the config. We should try to honor these runtime changes whenever possible/makes sense.
- Not all params may always be changeable at runtime through this method (e.g. setting a matrix server after matrix is already running has no effect, as it was already used, we don't need to reconstruct/re-login on matrix to apply this setting), but we need to document well which parameters only make sense at startup, and which one are dynamic and applied at runtime.
- This config (in itself) is NOT saved on state, but some parts or derived data from it may.
- We have good documentation on each param, where it's used, and what it does.
- We keep test coverage over these config facilities.

## Tasks
- [ ] `src/config.ts` module with `RaidenConfig` codec and type
  - [ ] RaidenConfig params which have default values should be required in the type, and the ones which may be undefined, optional.
- [ ] On startup, a default RaidenConfig, with default parameters should be constructed, and then a partial config param is accepted as parameters to (shallow) overwrite any param in the default.
- [ ] This config then goes to a `BehaviorSubject`, which is passed to `RaidenEpicDeps` as `config$: Observable<RaidenConfig>`, and epics and any other code should use this reactive object to access the config params.
- [ ] A `Raiden.config(config: Partial<RaidenConfig>)` method accepts a new partial config object, merge it with the current config and then `next` it to the BehaviorSubject.
- [ ] Some suggestions (from top of the mind) of initial parameters which may go on this config:
  - matrix server (startup)
  - matrix known servers url (startup, used only if matrixServer wasn't provided, to provide a url from where to fetch a yaml list of potential servers to choose)
  - default reveal timeout
  - default settle timeout
  - messages retry delay/backoff strategy

## Investigate
[...]

PR #327, opened 2019-09-10
Added raiden configuration types

nephix:
Tackles #326

### DIFF
--- a/raiden-ts/src/config.ts
+++ b/raiden-ts/src/config.ts
@@ -1,0 +1,25 @@
+import * as t from 'io-ts';
+
+/**
+ * A Raiden configuration object with required parameters and
+ * optional parameters from [[PartialRaidenConfig]].
+ *
+ * @matrixServerLookup Matrix server URL to fetch existing matrix servers from.
+ * After intializing a [[Raiden]] instance, the matrix server can't be changed later on.
+ * @revealTimeout Timeout for secrets to be revealed
+ * @settleTimeout Timeout for channels to be settled
+ */
+export const RaidenConfig = t.readonly(
+  t.intersection([
+    t.type({
+      matrixServerLookup: t.string,
+      revealTimeout: t.number,
+      settleTimeout: t.number,
+    }),
+    t.partial({
+      matrixServer: t.string,
+    }),
+  ]),
+);
+
+export interface RaidenConfig extends t.TypeOf<typeof RaidenConfig> {}
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -1,6 +1,6 @@
 import { Wallet, Signer, Contract } from 'ethers';
 import { AsyncSendable, Web3Provider, JsonRpcProvider } from 'ethers/providers';
-import { Network, ParamType, BigNumber, bigNumberify, BigNumberish } from 'ethers/utils';
+import { Network, BigNumber, bigNumberify, BigNumberish, ParamType } from 'ethers/utils';
 import { Zero } from 'ethers/constants';
 
 import { MatrixClient } from 'matrix-js-sdk';
@@ -10,13 +10,13 @@ import { createEpicMiddleware } from 'redux-observable';
 import { isActionOf } from 'typesafe-actions';
 import { createLogger } from 'redux-logger';
 
-import { debounce, findKey, transform, constant, pick, isEmpty } from 'lodash';
+import { debounce, findKey, transform, constant, memoize, pick, isEmpty } from 'lodash';
 import { Observable, Subject, BehaviorSubject, AsyncSubject, from } from 'rxjs';
 import { first, filter, map, distinctUntilChanged, scan, concatMap } from 'rxjs/operators';
 
-import { TokenNetworkRegistry } from '../contracts/TokenNetworkRegistry';
-import { TokenNetwork } from '../contracts/TokenNetwork';
-import { HumanStandardToken } from '../contracts/HumanStandardToken';
+import { TokenNetworkRegistry } from './contracts/TokenNetworkRegistry';
+import { TokenNetwork } from './contracts/TokenNetwork';
+import { HumanStandardToken } from './contracts/HumanStandardToken';
 
 import TokenNetworkRegistryAbi from './abi/TokenNetworkRegistry.json';
 import TokenNetworkAbi from './abi/TokenNetwork.json';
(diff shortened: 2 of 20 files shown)


raiden-ts/src/raiden.ts
```diff
@@ -358,7 +369,14 @@ export class Raiden {
         `Mismatch between provided account and loaded state: "${address}" !== "${loadedState.address}"`,
       );
 
-    const raiden = new Raiden(provider, network, signer, contracts, loadedState);
+    const raidenConfig: RaidenConfigDefaults = {
+      matrixServerLookup: '',
+      settleTimeout: 60000,
+      revealTimeout: 60000,
+      ...config
```
nephix: These defaults need to be changed

raiden-ts/src/config.ts
```diff
@@ -0,0 +1,32 @@
+import * as t from 'io-ts';
+
+/**
+ * An Raiden configuration object to overwrite the Raiden default configuration.
+ * 
+ * @matrixServer Matrix server to initliaze the [[Raiden]] instance with.
+ * The matrix server can't be changed later on.
+ */
+export const PartialRaidenConfig = t.partial({
```
andrevmatos: This can go inside `export const RaidenConfig`, it isn't used explicitly anywhere, so it should be inlined there.

raiden-ts/src/config.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * A Raiden configuration object with required parameters and
+ * optional parameters from [[PartialRaidenConfig]].
+ * 
+ * @matrixServerLookup Matrix server URL to fetch existing matrix servers from. 
+ * After intializing a [[Raiden]] instance, the matrix server can't be changed later on.
+ * @revealTimeout Timeout for secrets to be revealed
+ * @settleTimeout Timeout for channels to be settled
+ */
+export const RaidenConfig = t.intersection([
```
andrevmatos: Please, make this `t.readonly`

raiden-ts/src/config.ts
```diff
@@ -0,0 +1,32 @@
...
+  t.type({
+    matrixServerLookup: t.string,
+    revealTimeout: t.number,
+    settleTimeout: t.number,
+  }),
+  PartialRaidenConfig,
+]);
+
+export type RaidenConfigDefaults = t.TypeOf<typeof RaidenConfig>;
```
andrevmatos: Please, make it an empty interface inheriting from the type (to not fall into [this pitfall](https://github.com/raiden-network/light-client/wiki/SDK-Development#a-note-about-io-ts-and-type-complexity)), and rename it to the same name of the codec (`RaidenConfig`) for consistency with the rest of the code which uses this pattern.

raiden-ts/src/config.ts
```diff
@@ -0,0 +1,32 @@
...
+export type RaidenConfig = Partial<RaidenConfigDefaults>;
```
andrevmatos: Don't need to export the Partial version, just the normal codec and type. In the few places it'll be needed (raiden factory, reset function), it's actually more descriptive to explicitly have `Partial<RaidenConfig>` there, as it informs the developer it accepts an object containing a subset of the configuration type.

raiden-ts/src/raiden.ts
```diff
@@ -203,6 +205,7 @@ export class Raiden {
     this.deps = {
       stateOutput$: state$,
       actionOutput$: action$,
+      config$: new BehaviorSubject<RaidenConfigDefaults>(config),
```
andrevmatos:
```suggestion
      config$: new BehaviorSubject<RaidenConfig>(config),
```

raiden-ts/src/raiden.ts
```diff
@@ -256,6 +265,7 @@ export class Raiden {
     account: string | number,
     storageOrState?: Storage | RaidenState | unknown,
     contracts?: ContractsInfo,
+    config?: RaidenConfig,
```
andrevmatos:
```suggestion
    config?: Partial<RaidenConfig>,
```

raiden-ts/src/raiden.ts
```diff
@@ -107,6 +108,7 @@ export class Raiden {
     signer: Signer,
     contractsInfo: ContractsInfo,
     state: RaidenState,
+    config: RaidenConfigDefaults
```
andrevmatos:
```suggestion
    config: RaidenConfig
```

raiden-ts/src/raiden.ts
andrevmatos: These timeouts are in blocks, not seconds. Defaults are `settleTimeout: 500`, `revealTimeout: 50` (as per today)

raiden-ts/src/types.ts
```diff
@@ -31,6 +32,7 @@ export interface RaidenContracts {
 export interface RaidenEpicDeps {
   stateOutput$: BehaviorSubject<RaidenState>;
   actionOutput$: Subject<RaidenAction>;
+  config$: BehaviorSubject<RaidenConfigDefaults>;
```
andrevmatos:
```suggestion
  config$: BehaviorSubject<RaidenConfig>;
```

raiden-ts/src/config.ts
```diff
@@ -0,0 +1,25 @@
...
+      revealTimeout: t.number,
+      settleTimeout: t.number,
+    }),
+    t.partial({
+      matrixServer: t.string,
+    }),
+  ]),
+);
+
+export interface RaidenConfig extends t.TypeOf<typeof RaidenConfig> {}
```
andrevmatos: Can you please add these type exports to `src/index.ts` as well? They may be relevant for the outer world

raiden-ts/src/raiden.ts
```diff
@@ -28,7 +28,7 @@ import kovanDeploy from './deployment/deployment_kovan.json';
 import goerliDeploy from './deployment/deployment_goerli.json';
 
 import { ContractsInfo, RaidenContracts, RaidenEpicDeps, TokenInfo } from './types';
-import { ShutdownReason } from './constants';
+import { ShutdownReason, MATRIX_KNOWN_SERVERS_URL } from './constants';
```
andrevmatos: Could you, please, move these constants also to `config.ts`? Maybe directly on a `const defaultConfig: RaidenConfig` export with these values on. Some of the default-constructed config parameters may be dynamic and depend on `create` factory/constructor context, but most of it should be constant and would go well in a single place like `config.ts`

raiden-ts/tests/unit/mocks.ts
```diff
@@ -173,9 +174,16 @@ export function raidenEpicDeps(): MockRaidenEpicDeps {
     return tokenContracts[address];
   };
 
+  const config: RaidenConfig = {
```
andrevmatos: If you go with the `config::defaultConfig`, this could be imported from there

PR #328, opened 2019-09-11
Added SDK configuration

nephix:
Addresses #326


raiden-ts/src/transport/epics.ts
```diff
@@ -94,45 +94,52 @@ const userRe = /^@(0x[0-9a-f]{40})[.:]/i;
...
-          }),
-          map(server => ({
-            server: server.includes('://') ? server : `https://${server}`,
-            setup: undefined,
-          })),
-        );
-      }
+
+      return config$.pipe(
+        first(),
```
nephix: We don't want the matrix server to be changed during runtime, do we accomplish this through `first()`?

raiden-ts/src/transport/epics.ts
```diff
@@ -94,45 +94,52 @@ const userRe = /^@(0x[0-9a-f]{40})[.:]/i;
...
-            return rtts[0].server;
-          }),
-          map(server => ({
-            server: server.includes('://') ? server : `https://${server}`,
-            setup: undefined,
-          })),
-        );
-      }
+
+      return config$.pipe(
```
nephix: Can we simplify this somehow, maybe through `combineLatest(state$, config$)` or something like that?

raiden-ts/src/raiden.ts
```diff
@@ -425,6 +448,20 @@ export class Raiden {
     return this.deps.provider.blockNumber || (await this.deps.provider.getBlockNumber());
   }
 
+  public config(newConfig: Partial<RaidenConfig>) {
+    this.deps.config$
+      .pipe(
+        first(),
```
nephix: We don't need to unsubscribe because of `first()` right?

raiden-ts/src/raiden.ts
```diff
@@ -500,7 +532,11 @@ export class Raiden {
         }),
       )
       .toPromise();
-    this.store.dispatch(channelOpen({ settleTimeout }, { tokenNetwork, partner }));
+
+    this.deps.config$.subscribe(({ settleTimeout }: RaidenConfig) =>
```
nephix: I feel like this is wrong as this is going to dispatch a `channelOpen` every time the config changes?

raiden-ts/src/config.ts
```diff
@@ -0,0 +1,31 @@
...
+  ]),
+);
+
+export const defaultConfig: RaidenConfig = {
+    matrixServerLookup: 'https://raw.githubusercontent.com/raiden-network/raiden-transport/master/known_servers.test.yaml',
+    settleTimeout: 500,
+    revealTimeout: 50,
+}
+
+export interface RaidenConfig extends t.TypeOf<typeof RaidenConfig> {}
```
andrevmatos: Nitpick: the type usually goes right after the codec on the rest of the codebase, and also the `const defaultConfig` uses the type, not the codec, so it'd be better if you move it above the defaultConfig declaration

raiden-ts/src/raiden.ts
andrevmatos: Right, `first` ensures this observable (or actually this pipeline of it) completes on first element (which on BehaviorSubject and ReplaySubject is emitted immediatelly), and a completing observable also unsubscribe any subscription.

raiden-ts/src/raiden.ts
```diff
@@ -425,6 +448,20 @@ export class Raiden {
     return this.deps.provider.blockNumber || (await this.deps.provider.getBlockNumber());
   }
 
+  public config(newConfig: Partial<RaidenConfig>) {
+    this.deps.config$
+      .pipe(
+        first(),
+        map((currentConfig: RaidenConfig) => ({ ...currentConfig, ...newConfig })),
```
andrevmatos: No need to map and then again on subscribe. I think in the subscribe callback is the best place for it, so you can take this map out

raiden-ts/src/raiden.ts
andrevmatos: Yes, it will.
I think the best place to handle the default `settleTimeout` is in the `channelOpenEpic`, so the request (`channelOpen`) action is dispatched with a possibly undefined `settleTimeout` value, to signal it was called without specifying one, and on that epic you get the config and replace the default value before sending the transaction.

raiden-ts/src/raiden.ts
```diff
@@ -475,14 +512,9 @@ export class Raiden {
...
    * @param partner  Partner address
-   * @param settleTimeout  openChannel parameter, defaults to 500
    * @returns  txHash of channelOpen call, iff it succeeded
    */
-  public async openChannel(
-    token: string,
-    partner: string,
-    settleTimeout: number = 500,
-  ): Promise<Hash> {
+  public async openChannel(token: string, partner: string): Promise<Hash> {
```
andrevmatos: Don't remove the `settleTimeout` parameter, one may want to specify a different than default at `openChannel` time. Instead, make it optional, or even better, in an `opts` optional parameter object for the case we want to add other options here in the future (as done in `transfer`):
```suggestion
  public async openChannel(token: string, partner: string, opts?: { settleTimeout?: number; }): Promise<Hash> {
```

raiden-ts/src/transport/epics.ts
```diff
@@ -94,45 +94,52 @@ const userRe = /^@(0x[0-9a-f]{40})[.:]/i;
 export const initMatrixEpic = (
   {  }: Observable<RaidenAction>,
   state$: Observable<RaidenState>,
-  { address, network, signer, matrix$ }: RaidenEpicDeps,
+  { address, network, signer, matrix$, config$ }: RaidenEpicDeps,
 ): Observable<ActionType<typeof matrixSetup>> =>
   state$.pipe(
     first(),
     mergeMap(state => {
```
andrevmatos: You should use `withLatestFrom(config$),` before this:
```suggestion
    withLatestFrom(config$),
    mergeMap(([state, config]) => {
```

raiden-ts/src/transport/epics.ts
andrevmatos: `withLatestFrom` before the top-level `mergeMap`

raiden-ts/src/transport/epics.ts
andrevmatos: `withLatestFrom` there will also ensure it doesn't react on `config$` changes, but instead just use it when an element is already going through the pipeline, in this case, a single/first `state$` (i.e. only once, at startup)

raiden-ts/tests/e2e/raiden.spec.ts
```diff
@@ -172,11 +172,13 @@ describe('Raiden', () => {
     test('tx fail', async () => {
       expect.assertions(1);
       // settleTimeout < min of 500
-      await expect(raiden.openChannel(token, partner, 499)).rejects.toThrow();
```
andrevmatos: `openChannel` is expected to be able to receive a custom `settleTimeout` per-call, besides the default one

nephix: Wasn't aware of this requirement, will fix

raiden-ts/src/raiden.ts
nephix: I guess I missed that one after trying something out and then forgot to remove it, will fix

PR #330, opened 2019-09-13
Use free PFS for mediated transfers

### LINKED ISSUE #245: I want to send payments to any Raiden client reachable free of charge in the network
## User Story
As Chris I want to send payments to any Raiden client reachable free of charge (without paying mediation fees) in the network so that I can safe tokens.

## Acceptance criteria
- PFS is set in a configuration file or/and at the start of the dApp
- Getting the route from a free PFS server
- Only use routes free of charge (see `fee_penalty` in [the Raiden spec](https://raiden-network-specification.readthedocs.io/en/latest/pathfinding_service.html#routing-preferences))
- Clients without a free path won't be reachable
- SDK accepting a configuration object
- We have got a default configuration for each testnet
- We implement a setting screen later
- Default configuration for testnets is available

## Tasks
- [ ] Splitting the issue into separate tasks/issues @andrevmatos 

## Prerequisite
- We need a testnet client with testnet services (we need services working with v0.100.5a0 - Antifragile Crocodile)

CC @Dominik1999 @palango

andrevmatos:
For now, it just validates a direct transfer and return the trivial route. But this architecture allows async requesting a path from the actual PFS.
When finished, this PR implements free PFS usage, closes #245


.circleci/config.yml
```diff
@@ -1,6 +1,6 @@
 version: 2.1
 
-var_1: &docker_image circleci/node:10.13.0-browsers
+var_1: &docker_image circleci/node:12-browsers
```
nephix: Is this related to the updated project dependencies or to performance?

raiden-ts/src/raiden.ts
```diff
@@ -684,13 +706,19 @@ export class Raiden {
...
   public async transfer(
     token: string,
     target: string,
     amount: BigNumberish,
-    options: { paymentId?: BigNumberish; secret?: string; secrethash?: string } = {},
+    options: {
+      paymentId?: BigNumberish;
+      secret?: string;
+      secrethash?: string;
+      metadata?: { readonly routes: readonly { route: readonly string[] }[] };
```
nephix:
```suggestion
      metadata?: Metadata
```

raiden-ts/src/raiden.ts
```diff
@@ -684,13 +706,19 @@ export class Raiden {
    *                - secret  Secret to register, a random one will be generated if missing
    *                - secrethash  Must match secret, if both provided, or else, secret must be
    *                              informed to target by other means, and reveal can't be performed
+   *                - metadata  Used to specify possible routes instead of querying PFS.
```
nephix: Please use
```
<ul>
   <li>...</li>
...
```

because typedoc will strip off all linebreaks etc (e.g. https://lightclient.raiden.network/docs/classes/raiden.html#transfer)

raiden-ts/src/raiden.ts
```diff
@@ -700,35 +728,82 @@ export class Raiden {
...
+      // wait for pathFind response
+      this.action$.pipe(
+        filter(isActionOf([pathFound, pathFindFailed])),
+        first(
+          action =>
+            action.meta.tokenNetwork === tokenNetwork &&
+            action.meta.target === target &&
+            action.meta.value.eq(amount),
+        ),
+        timeout(10e3), // timeout if no corresponding success/error response before 10s
```
nephix: Might make sense to have an `httpTimeout` in `defaultConfig` and also give devs the options to define their own timeout

.circleci/config.yml
andrevmatos: This is related to the `globalThis` usage in tests. It's supported by all current browsers versions, but Node (used in the tests) just picked it up in v12

raiden-ts/src/raiden.ts
andrevmatos: We can't do that: `Metadata` uses the internal, strong/branded types (e.g. `Address`), which either the user need to know about and do the proper validation process, or we expose these "simplified" versions of the types/schemas (in this case, it's like Metadata but using the raw strings instead of Address), and then validate inside, as I did right below, and from there on, it's recognized as actually being `Metadata`. The same happens for all Addresses in the public methods.

raiden-ts/src/raiden.ts
andrevmatos: Ok, but then it doesn't go here, but inside the PFS epic itself. Will do

andrevmatos: Done, included an `httpTimeout` config option

.circleci/config.yml
andrevmatos: Moved to `global` on tests and removed this change. We may upgrade it when dApp gets upgraded as well (got some non-sdk related test errors on the dApp on node12).

raiden-ts/src/raiden.ts
nephix: We use `RaidenChannel` in the dapp, which also uses `Address`. User can validate with `ethers.utils.getAddress()` can't they?

raiden-ts/src/utils/matrix.ts
```diff
@@ -30,22 +32,28 @@ export function yamlListToArray(yml: string): string[] {
...
-    if (resp.status < 200 || resp.status >= 300) throw NaN;
-  } catch (e) {
-    start = NaN; // return will also be NaN
-  }
-  return { server, rtt: Date.now() - start };
+  return defer(() => {
+    const start = Date.now();
+    return fromFetch(url).pipe(
+      map(response => (response.ok ? Date.now() : NaN)),
+      timeout(httpTimeout || Number.MAX_SAFE_INTEGER),
```
nephix: `9007199254740991` might be too optimistic?

raiden-ts/tests/unit/epics/path.spec.ts
```diff
@@ -0,0 +1,334 @@
...
+    partner,
+    settleTimeout,
+    isFirstParticipant,
+    txHash,
+    state,
+    matrixServer,
+    partnerUserId,
+  } = epicFixtures(depsMock);
+
+  const target = '0x0100000000000000000000000000000000000005' as Address,
```
nephix: Might make sense to move this to `epicFixtures()` so we don't accumulate a lot of new test data here over time

raiden-ts/tests/unit/epics/path.spec.ts
```diff
@@ -0,0 +1,334 @@
...
+    targetUserId = `@${partner.toLowerCase()}:${matrixServer}`,
+    openBlock = 121,
+    state$ = new BehaviorSubject(state);
+
+  const fetch = jest.fn(async () => ({
```
nephix:
```suggestion
  global.fetch = jest.fn(async () => ({
```

raiden-ts/tests/unit/epics/raiden.spec.ts
```diff
@@ -75,10 +73,12 @@ describe('raiden epic', () => {
 
   (createClient as jest.Mock).mockReturnValue(matrix);
 
-  (fetch as jest.Mock).mockResolvedValue({
-    ok: true,
-    status: 200,
-    text: jest.fn(async () => `- ${matrixServer}`),
+  Object.assign(global, {
```
nephix:
```suggestion
  global.fetch = ...
```

raiden-ts/tests/e2e/raiden.spec.ts
```diff
@@ -411,55 +412,155 @@ describe('Raiden', () => {
...
+
+        await expect(raiden.transfer(token, target, 23)).resolves.toMatch(/^0x[0-9a-fA-F]{64}$/);
+
+        expect(fetch).toHaveBeenCalledWith(
+          expect.stringMatching(new RegExp(`^${pfs}/.*/${tokenNetwork}/paths$`)),
+          expect.objectContaining({ method: 'POST' }),
+        );
+
+        raiden2.stop();
+      });
```
nephix: I think devs should have the option to get the computed route (e.g. caching it to reuse later, analytics, ...) as this is also something they are going to pay for. 

As you've explained before they can get it from `raiden.transfers$` in `transfer.metadata`. As this is not specifically being mentioned anywhere, it might make sense to include this inside the test case above (not sure, looked most suitable to me)

What do you think?

raiden-ts/src/raiden.ts
```diff
@@ -700,35 +727,81 @@ export class Raiden {
...
+    if (secret && getSecrethash(secret) !== secrethash)
+      throw new Error('Provided secrethash must match the sha256 hash of provided secret');
+
+    const metadata = options.metadata;
+    if (metadata && !Metadata.is(metadata)) throw new Error('Invalid options.metadata');
+
+    return merge(
+      // wait for pathFind response
+      this.action$.pipe(
+        filter(isActionOf([pathFound, pathFindFailed])),
```
nephix: If we fire the `pathFind` action with the intention of only wanting to see the `estimated_fee` for a path, we would run into trouble right? We will probably need a separate action for this?

raiden-ts/src/path/epics.ts
```diff
@@ -0,0 +1,166 @@
...
+                  })),
+                );
+              } else {
+                throw new Error(`PFS disabled and no direct route available`);
+              }
+            }),
+            withLatestFrom(statePresencesConfig$),
+            // validate/cleanup received routes/metadata
+            map(([{ routes: result }, [state, presences]]) => {
+              const routes: { route: readonly Address[] }[] = [],
```
nephix: Should we already include the `estimated_fee` here or will we have a new epic that handles estimation requests?

raiden-ts/src/path/types.ts
```diff
@@ -0,0 +1,22 @@
...
+        t.readonly(
+          t.type({
+            path: t.readonlyArray(Address),
+            estimated_fee: t.number,
+          }),
+        ),
+      ),
+    }),
+    t.partial({
+      feedback_token: t.string,
```
nephix: According to the PFS docs we somehow need to provide feedback about the requested routes with this `feedback_token`, when doing a payment.

Although the docs don't go into detail on how this should be done, do we still need to do something from the SDK side with it?

nephix: According to @palango it's completely optional and the feedback data is not actively used at the moment

raiden-ts/src/raiden.ts
andrevmatos: `RaidenChannel` is an output. There, it's safe to use specific types, as any of them can be assigned to just the primitives as usual. Inputs can't be validated only in the usual way (like `getAddress`) to narrow the type, so if the dApp needs to know about and pass an actual `Address`, they need to import it and use things like `Address.is` or `Address.decode` and then pass it.
If we accept the primitives, we just validate it internally and throw if not valid, and users don't need to know about the specific/branded/internal types, as outputs are safely upcasted.

raiden-ts/src/utils/matrix.ts
andrevmatos: This is just a way to disable the timeout conditionally without actually removing the operator from the pipeline in the case it's not provided (it is in the only place we use, but I didn't want to enforce this on the interface).
The alternative to keep the parameter optional would be something like:
```
let ret$ = fromFetch(...).pipe(first operators);
if (httpTimeout) ret$ = ret$.pipe(timeout(httpTImeout));
return ret$.pipe(tail operators);
```
Which is way uglier IMO, and disabling timeouts using large-enough timeouts is very common pattern.

raiden-ts/tests/unit/epics/path.spec.ts
andrevmatos: We can't. This keeps the type of `global.fetch` as the original type, TS complains about the type not being assignable to the other, and we need to cast it wherever we use it as the mocked function. This creates a local-context variable, which can then be used as actually being the mock, and Object.assign is more complacent on these cases.

raiden-ts/tests/unit/epics/raiden.spec.ts
andrevmatos: Same as above.

raiden-ts/src/raiden.ts
andrevmatos: No. Why would we? The filter below ensures we match only the pathFind/Found/FindFailed, so it's safe to fire them whenever we want and get the response. It just happens that we do always call it on `transfer`, but could be done standalone as well. I hope to create a method to allow querying it oneself as well later.

andrevmatos: The only point I see is that we don't expose `estimated_fee` yet, but that'll be done when we work on the paid PFS.

raiden-ts/src/raiden.ts
nephix: Ok cool make sense, thanks for clarifying

raiden-ts/src/path/epics.ts
andrevmatos: It'll be here, but I expected to expose it only when we work on supporting the paid PFS.

raiden-ts/src/raiden.ts
nephix: > I hope to create a method to allow querying it oneself as well later.

That's what I was expecting, just wanted to confirm

raiden-ts/src/path/types.ts
nephix: I have created the follow up ticket #358

raiden-ts/tests/e2e/raiden.spec.ts
andrevmatos: There's comments on the interface itself, but we can improve that, add tests to show it here, and add more documentation somewhere else. I think we have open issues about improving these dev-facing docs.

andrevmatos: Done, added inspection and comment on test

PR #337, opened 2019-09-18
Fetching local config depending on .env file

nephix:


raiden-dapp/src/services/raiden-service.ts
```diff
@@ -12,14 +12,18 @@ import {
...
   private store: Store<RootState>;
 
-  private static async createRaiden(provider: any): Promise<Raiden> {
+  private static async createRaiden(
+    provider: any,
+    account?: string
+  ): Promise<Raiden> {
     try {
-      return await Raiden.create(provider, 0, window.localStorage);
+      return await Raiden.create(provider, account || 0, window.localStorage);
```
andrevmatos: I think this would be more descriptive if you had `account: string | number = 0` on the method parameter, and then just pass `account` to `create`, so it will default to `0` if omitted but still suport either `string | number`

raiden-dapp/.env.raiden-package
```diff
@@ -0,0 +1 @@
+VUE_APP_RAIDEN_PACKAGE=http://localhost:8080/wizard.json
```
andrevmatos: May you please use a relative url here, like `/wizard.json`, so it'll be agnostic in regard to `hostname:port`

raiden-dapp/public/wizard.json.example
```diff
@@ -0,0 +1,4 @@
+{
+    "INFURA_ENDPOINT": "goerli.infura.io/v3/6d333faba41b4c3d8ae979417e281832",
```
andrevmatos: You need to include the schema, or at least `//` (to use the same schema as current page) in the beginning, or else it's going to be seen as a relative URL and fetched like `http://localhost:8080/goerli.infura.io/v3...`

nephix: This is what is copied from Infura as endpoint, so with the highest usability. We don't fetch from `http://localhost:8080/goerli.infura.io/v3...`

andrevmatos: Then `https://` needs to be hardcoded. `fetch(url)` without schema treats it as relative.

andrevmatos: Oh, I noticed it's prepended on web3-provider. Thank you

PR #339, opened 2019-09-19
Optimizes token loading

### LINKED ISSUE #7: I want to be able to set a payment target, which is transferred as a parameter
## User Story
As Chris, I want to be able to set a payment target, which is transferred as a parameter so that I know who I am going to pay.

### LINKED ISSUE #8: I want to be able to set a payment amount, which is transferred as a parameter
## User Story
As Chris, I want to be able to set a payment amount, which is transferred as a parameter, so that I know which amount I am going to pay.

### LINKED ISSUE #10: I want to do a payment with paramenters and confirm the payment
## User Story
As Chris, I want to do a payment with paramenters and confirm the payment so that the payment cannot be done automatically.

### LINKED ISSUE #62: I want to be able to open a channel to a hub with a transferred parameter
## User Story
As Chris, I want to be able to open a channel to a hub with a transferred parameter so that I can make a deposit.

## Acceptance criteria
- Parameters include token address, hub address and deposit amount
- If parameters are transferred the open channel screen is openend and the parameters are filled in the form
- Form is not submitted (user can change the value)

## Tasks
- [ ] Find LN link -> @christianbrb

### LINKED ISSUE #318: I want to know when the Light Client is connecting
## User Story
When clicking **Connect** on the starting screen of the light client, the connection process can sometimes take quite a bit of time. As of now, user doesn't receive any feedback from the UI and the UI remains active while waiting:

light-client-starting-screen-unresponsive

In that case connecting the light client took roughly 20secs (MetaMask & Goerli test network)

## Acceptance criteria
* TBD
* Replace button with loading spinner?
* Disable button and show loading spinner inside of the button?

## Tasks
- [ ] Design
- [ ] Implementation

kelsos:
Fixes #318 
Fixes #62 
Fixes #10
Fixes #7
Fixes #8


raiden-dapp/src/views/SelectHub.vue
```diff
@@ -24,48 +24,51 @@
...
+    if (!AddressUtils.checkAddressChecksum(address)) {
       this.navigateToHome();
       return;
     }
 
-    let tokenInfo = this.$store.getters.token(token);
-    if (!tokenInfo) {
-      tokenInfo = await this.$raiden.getToken(token);
-    }
+    await this.$raiden.fetchTokenData([address]);
```
andrevmatos: Hmm, straight up calling `fetchTokenData` may be a problem. It'll call `raiden.getTokenInfo` & `raiden.getTokenBalance`, which will then fail if the token isn't in state yet (which happens e.g. on first launch, if the user is redirected directly to a SelectHub route from the Wizard). `$store.state.tokens` is populated by the async fetching of the tokens. I think ideally you should `watch` for a few seconds if the token shows up in your state (from the subscription in the service), and then if so, `fetchTokenData`, or else navigate to home if it times out. Or possibly include this check/watch/timeout directly in service's `fetchTokenData` and handle the rejection here.

raiden-dapp/src/views/Payment.vue
```diff
@@ -161,16 +173,31 @@ export default class Payment extends Vue {
...
+    this.target = getAddress(target);
+
+    const { token: address } = this.$route.params;
+
+    if (!AddressUtils.checkAddressChecksum(address)) {
+      this.navigateToHome();
+      return;
+    }
+
+    await this.$raiden.fetchTokenData([address]);
```
andrevmatos: Same as the SelectHub route comment.

raiden-dapp/src/views/OpenChannel.vue
```diff
@@ -160,33 +167,30 @@ export default class OpenChannel extends Mixins(NavigationMixin) {
...
-    const { token, partner } = this.$route.params;
+    const { token: address, partner } = this.$route.params;
 
-    if (!AddressUtils.checkAddressChecksum(token)) {
+    if (!AddressUtils.checkAddressChecksum(address)) {
       this.navigateToHome();
       return;
     }
 
+    await this.$raiden.fetchTokenData([address]);
```
andrevmatos: And here

raiden-dapp/src/views/Channels.vue
```diff
@@ -92,9 +105,24 @@ export default class Channels extends Vue {
...
+  }
+
+  async created() {
+    const { token: address } = this.$route.params;
+    if (!AddressUtils.checkAddressChecksum(address)) {
+      this.navigateToHome();
+      return;
+    }
+
+    await this.$raiden.fetchTokenData([address]);
```
andrevmatos: And here

raiden-dapp/src/views/Channels.vue
```diff
@@ -55,18 +55,31 @@
...
+import AddressUtils from '@/utils/address-utils';
+import NavigationMixin from '@/mixins/navigation-mixin';
 
 @Component({
   components: { ListHeader, ChannelList },
-  computed: mapGetters(['channels'])
+  computed: {
+    ...mapGetters(['channels']),
+    ...mapGetters({
+      getToken: 'token'
```
andrevmatos: No need to change the pattern here, as it's just nitpick, or in all other components you applied it, but you really don't need to `map*` everything. If your `$store.getters.token`  is used just once (in the token computed/getter, afaics), you'd do better of going directly accessing `$store` there instead of polluting the class namespace with something used only in a single place and easily accessible directly from there.

raiden-dapp/src/views/SelectHub.vue
kelsos: This is not a problem at all during the first lunch due to the way everything is setup.

A route component (`SelectHub`, `Payment` etc) won't be created until the `loadComplete` mutation is executed since the `<router-view>` is not shown until the initialization is done.

Before committing `loadComplete` we await the `getTokenList` and update the tokens with the placeholders already, so by the time the router view is visible and thus the component is created the placeholders are already in the vuex state.

I have also verified that this works by hand.

raiden-dapp/src/views/Channels.vue
kelsos: Sure I will go and change it :)

raiden-dapp/src/views/SelectHub.vue
andrevmatos: This works well if the token on the route being accessed is already in SDK's state, so it gets pulled in by `getTokenList`, but on first run (no state), this method resolves as soon as the first token show up, the rest are lazy-loaded, but if the route is accessing one of the last, it won't be yet on the `getTokenList`, nor the `$store.state.tokens`, and `fetchTokenData` may fail.
But nvm, if it works by hand, I may be missing something that's going synchronous enough there. Thank you

Issue #305, opened 2019-09-03
Users should not be able to open new channels with partners they have unsettled channels with


nephix:
## User Story
As Chris I want to get an error message when I try to open a channel with a partner I have unsettled channel(s) with.

1. `A` opens channel with `B`
2. `A` closes channel with `B`
3. `A` tries to open a new channel with `B`, while `A` and `B` have unsettled channel(s)

## Acceptance criteria
- [ ] Opening a new channel should not be possible from the UI
- [ ] User gets informed about why it's not possible

PR #349, opened 2019-09-26
Preventing user from paying hub if channel is not in open state

nephix:
Tackles #305


raiden-dapp/src/locales/en.json
```diff
@@ -10,7 +10,7 @@
       "no-checksum": "Please enter an address in checksum format. You can convert your address on https://ethsum.netlify.com",
       "empty": "Please enter an address or ENS name",
       "ens-resolve-failed": "Please enter a resolvable ENS name",
-      "invalid-excluded-address": "Please enter a different address or ENS name. You can't use your own address or the address of the token network."
+      "invalid-excluded-address": "Please enter a different address or ENS name, because the address you have entered is not allowed. There can be multiple reasons: <br />1) You can't use your own address <br />2) You can't use the address of the token network <br />3) The channel to this hub address isn't open"
```
kelsos: Checking the screenshot and seeing this text gives me a feeling that we are heading towards the wrong direction with this. We should try to keep the error messages for the validation short and descriptive. Usually, the message should fit into a single like underneath the input element.

nephix: Feel free to create a follow up ticket for this 👍

kelsos: I like the `blocked` approach better. We certainly don't have to aggregate everything. Can you add this as part of this PR?

raiden-dapp/src/locales/en.json
```diff
@@ -10,7 +10,8 @@
       "no-checksum": "Please enter an address in checksum format. You can convert your address on https://ethsum.netlify.com",
       "empty": "Please enter an address or ENS name",
       "ens-resolve-failed": "Please enter a resolvable ENS name",
-      "invalid-excluded-address": "Please enter a different address or ENS name. You can't use your own address or the address of the token network."
+      "invalid-excluded-address": "Please enter a different address or ENS name. You can't use your own address or the address of the token network.",
+      "channel-not-open": "Please enter a hub address you have an open channel with."
```
kelsos: Are we still limited to sending to direct channels? if we support multihop then the message wording might be wrong.

nephix:
> Are we still limited to sending to direct channels? if we support multihop then the message wording might be wrong.

We don't support it in master, so the messaging is correct

PR #361, opened 2019-09-30
Updates dependencies

kelsos:
Updates dapp dependencies and fixes tests


kelsos:
@andrevmatos I kept eslint to 5.x because of warnings

```
vue-eslint-parser@6.0.4 requires a peer of eslint@^5.0.0 but none is installed. You must install peer dependencies yourself.
```

raiden-dapp/tests/unit/utils/stub.ts
```diff
@@ -1,6 +1,7 @@
 export function stub<T>(): T {
   const typeAssertion = <T>{};
   for (const prop in typeAssertion) {
+    // @ts-ignore
```
andrevmatos: Please, don't, if not needed:

raiden-dapp/tests/unit/utils/stub.ts
```diff
@@ -1,6 +1,7 @@
...
     if (typeAssertion.hasOwnProperty(prop)) {
```
andrevmatos: This is the same as:
```suggestion
     if (prop in typeAssertion) {
```
But I actually don't understand the point, can you clarify? First, all `const prop in typeAssertion` already are `in typeAssertion`. But despite the type T you declared, at runtime `typeAssertion` is an empty object, so the for loop is noop. In practice, at runtime this function just return an empty object, and on TypeScript it may be problematic because you _tell_ TypeScript it'll be an object of type T but empty object may not comply with the type, and TS will never know it.

raiden-dapp/src/store.ts
```diff
@@ -100,7 +101,7 @@ const store: StoreOptions<RootState> = {
     channels: (state: RootState) => (tokenAddress: string) => {
       let channels: RaidenChannel[] = [];
       const tokenChannels = state.channels[tokenAddress];
-      if (tokenChannels) {
+      if (tokenChannels && !isEmpty(tokenChannels)) {
```
andrevmatos: This doesn't solve the issue with the empty channels mapping. The issue is in `getters.token` map above, which uses the `state.channels`, not `getters.channels`. When `state.channels` contains something like `{ [token]: {} }`, `getters.tokens` will start returning emptyTokenModel without even the address, even though the address or token data is present in `state.tokens`. You need to either or both make `getters.tokens` behave well when finding a `[token]: {}` empty mapping in `state.channels` or ensure `updateChannels` mutation filter out these empty channels map from the mapping.

raiden-dapp/tests/unit/utils/stub.ts
kelsos: I wouldn't if it was not needed, without it the build is crashing

raiden-dapp/src/store.ts
kelsos: I would prefer to have each reducer do the check, since you already mentioned that `{ [token]: {} }` is a valid structure.

raiden-dapp/tests/unit/utils/stub.ts
kelsos: Theoretically, this was supposed to create a stub object with all the properties set to undefined. In practice what you mentioned happened. I also realized that this was used only in one place so no need for the extra function anymore, so I am removing it since it does not make any sense to keep a function for only one place.

raiden-dapp/tests/unit/utils/stub.ts
kelsos: Removing it. See the comment bellow :)

kelsos:
@andrevmatos any clue as to why this SDK test is failing?

raiden-dapp/src/store.ts
andrevmatos: Then the `getters.tokens` map/reduce over it, instead of over `state.channels`, or the problem I mentioned would stay

kelsos: `getters.tokens` should also be updated to filter out tokens without channels.

raiden-dapp/tests/unit/store.spec.ts
```diff
@@ -221,5 +221,13 @@ describe('store', () => {
         store.getters.channels('0xd0A1E359811322d97991E03f863a0C30C2cF029C')
       ).toEqual([]);
     });
+
+    test('when token has no channels token getter should return an empty array', () => {
+      const channels = {
+        '0xd0A1E359811322d97991E03f863a0C30C2cF029C': {}
+      };
+      store.commit('updateChannels', channels);
+      expect(store.getters.tokens).toEqual([]);
```
kelsos: Before the change, it was returning an empty accumulator instance that had an empty string as an address and only the initial values. This empty address was causing the issues with the $blockie.

raiden-dapp/src/store.ts
andrevmatos: Oh, right. So that should be it. Thank you =)

raiden-dapp/src/store.ts
```diff
@@ -83,24 +85,27 @@ const store: StoreOptions<RootState> = {
...
 
-      return map(flatMap(state.channels), tokenChannels => {
-        const model = reduce(tokenChannels, reducer, emptyTokenModel());
-        const tokenInfo = state.tokens[model.address];
-        if (tokenInfo) {
-          model.name = tokenInfo.name || '';
-          model.symbol = tokenInfo.symbol || '';
-        }
+      return map(
+        filter(flatMap(state.channels), channels => !isEmpty(channels)),
```
nephix: IMO we should be using [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap), [map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map), [reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) and [filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) here instead of importing from `lodash`

PR #365, opened 2019-10-02
Fixes validation when value is set programatically

### LINKED ISSUE #363: Fix validation while token is loading.
image

## Steps to Reproduce
1. Load the payment screen for a specific token `http://localhost:8080/payment/0x3ed0DaEDC3217615bde34FEDd023bC81ae49251B`
2. Press Connect

## Expected Result
- No error message shows

## Actual Result
- `Please use only up to decimals for this token`

Due to to the async nature of the new token information fetching mechanism the validation runs before the token is actually loaded while the decimal value is still undefined.

The validation error should not be there.
As a note, this also creates a problem when using links with query parameters.

kelsos:
Closes #363

### DIFF
--- a/raiden-dapp/src/components/AddressInput.vue
+++ b/raiden-dapp/src/components/AddressInput.vue
@@ -85,9 +85,10 @@ export default class AddressInput extends Mixins(BlockieMixin) {
   hint: string = '';
   errorMessages: string[] = [''];
 
-  created() {
+  mounted() {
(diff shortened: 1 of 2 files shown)


nephix:
@kelsos I didn't check but when connecting to a new hub there might be the same issue when trying to deposit that might need the same fix

kelsos:
@nephix I checked both /connect and /payment and the work properly :)

Issue #246, opened 2019-08-07
I want to send payments to any Raiden client with paying mediation fees


christianbrb:
## User Story
As Chris I want to send payments to any Raiden client with paying mediation fees so that I can send tokens to every client.

## Acceptance criteria
- Send the amount + expected fees provided by PFS + 10%
- The option [findRoutes ](#359) is used to request a path + fees from the LC
- If the user is accepting the path + fees the metadata is constructed and transfered to the LC
- Using the Deposit-Pop up
- Not setting `diversity_penalty` and `fee_penalty`
- Capacity updates are sent to the PFS
- [Workaround ](https://github.com/raiden-network/light-client/pull/360/files#diff-4dcc2ade77ff6c582c536559ed1af73dR824) is rolled back partly

## Tasks
- [x] Finalize tests for #366 
- [x] Implement a function to return the fees
- [x] Implementation in dApp

After #245 

## Docs

Mediation fee specs: https://raiden-network-specification.readthedocs.io/en/latest/mediation_fees.html
PFS Returns: https://raiden-network-specification.readthedocs.io/en/latest/pathfinding_service.html#returns
How to send capacity updates: https://raiden-network-specification.readthedocs.io/en/latest/pathfinding_service.html#capacity-and-fee-updates

christianbrb:
Within #330 
- The LC requests the route (from the PFS) from the Raiden hub to the target, because the light client does not publish channel information to the PFS and therefore (at least with the current PFS design) doesn't get back a route from the LC to the target)
- The route from the LC to the hub is concatenated with the route from the PFS and put together in the metadata field.

This has got the drawback, that the LC doesn't get the fee estimation from the hub.

There are three potential solutions:
1. Publish information from the LC to the PFS
2. Get information about the route from the LC to the target without the need to publish information to the PFS (new feature for the PFS?) 
3. Add a fee estimate for the first hop like the full client is doing it if using local routing (see https://github.com/raiden-network/raiden/pull/4760/files)

@palango @karlb Any other thoughts or proposals?

karlb:
> Get information about the route from the LC to the target without the need to publish information to the PFS (new feature for the PFS?)

I think this is the way to go. But I don't have and idea on the specifics, yet.

christianbrb:
Which route do I get, if I do not set the two optional parameters `diversity_penalty` and `fee_penalty`
From the docs: `If those parameters are omitted, reasonable defaults are chosen.`

[...]

PR #366, opened 2019-10-06
Publish PFS capacity update to global room, request path from self

andrevmatos:
Part of #246

### DIFF
--- a/raiden-ts/src/path/epics.ts
+++ b/raiden-ts/src/path/epics.ts
@@ -16,6 +16,7 @@ import { isActionOf, ActionType } from 'typesafe-actions';
 import { isLeft } from 'fp-ts/lib/Either';
 import { ThrowReporter } from 'io-ts/lib/ThrowReporter';
 import { bigNumberify } from 'ethers/utils';
+import { Zero } from 'ethers/constants';
 
 import { RaidenAction } from '../actions';
 import { RaidenState } from '../state';
@@ -25,13 +26,12 @@ import { messageGlobalSend } from '../messages/actions';
 import { PFSCapacityUpdate, MessageType } from '../messages/types';
 import { signMessage } from '../messages/utils';
 import { channelDeposited } from '../channels/actions';
+import { ChannelState } from '../channels/state';
 import { Address, UInt } from '../utils/types';
 import { losslessStringify } from '../utils/data';
 import { pathFind, pathFound, pathFindFailed } from './actions';
 import { channelCanRoute } from './utils';
 import { PathResults } from './types';
-import { ChannelState } from 'raiden-ts/channels';
-import { Zero } from 'ethers/constants';
 
 /**
  * Check if a transfer can be made and return a set of paths for it.
--- a/raiden-ts/src/path/epics.ts
+++ b/raiden-ts/src/path/epics.ts
@@ -1,4 +1,5 @@
-import { Observable, of, combineLatest, from } from 'rxjs';
+/* eslint-disable @typescript-eslint/camelcase */
+import { Observable, of, combineLatest, from, EMPTY } from 'rxjs';
 import {
   filter,
   mergeMap,
@@ -15,16 +16,23 @@ import { fromFetch } from 'rxjs/fetch';
 import { isActionOf, ActionType } from 'typesafe-actions';
 import { isLeft } from 'fp-ts/lib/Either';
 import { ThrowReporter } from 'io-ts/lib/ThrowReporter';
+import { bigNumberify } from 'ethers/utils';
 
 import { RaidenAction } from '../actions';
 import { RaidenState } from '../state';
 import { RaidenEpicDeps } from '../types';
 import { getPresences$ } from '../transport/utils';
+import { messageGlobalSend } from '../messages/actions';
+import { PFSCapacityUpdate, MessageType } from '../messages/types';
+import { signMessage } from '../messages/utils';
+import { channelDeposited } from '../channels/actions';
 import { Address, UInt } from '../utils/types';
 import { losslessStringify } from '../utils/data';
 import { pathFind, pathFound, pathFindFailed } from './actions';
 import { channelCanRoute } from './utils';
 import { PathResults } from './types';
+import { ChannelState } from 'raiden-ts/channels';
+import { Zero } from 'ethers/constants';
 
 /**
  * Check if a transfer can be made and return a set of paths for it.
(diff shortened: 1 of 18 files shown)


raiden-ts/src/path/epics.ts
```diff
@@ -164,3 +145,71 @@ export const pathFindServiceEpic = (
       );
     }),
   );
+
+/**
+ * Sends a PFSCapacityUpdate to PFS global room on new deposit on our side of channels
```
nephix: `[[PFSCapacityUpdate]]`

raiden-ts/src/transport/epics.ts
```diff
@@ -733,12 +731,12 @@ export const matrixCleanLeftRoomsEpic = (
   );
 
 /**
- * Handles a MessageSendAction and send its message to the first room on queue for address
+ * Handles a messageSend action and send its message to the first room on queue for address
```
nephix: `[[messageSend]]`

raiden-ts/src/transport/epics.ts
```diff
@@ -835,6 +825,51 @@ export const matrixMessageSendEpic = (
     ),
   );
 
+/**
+ * Handles a messageGlobalSend action and send one-shot message to a global room
```
nephix: `[[messageGlobalSend]]`

raiden-ts/src/transport/epics.ts
```diff
@@ -863,14 +898,21 @@ export const matrixMessageReceivedEpic = (
...
+        withLatestFrom(config$),
+        // filter for text messages not from us and not from global rooms
         filter(
-          ({ event, matrix }) =>
+          ([{ matrix, event, room }, config]) =>
             event.getType() === 'm.room.message' &&
+            event.event.content.msgtype === 'm.text' &&
             event.getSender() !== matrix.getUserId() &&
-            event.event.content.msgtype === 'm.text',
+            ![config.discoveryRoom, config.pfsRoom]
```
nephix: We do this quite often in this PR with the same params. Might make sense to have it in its own function somewhere

raiden-ts/tests/unit/utils.spec.ts
```diff
@@ -138,6 +138,36 @@ describe('getEventsStream', () => {
   });
 });
 
+test('patchSignSend', async () => {
```
nephix: Looks like we forgot this earlier :) thanks for adding

raiden-ts/src/path/epics.ts
```diff
@@ -164,3 +145,71 @@ export const pathFindServiceEpic = (
...
+ * @param state$ - Observable of RaidenStates
+ * @returns Observable of messageGlobalSend actions
+ */
+export const pfsCapacityUpdateEpic = (
+  action$: Observable<RaidenAction>,
+  state$: Observable<RaidenState>,
+  { address, network, signer, config$ }: RaidenEpicDeps,
+): Observable<ActionType<typeof messageGlobalSend>> =>
+  action$.pipe(
+    filter(isActionOf(channelDeposited)),
```
nephix: More of a question: Shouldn't we inform the PFS on `channelWithdrawn` events about updated capacities, since these events alter the channel capacity as well? I think we don't fire these yet, but we already listen for them when they happen.

raiden-ts/tests/unit/epics/transport.spec.ts
```diff
@@ -1008,10 +1060,112 @@ describe('transport epic', () => {
...
+      await expect(
+        matrixMessageReceivedUpdateRoomEpic(action$, state$).toPromise(),
+      ).resolves.toEqual(matrixRoom({ roomId }, { address: partner }));
     });
   });
+
+  test('deliveredEpic', async () => {
+    expect.assertions(5);
+
+    const roomId = `!roomId_for_partner:${matrixServer}`,
```
nephix: Maybe move these to fixture generator

raiden-ts/tests/unit/epics/transport.spec.ts
```diff
@@ -1008,10 +1060,112 @@ describe('transport epic', () => {
...
+    // if you pass something different than the accepted messages, no Delivered is replied
+    await expect(
+      deliveredEpic(of(otherAction), EMPTY, depsMock).toPromise(),
+    ).resolves.toBeUndefined();
+  });
+
+  test('matrixMessageGlobalSendEpic', async () => {
+    expect.assertions(5);
+
+    const processed: Processed = {
```
nephix: Maybe move these to fixture generator

raiden-ts/src/path/epics.ts
andrevmatos: But your coment made me realise we should consider the withdraw amount on capacity calculation, so I'll get that on as well

nephix: Thanks for clarifying @andrevmatos

raiden-ts/tests/unit/epics/transport.spec.ts
andrevmatos: Quite often this `Processed` refers to the `message_identifier` of the message its confirming, not sure if it's worth the hassle to change some cases not others.

andrevmatos: Ok, the ones which had custom `message_identifier` were only on the `transfers.spec.ts` suite. All the others, `Processed` messages were used quite simply as a generic, easily composable `Message`, and therefore I created a `processed` fixture to cleanup these constants. Thank you for pointing this out.

raiden-ts/src/transport/epics.ts
```diff
@@ -195,16 +194,20 @@ export const initMatrixEpic = (
...
-            `#raiden_${network.name || network.chainId}_discovery:${getServerName(server)}`,
+          merge(
+            // ensure we joined global rooms
+            ...globalRoomNames(config).map(globalRoom =>
+              from(matrix.joinRoom(`#${globalRoom}:${getServerName(server)}`)),
+            ),
           ),
         ),
-        mapTo({ matrix, server, setup }),
+        toArray(), // wait all promises to complete
```
nephix: More of a question: I think we could simplify using [forkJoin](http://xgrommx.github.io/rx-book/content/observable/observable_methods/forkjoin.html)? Looks like the operator to be closest to `Promise.all()`. No need to change the code, it's just a question.

nephix:
Good job, looks very clean to me!

raiden-ts/src/transport/epics.ts
andrevmatos: forkJoin would work if I made it a higher order observable, an observable which emitted observables, and cared more about the final values. In this case, I just want to wait for everything to finish, and as toArray was already around, went for it. But yet, forkJoin would work as well

Issue #358, opened 2019-09-27
Provide feedback to PFS for routes


nephix:
## Description

Follow up ticket for https://github.com/raiden-network/light-client/pull/330#discussion_r328995823

As a PFS maintainer I want to collect feedback on the routes that PFS provides

## Acceptance criteria
- When processing payments, send the `feedback_token` to PFS' `/feedback` route as explained [here](https://raiden-network-specification.readthedocs.io/en/latest/pathfinding_service.html#post-api-v1-token-network-address-feedback)

christianbrb:
@karlb Are you already using the PFS route feedback within the PFS?

nephix:
I talked to @palango about this a while ago and back then he said it's not used

christianbrb:
@nephix Thanks :)

I am doing issue housekeeping to reduce our inventory. I am looking at the least recently updated issues.

It is a nice to have. If the request is going to be raised by a user or the services team we'll reopen the issue.

karlb:
> @karlb Are you already using the PFS route feedback within the PFS?

Still not in use. We wanted to have it in place so that we can start incorporating route feedback without having to change the clients. Updating a PFS is much easier than updating all clients.

christianbrb:
Ok, make sense :) Thanks for your feedback

christianbrb:
Closing this issue and wait for it to be implemented in the PFS.

PR #384, opened 2019-10-13
Expose PFS estimated fees

andrevmatos:
This PR is based on top of #383 
Part of #246 
It changes the interface with PFS epics from `Metadata` type (which didn't contain fees) to `Paths`, a  schema which contains the (estimated) fees for each route/path it returns, as well as an optional `feedbackToken`, which we don't use right now (but may, see #358).
There's also an interface, `RaidenPaths`, which is derivated from `Paths` but doesn't include custom types, so can be imported and used as public interface export.

Additionally, this PR implements the whole current mediation fee logic:
- introduce signed `Int` type, to allow for negative fees (allowed before https://github.com/raiden-network/raiden/issues/5053)
- the `fee` is persisted in `SentTransfer` state
- make the distinction between `value` of a transfer request (the value intended to reach the target) and `amount`, which is the effective balance proof amount sent to hub, equals `value + fee`
- fee gets added to the lock total amount, `LockedTransfer.fee` field still there but zeroed (before https://github.com/raiden-network/raiden/issues/4991)
- SecretReveal only happens if target informs it received at least the expected `value`


raiden-ts/src/transfers/epics.ts
```diff
@@ -106,26 +106,42 @@ function makeAndSignTransfer(
...
 
-      // assume metadata is valid and recipient is first hop of first route
-      const metadata: Metadata = action.payload.metadata,
-        recipient = metadata.routes[0].route[0];
+      // assume paths are valid and recipient is first hop of first route
+      // compose metadata from it, and use first path fee
+      const metadata: Metadata = {
+          routes: action.payload.paths.paths.map(({ path }) => ({ route: path })),
+        },
+        fee = action.payload.paths.paths[0].fee,
```
nephix: The `action.payload.paths.paths` being used 3 times here looks rather cumbersome, e.g. why the doubled `paths.paths`?

raiden-ts/tests/unit/reducers.spec.ts
```diff
@@ -680,8 +680,9 @@ describe('raidenReducer', () => {
   describe('transfers', () => {
     const secret = keccak256('0xdeadbeef') as Secret,
       secrethash = getSecrethash(secret),
+      fee = bigNumberify(3) as Int<32>,
```
nephix: Reuse from fixture

nephix:
Looks good to me overall, good job!

raiden-ts/src/path/epics.ts
```diff
@@ -81,64 +83,67 @@ export const pathFindServiceEpic = (
...
+                    return decode(PathResults, losslessParse(text));
                   }),
-                  map(({ result }) => ({
-                    routes: result
-                      .slice()
-                      .sort(({ estimated_fee: a }, { estimated_fee: b }) => a - b)
-                      .map(({ path }) => ({ route: path })),
-                  })),
+                  map(
+                    (results: PathResults): Paths => ({
```
nephix: Would be nice for developers if they can still receive the routes already sorted by fees

raiden-ts/src/raiden.ts
```diff
@@ -705,37 +706,35 @@ export class Raiden {
...
     token: string,
     target: string,
-    amount: BigNumberish,
+    value: BigNumberish,
     options: {
       paymentId?: BigNumberish;
       secret?: string;
       secrethash?: string;
-      metadata?: { readonly routes: { readonly route: string[] }[] };
+      paths?: RaidenPaths;
```
nephix: As a developer I would prefer to pass `paths` as an argument consisting of only native, simple JavaScript types

andrevmatos: Yes, that's exactly what this `RaidenPaths` is about, it's a way to expose `Paths` type as simple JS types. It's not a codec, nor include our specialized types. Notice we use `BigNumberish` as just an union which includes all types which can be converted to `BigNumber`: `number | string | UInt8Array | BigNumber`. With the exception of the last, everything else is pure JS types, and accepting `BigNumber` directly is important if the developer has access to them (e.g. by importing from `ethers/utils`, as we do in the dApp).

raiden-ts/src/path/epics.ts
andrevmatos: The routes are generaly sorted by fees. I removed our custom sorting here because the PFS knows a couple of heuristics to better sort this end result, so I decided to pass it through. The cleanup below is mostly to limit to what we can do (e.g. single output node) and/or revalidate in case PFS sent routes which we locally know can't be used anymore, but overall we respect the PFS hints and sorting.

raiden-ts/src/transfers/epics.ts
andrevmatos: The second `paths` is called `result` in the `PathResults`, but I think `paths` is more descriptive since it's an array of objects whose main property is `path`, which is the path as an array of addresses. The outer `paths` is just because we're refering to the passed `Paths`. I.e.: second paths is the array, first paths is the `Paths` object containing all routing information

raiden-ts/tests/unit/reducers.spec.ts
andrevmatos: Hmm, `epicFixtures` are epic, we can't use it here because we don't have epics or depsMocks around. Will need to stay as is for now, we may refactor these constants and fixtures later.

PR #404, opened 2019-10-21
Persist config in state

### LINKED ISSUE #399: Persist config in state
## Description

This is a small architecture change for the SDK config & state. Config is persisted and restored alongside state, so we can avoid changing some config values (e.g. chosen PFS servers) between runs/sessions.
This is needed for #253, so user can check if `pfs` is undefined, and if so, proceed to chose, set and have one persisted in config/state for next runs.

## Tasks
- [x] add a `config: RaideConfig` property to `RaidenState`, which then becomes the single source of truth for the config, ensuring it's persisted as well.
- [x] A default/new config is only created when creating a default state as well.
- [x] `deps.config$` becomes wired to `deps.state$` with `pluck('config')` & `distinctUntilChanged`, as all internal uses already assume an observable which emits immediately its last value when subscribed, so no next on it.
- [x] A `raidenConfigUpdate(Partial<RaidenConfig>)` action pushes config changes to top-level properties to the reducer, to change/store/update the config inside redux.
- [x] readonly config property is exposed directly from `store.state` in the public interface (getter).

andrevmatos:
Fix #399 
Config:
- It's stored & rehydrated from state    
- deps.config$ reflects state.config updates
- Raiden.config -> updateConfig commits an action to redux to update state instead of next'ing new config directly
- Raiden.config becomes getter for current config (from state)
- initial state & default config have their proper factories

### DIFF
--- a/raiden-ts/src/actions.ts
+++ b/raiden-ts/src/actions.ts
@@ -5,6 +5,7 @@ import { pick } from 'lodash';
 import { ActionType, createStandardAction, getType, Action } from 'typesafe-actions';
 import { ShutdownReason } from './constants';
 
+import { RaidenConfig } from './config';
 import * as ChannelsActions from './channels/actions';
 import * as TransportActions from './transport/actions';
 import * as MessagesActions from './messages/actions';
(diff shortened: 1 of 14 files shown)


andrevmatos:
I'm not sure why coverage is wrong after an amend, but it's not -6%, but +0.02% on the #sdk 
Ignore, anyway.

Issue #254, opened 2019-08-08
I want to use a payed PFS service


christianbrb:
## User Story

As Chris I want to use a paid PFS service so that I am able to use any PFS within the network.

## Acceptance criteria
- Tokens are deposited to the UDC (with one function -> see Wizard)
- The user is getting asked how many tokens he wants to get minted
- Tokens are minted on the testnet
- Prompt to ask the user to pay RDN to get the route

## Docs
- https://raiden-network-specification.readthedocs.io/en/latest/service_contracts.html#communication-between-client-and-pfs

## Remarks
- Create a separate issue for buying RDN on the mainnet

christianbrb:
![AA8F8837-EA75-4320-91A0-51AFA71E818D](https://user-images.githubusercontent.com/35398162/62555765-d113db00-b873-11e9-82de-0c3394c64caf.jpeg)

christianbrb:
Outcome of the refinement. @taleldayekh and @christianbrb are going to split the issue

christianbrb:
Split it in stories
1. Show RDN balance and fee when sending the route request
2. Deposit to UDC
3. Mint RDN tokens (testnet)

![grafik](https://user-images.githubusercontent.com/35398162/66743803-042da880-ee7b-11e9-86d4-491c23b22be2.png)

christianbrb:
The following stories are now covering this issue, therefore closing it

    Show fee when sending the routing request #386
    Show UDC balance when sending the routing request #387
    Deposit to UDC #388
    Mint RDN tokens (testnet) #389

PR #405, opened 2019-10-22
Fetch PFS list from ServiceRegistry

### LINKED ISSUE #253: I want to get a list of PFS services from the service registry
## User Story

As Chris I want to get a list of PFS services from the service registry so that I can use any PFS service in the network
See [Raiden Client implementation](https://github.com/raiden-network/raiden/commit/bce57dad6c8d183387d658065b7c0838544a8bfa)

## Acceptance criteria
- At start-up, the service registry is queried
- The first one is stored in the config
- It is an asynchronous operation
- The whole config is stored in the state (#399)

andrevmatos:
Fix #253 

- `config.pfs` can be 3 values:
  - undefined, currently the default, enables fetching pfs list from service registry
  - null, disable PFS if one isn't specified explicitly in `findRoutes` or `transfer` options
  - a string, which will be used as pfs URL to be used
- on the pfs list fetching, ServiceRegistry is monitored, and one can use `Raiden.findPFS` public API to fetch a list of available PFSs
  - in this case, the selected PFS can be passed to `findRoutes`, and this PFS will be used by findRoutes
  - if not specified, `findRoutes` will fetch, sort & select automatically a PFS from the registry

Notice that, currently, only the paid PFSs are registered on-chain, and we don't support paying PFS yet (#254), so findRoutes will fail when the PFS request fails due to the lack of an IOU. Workaround is to set the free PFS explicity with `raiden.updateConfig({ pfs: 'free-pfs' })`, if not set yet, on console.

### DIFF
--- a/raiden-ts/src/path/epics.ts
+++ b/raiden-ts/src/path/epics.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/camelcase */
-import { Observable, of, combineLatest, from, EMPTY, merge, timer } from 'rxjs';
+import { Observable, of, combineLatest, from, EMPTY, merge } from 'rxjs';
 import {
   filter,
   mergeMap,
@@ -15,15 +15,15 @@ import {
   distinctUntilChanged,
   groupBy,
   switchMap,
-  mapTo,
   scan,
   startWith,
   tap,
+  delay,
 } from 'rxjs/operators';
 import { fromFetch } from 'rxjs/fetch';
 import { isActionOf, ActionType } from 'typesafe-actions';
 import { bigNumberify, BigNumber } from 'ethers/utils';
-import { Zero } from 'ethers/constants';
+import { Zero, Two } from 'ethers/constants';
 import { Event } from 'ethers/contract';
 import { isNil } from 'lodash';
 
@@ -288,19 +288,17 @@ export const pfsServiceRegistryMonitorEpic = (
                 groupBy(([service]) => service),
                 mergeMap(grouped$ =>
                   grouped$.pipe(
-                    // switchMap ensures new events for each server (grouped$) picks most recent valid_till
-                    switchMap(([service, valid_till]) =>
-                      valid_till.lt(Math.floor(Date.now() / 1000))
-                        ? // this event already expired
-                          EMPTY
-                        : // ^S-------s| emits service+valid=true, then with valid=false after valid_till
-                          merge(
-                            of({ service, valid: true }),
-                            timer(new Date(valid_till.toNumber())).pipe(
-                              mapTo({ service, valid: false }),
-                            ),
-                          ),
-                    ),
+                    // switchMap ensures new events for each server (grouped$) picks latest event
+                    switchMap(([service, valid_till]) => {
+                      const now = Date.now(),
+                        validTill = valid_till.mul(1000); // milliseconds valid_till
+                      if (validTill.lt(now)) return EMPTY; // this event already expired
+                      // end$ will emit valid=false iff <2^31 ms in the future (setTimeout limit)
+                      const end$ = validTill.sub(now).lt(Two.pow(31))
+                        ? of({ service, valid: false }).pipe(delay(new Date(validTill.toNumber())))
+                        : EMPTY;
+                      return merge(of({ service, valid: true }), end$);
+                    }),
                   ),
                 ),
                 scan(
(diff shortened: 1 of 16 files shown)


raiden-ts/src/path/utils.ts
```diff
@@ -33,3 +44,105 @@ export function channelCanRoute(
...
+    version: t.string,
+  });
+  // if it's an address, fetch url from ServiceRegistry, else it's already the URL
+  const url$ = Address.is(pfsAddrOrUrl)
+    ? from(serviceRegistryContract.functions.urls(pfsAddrOrUrl))
+    : of(pfsAddrOrUrl);
+  return url$.pipe(
+    withLatestFrom(config$),
+    mergeMap(([url, { httpTimeout }]) => {
+      if (!url || !url.includes('://')) throw new Error(`Invalid URL: ${url}`);
```
nephix: Do we need to check for `https` as well or can we expect to only receive https URLs from the contract?

raiden-ts/src/raiden.ts
```diff
@@ -41,18 +42,19 @@ import {
 } from 'rxjs/operators';
 
 import './polyfills';
-import { TokenNetworkRegistry } from './contracts/TokenNetworkRegistry';
-import { TokenNetwork } from './contracts/TokenNetwork';
-import { HumanStandardToken } from './contracts/HumanStandardToken';
-
-import TokenNetworkRegistryAbi from './abi/TokenNetworkRegistry.json';
-import TokenNetworkAbi from './abi/TokenNetwork.json';
-import HumanStandardTokenAbi from './abi/HumanStandardToken.json';
```
nephix: Looks like we don't need `src/abi/` anymore? If yes, might make sense to remove the `abi` related code from `scripts/copyContracts.js:preBuild()` as well unless we need it in the future

nephix:
Good job

kelsos:
@nephix My guess would be that if there is no newer RC or release version than the one we were using during testing then it wouldn't make sense to upgrade the contracts.

raiden-ts/src/path/utils.ts
kelsos: @Dominik1999 could you help us here?

Dominik1999: I'd love to help. Can you elaborate on the question a bit more? Then either me or Karl can answer that, I am sure

raiden-ts/src/raiden.ts
andrevmatos: We still need the abis to build the contracts with typechain. We just don't need it at runtime. Unless typechain gained the ability to parse the bundled contracts.json, we either keep it, or remove it after running typechain, but it still needed.

raiden-ts/src/path/utils.ts
nephix: The question is whether PFS servers are always running on HTTPS and the ServiceRegistry always returns HTTPS URLs or if there is any requirement to support HTTP as well

andrevmatos: We could either require `startsWith('https://')` or add it on our own, seems there's at least one case on the current contracts where only the domain is set (without schema), which is filtered now, but could be useful to be kept instead of filtered out here.

raiden-ts/src/path/epics.ts
```diff
@@ -63,21 +84,46 @@ export const pathFindServiceEpic = (
...
+              } else if (!action.payload.pfs && configPfs === null) {
+                // pfs not specified in action and disabled (null) in config
+                throw new Error(`PFS disabled and no direct route available`);
+              } else {
+                // else, request a route from PFS.
+                // pfs$ - Observable which emits one PFS info and then completes
+                const pfs$ = action.payload.pfs
+                  ? // first, honor action.payload.pfs
+                    of(action.payload.pfs)
+                  : !isNil(configPfs)
```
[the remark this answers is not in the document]
andrevmatos: `isNil` actually test if `null` || `undefined`. It stayed because a previous version used it in some more places, including over result of function call, and to do the double check without the helper I'd need to store it in a variable to test, which would be troublesome in the `if`. Now, it's simple enough to be removed, although I think it's already used somewhere else so wouldn't save space anyway.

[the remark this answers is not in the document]
nephix: `foo == null` also tests `null || undefined` and `isNil` actually uses `==` [under the hood](https://github.com/lodash/lodash/blob/master/isNil.js) as well. Doesn't seem to be used on `lib2` branch if i search for it

nephix:
@andrevmatos can you create a follow up issue or pull request that addresses the feedback, if needed?

raiden-ts/src/path/utils.ts
andrevmatos: Done in #497

raiden-ts/src/path/epics.ts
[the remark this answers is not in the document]
andrevmatos: Done in #497

PR #409, opened 2019-10-23
Fix/secret request too small

### LINKED ISSUE #397: SecretRequest for amount too small
## Steps to Reproduce
1. Start two raiden v0.200.0-rc1 clients
2. Start light client
3. Establish & fund channels between light client <-> 1st raiden client <-> 2nd raiden client
4. Try to pay 2nd raiden client from light client

## Expected Result
- Payment goes through
- No errors in console

## Actual Result
- Payment "gets stuck"
- Warning in console **SecretRequest for amount  too small!**

amount-too-small

## Additional Infos

- Amount typed into light client: `0.000000011000000000`
- Fees displayed in light client: `0.000000000251947751`
- Total Amount in light client: `0.000000011251947751`
- Amount in warning: `0x028fa62a04` is `0.000000010999966212`

I'm passing the amount `0.000000011000000000` to the SDK's `transfer()` method and the route that includes the fee of `0.000000000251947751`.

When trying it with a smaller amount (`0.000000000000000100`) it somehow worked, so it's not reproducible 100% of the time.

Console output:
[localhost-1571298604780.log](https://github.com/raiden-network/light-client/files/3739689/localhost-1571298604780.log)

nephix:
Fixes #397

### DIFF
--- a/raiden-ts/src/path/epics.ts
+++ b/raiden-ts/src/path/epics.ts
@@ -89,7 +89,13 @@ export const pathFindServiceEpic = (
                   }),
                   map(
                     (results: PathResults): Paths =>
-                      results.result.map(r => ({ path: r.path, fee: r.estimated_fee })),
+                      results.result.map(r => ({
+                        path: r.path,
+                        // Add PFS safety margin to estimated fees
+                        fee: r.estimated_fee.mul(state.config.pfsSafetyMargin * 10).div(10) as Int<
+                          32
+                        >,
+                      })),
                   ),
                 );
               } else {
--- a/raiden-ts/src/path/epics.ts
+++ b/raiden-ts/src/path/epics.ts
@@ -52,7 +52,7 @@ export const pathFindServiceEpic = (
         concatMap(action =>
           statePresencesConfig$.pipe(
             first(),
-            mergeMap(([state, presences, { pfs, httpTimeout }]) => {
+            mergeMap(([state, presences, { pfs, httpTimeout, pfsSafetyMargin }]) => {
               const { tokenNetwork, target } = action.meta;
               if (!(tokenNetwork in state.channels))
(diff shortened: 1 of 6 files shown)


raiden-ts/src/path/epics.ts
```diff
@@ -89,7 +89,11 @@ export const pathFindServiceEpic = (
                   }),
                   map(
                     (results: PathResults): Paths =>
-                      results.result.map(r => ({ path: r.path, fee: r.estimated_fee })),
+                      results.result.map(r => ({
+                        path: r.path,
+                        // Add PFS safety margin to estimated fees
+                        fee: r.estimated_fee.mul(pfsSafetyMargin * 10).div(10) as Int<32>,
```
andrevmatos: This will break if receiving a `pfsSafetyMargin` with steps smaller than `0.1`, e.g. `1.125`. Use `Math.round`, floor or other integer-assuring function to ensure an integer is passed to `mul`, and use a bigger multiplier to ensure more granularity (e.g. 1e6 will allow granularity down to 1e-6, it can be too much, but it doesn't add complexity and at least 100 for percentages would be desired IMO):
```suggestion
                        fee: r.estimated_fee.mul(Math.round(pfsSafetyMargin * 1e6)).div(1e6) as Int<32>,
```

andrevmatos:
Or you can do it in your PR, by replacing this assign
https://github.com/raiden-network/light-client/blob/a6b02638850162df9832b9aff580ac6ab5210a7d/raiden-ts/src/raiden.ts#L403-L407
With `lodash.merge`, as so the new `state.config.pfsSafetyMargin` property gets properly defaulted on previous state. It'll be validated anyway right below, so it's safe to do.

raiden-ts/src/path/epics.ts
nephix: Thanks, wasn't sure what kind of values for safety margins are going to be used here. Will fix

PR #427, opened 2019-10-24
Invalidating text input while it's resolving/validating an ens address

### LINKED ISSUE #214: Address input should not be valid when resolving an ens domain
## Description

Address input has a valid state while resolving an ENS domain. This allows the user to press the screen button breaking the flow.

## Acceptance criteria
- 

## Tasks
- [ ]

nephix:
Fixes #214

### DIFF
--- a/raiden-dapp/src/components/AddressInput.vue
+++ b/raiden-dapp/src/components/AddressInput.vue
@@ -6,6 +6,7 @@
       :hint="hint"
       :value="address"
       :error-messages="errorMessages"
+      :rules="isAddressValid"
       :class="{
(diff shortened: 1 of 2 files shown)


nephix:
> It looks good but can we also add a test for this?

I added one, although I don't find it particually useful and would be better suited to test this in integration tests

kelsos:
Sure, but currently we don't have any integration tests right?

nephix:
> Sure, but currently we don't have any integration tests right?

Yes just saying that this doesn't mean the solution is to unit testify all our missing integration tests

PR #428, opened 2019-10-25
Updates the grid system to that of vuetify 2.x

### LINKED ISSUE #334: Upgrade to vuetify 2 grid format
## Description

Vuetify v1.x used a different grid layout system. Vuetify 2.x that we currently use migrated to a grid system that is similar to bootstrap's layout system. `v-flex` and `v-layout` has been replaced with `v-col` and `v-row`.

There is already a plugin that can assist with migration:
https://github.com/vuetifyjs/eslint-plugin-vuetify

You need to add a couple of rules in eslint

```
'vuetify/grid-unknown-attributes': 'error',
'vuetify/no-legacy-grid': 'error',
```

## Acceptance criteria
- `v-flex` and `v-layout` tags are replaced with `v-col` and `v-row`
- UI should look the same.

## Tasks
- [ ]

kelsos:
* Adds legacy grid as an eslint error
* Updates the components to the new grid system

Closes #334

### DIFF
--- a/raiden-dapp/.eslintrc.js
+++ b/raiden-dapp/.eslintrc.js
@@ -17,6 +17,8 @@ module.exports = {
     'no-console': process.env.NODE_ENV === 'production' ? 'error' : 'off',
     'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
     'vuetify/no-deprecated-classes': 'error',
+    'vuetify/grid-unknown-attributes': 'error',
+    'vuetify/no-legacy-grid': 'error',
     'vue/multiline-html-element-content-newline': ["error", {
       "ignoreWhenEmpty": true,
       "ignores": ["pre", "textarea"],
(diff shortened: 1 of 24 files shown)


kelsos:
@nephix would it be possible to also run it locally and verify that everything looks proper on your side?

nephix:
> @nephix would it be possible to also run it locally and verify that everything looks proper on your side?

Sure will do as soon as I have a stashable state with my current changes

nephix:
First peek looked good, although I'm wondering why you don't wrap with `<v-container>` at all as on https://vuetifyjs.com/en/components/grids

Will try to checkout the actual layouts later.

kelsos:
So I am not sure if you are supposed to have on `v-container` per app or wrap each component to its own `v-container`. I would suggest going like this and opening a new issue about figuring out what the best practice is, and if changes are required apply them in a new PR.

PR #430, opened 2019-10-25
Pfs service select

### LINKED ISSUE #386: I want to see how many RDN tokens I have to pay for a routing request
## User Story
As Chris, I want to see how many RDN tokens I have to pay for a routing request so that I know how much it costs me to request a route.

kelsos:
Requires #405
Closes #386


raiden-dapp/src/views/Payment.vue
```diff
@@ -78,21 +78,31 @@
...
             :enabled="valid"
-            @click="findingRoutes = true"
+            @click="serviceSelection = true"
             :text="$t('payment.pay-button')"
             v-on="on"
             class="payment__pay-button"
           ></action-button>
         </template>
+        <v-card class="payment__route-dialog">
+          <pathfinding-services
```
andrevmatos: Please, make sure it also works if sdk's config.pfs isn't on auto. You need to check `raiden.config.pfs`, and if it's **not** undefined, you must skip the pathfinding-services dialog and call `findRoutes` directly without providing a `pfs` option.

andrevmatos: Also, I noticed the routes selection dialog is shown even if there's a single route with `fee=0`, which is the direct route case. You should skip the dialog in this specific case and pass this route to `transfer` directly.

andrevmatos: Would it be better if I just resolved to `null` or something like that on `findPFS` to indicate this case, instead of rejecting and requiring you to inspect config, even if it's public? Should be easy, but it's important to skip the dialog completely if that's the case.

andrevmatos:
Also, I noticed the routes selection dialog is shown even if there's a single route with `fee=0`, which is the direct route case. You should skip the dialog in this specific case and pass this route to `transfer` directly.

raiden-dapp/src/views/Payment.vue
kelsos: Is the config exposed somehow?

kelsos:
@andrevmatos Since the route selection dialog is a different dialog out of the scope of the current PR I would suggest to move the possible improvement in a new issue because that would require a lot of changes. 

Maybe it would make sense getting input from @taleldayekh and @roy-castrillo on how to integrate the two dialogs into a single screen/dialog/flow for a better experience.

nephix:
The related issue #386 has the acceptance criteria
> We have got the workflow to mint RDN and deposit it to the UDC in the "Try out Raiden section"
Should this be included in the same pull request or in a different one?

raiden-dapp/src/views/Payment.vue
andrevmatos: Yes, it is, there's a `config` getter, and you can check if `pfs` is undefined. You can only call `findPFS` if so.

kelsos:
@nephix I am not sure but it would be nice to have a better flow in the Pathfinding selection process.

nephix:
@kelsos I think the acceptance criteria is about steps on minting and depositing RDN was meant to be about how to do it on EtherScan, not in the dApp.

I agree that the current payment process isn't ideal. We decided to take a quick & good enough approach for the payment flow and iterate & improve on it when everything works

raiden-dapp/src/components/FindRoutes.vue
```diff
@@ -102,6 +103,9 @@ export default class FindRoutes extends Vue {
   @Prop({ required: true })
   target!: string;
 
+  @Prop({ required: true })
+  pfs!: RaidenPFS | null;
```
andrevmatos: This propm doesn't seem to be used. If it's not null, it should be passed to `$raiden.findRoutes`, shouldn't it?

nephix: Hm did we somehow break eslint or did it never scream about unused props/vars/...?

PR #433, opened 2019-10-29
New payment screen & token network switch overlay UI

### LINKED ISSUE #391: I want to easily change the token I am transferring
## User Story
As Chris I want to easily change the token I am transferring so that I don't have to go back from the pay screen to the token screen to select a different token.

## Acceptance criteria
- Invision link (@roy-castrillo to check)
- Start screen is still there
- When you hit pay on the start screen the user come to the payment screen
- Clicking on the arrow shows a modal screen (overlay)
- The modal screen shows all connected tokens
- If I select a token the token in the pay screen gets changed

nephix:
Fixes #391

### DIFF
--- a/raiden-dapp/package.json
+++ b/raiden-dapp/package.json
@@ -56,7 +56,7 @@
     "eslint-plugin-jsdoc": "^15.9.5",
     "eslint-plugin-prettier": "^3.1.1",
     "eslint-plugin-vue": "^5.2.3",
-    "eslint-plugin-vue-i18n": "^0.2.3",
+    "eslint-plugin-vue-i18n": "^0.3.0",
     "eslint-plugin-vuetify": "^1.0.0-beta.4",
     "flush-promises": "^1.0.2",
     "jest-junit": "^8.0.0",
--- a/raiden-dapp/src/App.vue
+++ b/raiden-dapp/src/App.vue
@@ -116,12 +116,11 @@ export default class App extends Vue {
   left: 0;
   right: 0;
   width: 220px;
-  margin-left: auto;
-  margin-right: auto;
-  padding: 0 0 27px;
+  margin: 27px auto 27px auto;
 
   a {
     color: #646464;
+    text-decoration: none;
   }
 }
 </style>
--- a/raiden-dapp/src/components/AppHeader.vue
+++ b/raiden-dapp/src/components/AppHeader.vue
@@ -90,10 +90,9 @@
 <script lang="ts">
 import { Component, Mixins } from 'vue-property-decorator';
 import { mapGetters, mapState } from 'vuex';
-
 import BlockieMixin from '@/mixins/blockie-mixin';
-import NavigationMixin from '@/mixins/navigation-mixin';
 import { RouteNames } from '@/route-names';
+import NavigationMixin from '@/mixins/navigation-mixin';
 
 @Component({
   computed: {
--- a/raiden-dapp/src/components/AppHeader.vue
+++ b/raiden-dapp/src/components/AppHeader.vue
@@ -90,9 +90,10 @@
 <script lang="ts">
 import { Component, Mixins } from 'vue-property-decorator';
 import { mapGetters, mapState } from 'vuex';
+
 import BlockieMixin from '@/mixins/blockie-mixin';
-import { RouteNames } from '@/route-names';
 import NavigationMixin from '@/mixins/navigation-mixin';
+import { RouteNames } from '@/route-names';
 
 @Component({
   computed: {
--- a/raiden-dapp/src/components/TokenOverlay.vue
+++ b/raiden-dapp/src/components/TokenOverlay.vue
@@ -7,9 +7,9 @@
         </v-btn>
       </v-row>
 
-      <v-row>
+      <v-row id="connectNew">
         <v-col cols="12">
-          <v-list class="token-list">
+          <v-list class="item-list">
             <v-list-item @click="navigateToTokenSelect()">
               <v-col cols="2">
                 <v-list-item-avatar>
@@ -33,9 +33,9 @@
         </v-col>
       </v-row>
 
-      <v-row>
+      <v-row id="tokenList">
         <v-col cols="12">
-          <v-list v-for="(token, i) in tokens" :key="i" class="token-list">
+          <v-list v-for="(token, i) in tokens" :key="i" class="item-list">
             <v-list-item
               :key="token.address"
               :to="`/payment/${token.address}`"
@@ -145,7 +145,7 @@ export default class Payment extends Mixins(BlockieMixin, NavigationMixin) {
     margin: 15px;
   }
 
-  .token-list {
+  .item-list {
     height: 100%;
     background-color: transparent !important;
     padding-bottom: 0;
--- a/raiden-dapp/src/components/TokenOverlay.vue
+++ b/raiden-dapp/src/components/TokenOverlay.vue
@@ -2,14 +2,14 @@
   <v-overlay :value="show" absolute opacity="1.0" class="token-network-overlay">
     <v-container class="container">
       <v-row no-gutters justify="end">
-        <v-btn @click="cancel" icon class="close-button">
+        <v-btn @click="cancel" icon class="token-network-overlay__close-button">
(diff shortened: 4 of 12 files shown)


raiden-dapp/package.json
```diff
@@ -56,7 +56,7 @@
     "eslint-plugin-jsdoc": "^15.9.5",
     "eslint-plugin-prettier": "^3.1.1",
     "eslint-plugin-vue": "^5.2.3",
-    "eslint-plugin-vue-i18n": "^0.2.3",
+    "eslint-plugin-vue-i18n": "^0.3.0",
```
nephix: Updated manually in favor of https://github.com/kazupon/eslint-plugin-vue-i18n/issues/25 so we can ignore `v-icon` tags for the `vue-i18n/no-raw-text` rule

raiden-dapp/src/App.vue
```diff
@@ -116,12 +116,11 @@ export default class App extends Vue {
...
   width: 220px;
-  margin-left: auto;
-  margin-right: auto;
-  padding: 0 0 27px;
+  margin: 27px auto 27px auto;
 
   a {
     color: #646464;
+    text-decoration: none;
   }
```
nephix: Random diplay fix for privacy policy link display, unrelated to ticket

raiden-dapp/src/components/AppHeader.vue
```diff
@@ -90,9 +90,10 @@
 <script lang="ts">
 import { Component, Mixins } from 'vue-property-decorator';
 import { mapGetters, mapState } from 'vuex';
+
 import BlockieMixin from '@/mixins/blockie-mixin';
-import { RouteNames } from '@/route-names';
 import NavigationMixin from '@/mixins/navigation-mixin';
+import { RouteNames } from '@/route-names';
```
kelsos: Just curious was there any reason this was moved, or it was don by the editor?

raiden-dapp/src/components/TokenOverlay.vue
```diff
@@ -0,0 +1,188 @@
+<template>
+  <v-overlay :value="show" absolute opacity="1.0" class="token-network-overlay">
+    <v-container class="container">
+      <v-row no-gutters justify="end">
+        <v-btn @click="cancel" icon class="close-button">
+          <v-icon>mdi-close</v-icon>
+        </v-btn>
+      </v-row>
+
+      <v-row id="connectNew">
```
kelsos: Would it be possible to remove camelCase from the id? usually so far in templates, camelCase is either used for a variable or a method.

raiden-dapp/src/components/TokenOverlay.vue
```diff
@@ -0,0 +1,188 @@
...
+      <v-row>
+        <v-col cols="2" align-self="center"></v-col>
+        <v-col cols="10" align-self="center" class="header">
+          {{ $t('tokens.connected.header') }}
+        </v-col>
+      </v-row>
+
+      <v-row id="tokenList">
+        <v-col cols="12">
+          <v-list v-for="(token, i) in tokens" :key="i" class="item-list">
```
kelsos: in most of the components, we follow http://getbem.com/ naming.  Can we make sure that this complies to the same naming scheme for consistency?

raiden-dapp/src/components/TokenOverlay.vue
```diff
@@ -0,0 +1,188 @@
...
+          <v-list class="item-list">
+            <v-list-item @click="navigateToTokenSelect()">
+              <v-col cols="2">
+                <v-list-item-avatar>
+                  <v-btn class="mx-2" fab dark small color="primary">
+                    <v-icon dark large>mdi-plus</v-icon>
+                  </v-btn>
+                </v-list-item-avatar>
+              </v-col>
+              <v-col cols="10" align-self="center" class="connect-new-token">
```
kelsos: naming consistency like below

raiden-dapp/src/components/icons/DownArrow.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+  <svg
+    width="11.5px"
+    height="5.5px"
+    viewBox="0 0 18 9"
+    version="1.1"
+    xmlns="http://www.w3.org/2000/svg"
+    xmlns:xlink="http://www.w3.org/1999/xlink"
+  >
```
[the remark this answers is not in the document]
nephix: It's necessary to access the SVG's DOM elements and style them via CSS

raiden-dapp/src/components/TokenOverlay.vue
nephix: Any specific reason we use BEM? Seems super verbose and get's ugly quickly when nested in SASS. But will change

kelsos: Sure the main reason is clarity and uniqueness of the elements when we are editing CSS. Sure it is verbose but that is the point. While the component scoping might work for the page itself, it's harder to figure out which rule goes where. This becomes even more obvious when you have a lot of components and maybe also some global style in the mix using the same class. So far I am happier using BEM that I was dealing with CSS before introducing it.

nephix: Yeah but the verboseness and DOM readability could easily be improved by using CSS modules. That way you get the best of both worlds without having any compromises. BEM solves the global CSS scope problem but has a lot of drawbacks. I will do the change for now but we should rethink this in the future

kelsos:
Could you mention a few of the drawbacks compared to BEM?

For a start, I would like to mention that I don't consider the verbosity a
drawback in any case. On the contrary actually prefer the verbosity. It
makes it clear for me, and for the editor on what CSS rule I am editing or
renaming, etc. Maybe CSS modules could work with if they have proper editor
support, but after a quick look at examples with CSS modules and I am not
convinced that it is the best or even a better approach, sorry.

nephix:
> Could you mention a few of the drawbacks compared to BEM? For a start, I would like to mention that I don't consider the verbosity a drawback in any case. On the contrary actually prefer the verbosity. It makes it clear for me, and for the editor on what CSS rule I am editing or renaming, etc. Maybe CSS modules could work with if they have proper editor support, but after a quick look at examples with CSS modules and I am not convinced that it is the best or even a better approach, sorry.

Have to strongly disagree here.

1. File size: BEM files, even minified, are a magnitude larger than scoped CSS modules
2. Cluttered templates: BEM CSS selectors clutter templates and therefore decrease readability and maintainability. This again gets worse by magnitudes if you begin add multiple BEM classes to a DOM node
3. Global modifiers can't be inherited as they're part of the BEM class name
4. BEM discourages usage of tag selectors, a lot of potential down the toilet

I could go on but just leave it for now. I think BEM had it's time and place in like early 2010s but since then better approaches, like CSS modules, have emerged that eliminated all the drawbacks

kelsos:
1. Generated CSS names from modules to the examples look quite similar so I
don't really see the magnitude here, more like a small percentage.
2. Why would you add multiple BEM classes to a DOM node? And I am not
really sure what you mean by clutter?
3. Global modifiers, could you please clarify?
4. Do you have a specific example in this project where this happened to
you and BEM has an inhibiting factor?

Like I said before I am still not convinced.

nephix:
Don't really want to bother this pull request more with this debate. My points were meant for you to do your own research and not me trying to convince you or trigger an endless back-and-forth discussion. 

If it's more important to "win" a discussion by turning draw backs into supposed-to-be-advantages and challenging all advantages of the opposition instead of finding the best solution for the project, then so be it

kelsos:
@nephix I will continue looking into it, so your points were successful in triggering that if it was not clear. Thank you very much for that.

I have to apologize for the back and forth, and challenging everything, no it's not about "winning", but at this stage of the project, I also take into consideration the effort required to do something, along with the potential benefit over the way things are currently done. 

It is my personal opinion, but I always think of the best solution in regards to the problems we are currently facing.

Anyway, thank you for starting this debate it was a starting point, and sorry If I come along as stubborn.

nephix:
ty @kelsos and it didn't come across stubborn and no need to apologize. Technical debates can be heated as long as we stay open minded 👍

PR #446, opened 2019-10-30
Fix/eslint no unused vars

nephix:
This pull requests adds eslinting for unused imports and parameters in the Dapp, to prevent them e.g. in #430 

For strict function signatures, we force the eslint rule ignore function parameters starting with a `_`, e.g. `function(a: string, _b: string)`

For `raiden-ts` this already works so no adjustments necessary


kelsos:
@nephix :D let me know when this will be ready for review, that is a great addition :)

nephix:
@kelsos yes I thought so 😅

PR #472, opened 2019-11-01
Updates to vue-cli 4.x

kelsos:


raiden-dapp/vue.config.js
```diff
@@ -9,8 +9,6 @@ module.exports = {
         .pre()
         .use('source-map-loader')
         .loader('source-map-loader');
-
-      config.devtool = 'eval-source-map';
```
kelsos: Moved it to `configureWebpack`

raiden-dapp/tsconfig.json
```diff
@@ -32,6 +32,9 @@
       "rxjs/*": [
         "node_modules/rxjs/*"
       ],
+      "ethers": [
+        "node_modules/ethers/"
+      ],
```
kelsos: @andrevmatos I spend a day trying to figure out why the build would fail. That is the reason.

raiden-dapp/package.json
```diff
@@ -26,27 +26,30 @@
...
-    "vue-property-decorator": "^8.2.2",
+    "vue-property-decorator": "^8.3.0",
     "vue-router": "^3.1.3",
     "vue-virtual-scroller": "^1.0.0-rc.2",
     "vuetify": "2.1.9",
     "vuex": "^3.1.1"
   },
   "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^2.6.0",
+    "@typescript-eslint/parser": "^2.6.0",
```
kelsos: `@vue/eslint-config-typescript` depends on a really old version of this. This causes the warnings we get about unsupported versions. I did a quick check and everything looks ok. I am currently adding the dependency here, but we should keep an eye on the `@vue/eslint-config-typescript` and maybe remove the dependency.

kelsos:
I also moved the store.ts/routes.ts to their new [folders](https://cli.vuejs.org/migrating-from-v3/#the-global-vue-cli).

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -170,7 +169,6 @@ export default class RaidenService {
         address: tokenAddress
       };
     } catch (e) {
-      console.error(e);
```
nephix: The deletion of `console.error()` above makes sense since we push an error into the store instead. But here it might make sense to have console output if something goes wrong

kelsos: I would argue that it would be better to enhance the error message with an expandable area (show error message) and show it in the login screen. Instead of logging it in the console.

nephix: This would be better, but having an error message in the console as having nothing (which is the current state)

Issue #400, opened 2019-10-18
Set Total Deposit fails


christianbrb:
## Steps to Reproduce
1. Connect to https://lightclient.raiden.network/channels/0x3ed0DaEDC3217615bde34FEDd023bC81ae49251B on Goerli
2. Deposit 0.000000000000000021 in the channel to 0x456e2459b6E01FF30375D59f24d359655bEB6754
3. The first transaction is successful (see https://goerli.etherscan.io/tx/0x642c10aa82db5ced1d1a20a6d389c1ad9db3d681613bff2a555c18d68fc137f1)

## Expected Result
- The second transaction is successful and there are more tokens deposited to the channel

## Actual Result
- The second transaction (Set Total Deposit) fails (see https://goerli.etherscan.io/tx/0xc82ed9771ab83c64fe5bed84b47f278f4752812192473a7a7e723e8cc23f5971)

christianbrb:
Set Total Deposit when opening a channel to a new hub (0x5257964Ef9b81fba7276AF2a97c111AaD7B840D6) workded

kelsos:
@christianbrb here are the findings of my investigation.

While it was not easy because I needed an archive node in order to debug the reverted transaction, we are lucky because it is easy to sync a new Goerli archive node :D

After that, I used [Remix](https://remix.ethereum.org/) to debug the failing transaction and I found out that the transaction reverts at the following line:

https://github.com/raiden-network/raiden-contracts/blob/a8af89a663770d6cf4312017c684293f8b144e91/raiden_contracts/data/source/raiden/TokenNetwork.sol#L349

The two failing transactions attempted to set the total deposit to:

## Transaction 1
[0x7aa280944857eeed6d8005cc3104ce3c33ec482a5c52fa11e0378328ee86d7e1](https://goerli.etherscan.io/tx/0x7aa280944857eeed6d8005cc3104ce3c33ec482a5c52fa11e0378328ee86d7e1)@`1490832`

```
Function: setTotalDeposit(uint256 channel_identifier, address participant, uint256 total_deposit, address partner) ***

MethodID: 0x63ea0143
[0]:  0000000000000000000000000000000000000000000000000000000000000024
[1]:  000000000000000000000000b84c6bfba1108543ea47e8b9a669cb39ffc70f6a
[2]:  0000000000000000000000000000000000000000000000000000000000000064
[3]:  000000000000000000000000456e2459b6e01ff30375d59f24d359655beb6754
```

[...]

PR #497, opened 2019-11-05
SDK fixes for events fetching

andrevmatos:
- [x] Remove `toBlock$` from `getEventsStream`, ensure intervals don't interesect nor are disjunct, to avoid losing or getting duplicated events
- [x] Fix for long standing TODO to use a single event filter per channel, also fixes ordering of initial event fetching
- [x] Fix small issue with memoized `getTokenInfo` accessing wrong `this` and failing on some edge scenarios
- [x] Remove dependency of lodash's `isNil`, introduced in #405 , replace with `!= null`
- [x] Better validation of PFS's registered URL, accept domain-only (without schema, defaulting to `https://`), require `https://` otherwise.
- [x] Add some logs for message signature prompts.

This also ensures #400 doesn't happen anymore, as the most probable cause of it was some previous deposit event being missed.

### DIFF
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -13,7 +13,6 @@ import {
   first,
   publishReplay,
   switchMap,
-  pluck,
 } from 'rxjs/operators';
 import { isActionOf, ActionType } from 'typesafe-actions';
 import { findKey, get, isEmpty, negate } from 'lodash';
@@ -95,7 +94,6 @@ export const initMonitorRegistryEpic = (
               isEmpty(state.tokens)
                 ? of(contractsInfo.TokenNetworkRegistry.block_number)
                 : undefined,
-              isEmpty(state.tokens) ? state$.pipe(pluck('blockNumber')) : undefined,
             ).pipe(
               withLatestFrom(state$),
               map(([[token, tokenNetwork, event], state]) =>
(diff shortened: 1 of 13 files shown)


raiden-ts/src/channels/epics.ts
```diff
@@ -289,145 +282,145 @@ export const channelMonitoredEpic = (
...
+            [mergedFilter],
+            // if channelMonitored triggered by ChannelOpenedAction,
+            // fetch Channel's pastEvents since channelOpened blockNumber as fromBlock$
+            action.payload.fromBlock ? of(action.payload.fromBlock) : undefined,
+          ).pipe(
+            mergeMap(function*(data) {
+              if (isEvent<ChannelNewDepositEvent>(depositFilter, data)) {
+                yield channelDeposited(
+                  {
+                    id: data[0].toNumber(),
```
nephix: Maybe worth destructuring for better readability:
```
const [id, participant, transaction, block, ...] = data;
```

raiden-ts/src/raiden.ts
```diff
@@ -152,6 +152,7 @@ export class Raiden {
    * @returns token info
    */
   public getTokenInfo: (
+    this: Raiden,
```
nephix: Can't we make this a `function` so we have access to `this`? If not, typedoc needs to be updated with the new param

raiden-ts/src/utils/ethers.ts
```diff
@@ -39,62 +38,73 @@ export function fromEthersEvent<T>(
...
-  lastSeenBlock$?: Observable<number>,
 ): Observable<T> {
   const provider = contract.provider as JsonRpcProvider;
 
+  const logToEvent = (log: Log): T | undefined => {
+    // parse log into [...args, event: Event] array,
+    // the same that contract.on events/callbacks
+    const parsed = contract.interface.parseLog(log);
+    if (!parsed) return;
+    const args = Array.prototype.slice.call(parsed.values);
```
nephix: Maybe `parsed.values.slice()`

andrevmatos: No, this code came from `ethers`, `values` isn't an actual array, but an object with numeric keys and string keys for the parameters and a `length` property, so it is _Array-like_. This `Array.prototype.slice.call` is a way to construct an actual array from this object.

raiden-ts/src/raiden.ts
andrevmatos: First, no, because this isn't a method, but a member function, which is assigned in the constructor instead of lying on the prototype. `this` isn't a normal parameter either, but just a special TS syntax to declare the type of `this` constructor for these free-range functions, they don't get docs because of this.

raiden-ts/src/channels/epics.ts
andrevmatos: Would need to do separately each block for all parameters, do you think it's worth it if they're used only once?

nephix: Only for readability. Feel free to ignore if you think it's not worth it

Issue #389, opened 2019-10-14
I want to mint RDN tokens on the testnet


christianbrb:
## User Story
As Chris, I want to mint RDN tokens on the testnet so that I can use a paid PFS service.

## Acceptance criteria
- I see my ETH balance
- I have got a field to enter the amount of SVT / RDN to mint
- 1 SVT / RDN as default value
- I access this screen from the Select PFS screen with a button called Deposit
- Minting functionality is exposed from the SDK
- SDK function should be able to mint any token

![grafik](https://user-images.githubusercontent.com/35398162/67951925-bcce4880-fbec-11e9-94c5-82c7990087bb.png)

taleldayekh:
RDN test tokens are used on the testnet it is only one token but it changes whenever new contracts are deployed.

nephix:
I'm looking into how to mint RDN and exposing a function from the SDK

nephix:
SDK function `mint()` has been implemented

nephix:
The rest should be implemented when #402 is ready

kelsos:
The minting part of this issue was already done by Thomas. I am will create a PR to finalize the SDK part of the issue by exposing a deposit API. 

After getting the SDK part merged I will continue on the dApp integration part.

kelsos:
All the PRs related to this issue have been merged to master so I am closing this issue as it is considered done.

PR #513, opened 2019-11-06
Added token minting feature to SDK

nephix:
SDK part of #389

### DIFF
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -971,7 +971,7 @@ export class Raiden {
 
     // Check whether we are on a test network
     const { name } = await this.deps.provider.getNetwork();
-    if (!['goerli', 'rinkeby', 'ropsten'].includes(name)) {
+    if (name === 'homestead') {
       throw new Error('Minting is only allowed on test networks.');
     }
 
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -1,4 +1,3 @@
-import { CustomTokenFactory } from './contracts/CustomTokenFactory';
 import { Signer, ContractTransaction } from 'ethers';
 import { Wallet } from 'ethers/wallet';
 import { AsyncSendable, Web3Provider, JsonRpcProvider } from 'ethers/providers';
@@ -47,6 +46,7 @@ import { TokenNetworkRegistryFactory } from './contracts/TokenNetworkRegistryFac
 import { TokenNetworkFactory } from './contracts/TokenNetworkFactory';
 import { HumanStandardTokenFactory } from './contracts/HumanStandardTokenFactory';
 import { ServiceRegistryFactory } from './contracts/ServiceRegistryFactory';
+import { CustomTokenFactory } from './contracts/CustomTokenFactory';
 
 import ropstenDeploy from './deployment/deployment_ropsten.json';
 import rinkebyDeploy from './deployment/deployment_rinkeby.json';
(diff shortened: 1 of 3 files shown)


raiden-ts/src/raiden.ts
```diff
@@ -955,6 +956,29 @@ export class Raiden {
...
+   * @param amount - Amount to be minted
+   * @returns transaction
+   */
+  public async mint(tokenAddress: string, amount: BigNumberish): Promise<ContractTransaction> {
+    // Check whether address is valid
+    if (!Address.is(tokenAddress)) throw new Error('Invalid address');
+
+    // Check whether we are on a test network
+    const { name } = await this.deps.provider.getNetwork();
+    if (!['goerli', 'rinkeby', 'ropsten'].includes(name)) {
```
BOR4: If ever 4th testnet comes to life code won't work. Why don't you just look for mainnet and then it will always work?

nephix: The SDK will fail much earlier if a new test net comes along. But I guess it's worth having to change one part less of the code in that case. Will fix

raiden-ts/src/raiden.ts
```diff
@@ -955,6 +956,29 @@ export class Raiden {
...
+   * @param amount - Amount to be minted
+   * @returns transaction
+   */
+  public async mint(tokenAddress: string, amount: BigNumberish): Promise<ContractTransaction> {
+    // Check whether address is valid
+    if (!Address.is(tokenAddress)) throw new Error('Invalid address');
+
+    // Check whether we are on a test network
+    const { name } = await this.deps.provider.getNetwork();
+    if (name === 'homestead') {
```
andrevmatos: There exists `this.deps.network`, use it to avoid the uneeded async call here, please
```suggestion
    if (this.deps.network.name === 'homestead') {
```

raiden-ts/src/raiden.ts
```diff
@@ -955,6 +956,29 @@ export class Raiden {
...
+
+  /**
+   * Mints the amount of tokens of the provided token address.
+   * Throws an error if executed on main net.
+   *
+   * @param tokenAddress - Address of the token to be minted
+   * @param amount - Amount to be minted
+   * @returns transaction
+   */
+  public async mint(tokenAddress: string, amount: BigNumberish): Promise<ContractTransaction> {
```
andrevmatos: Please, avoid forcing dependency of ethers specific type here. For consistency, other transactioning methods return `Hash`, which is assignable to normal `string` containing txHash. Also for consistency, the parameter is usually called just `token`, as public interface only deals with addresses anyway.
```suggestion
  public async mint(token: string, amount: BigNumberish): Promise<Hash> {
```

raiden-ts/src/raiden.ts
```diff
@@ -955,6 +956,29 @@ export class Raiden {
...
+
+    // Check whether we are on a test network
+    const { name } = await this.deps.provider.getNetwork();
+    if (name === 'homestead') {
+      throw new Error('Minting is only allowed on test networks.');
+    }
+
+    // Mint token
+    const customTokenContract = CustomTokenFactory.connect(tokenAddress, this.deps.signer);
+    return customTokenContract.functions.mint(decode(UInt(32), amount));
```
andrevmatos: Please, await on the call, then await on tx to be mined, it's useful to ensure throwing early if tx fails:
```suggestion
    const tx = await customTokenContract.functions.mint(decode(UInt(32), amount));
    const receipt = await tx.wait();
    if (!receipt.status) throw...
    return tx.hash;
```

nephix:
Thanks for the input guys. Everything was addressed

nephix:
I've added it to `RaidenService` of the dapp as well

nephix:
> this patch coverage is still a little bit low, seems some lines are not hit/tested, but is not a big deal, address it or merge as is if you want, code itself seems solid. thank you for this

It's expected, as there're no direct tests of `RaidenService` functions. Coverage will go up once we have unit tests that trigger the actual minting. The coverage in the SDK is fine

Issue #429, opened 2019-10-25
 I want to know how much RDN / SVT I own


christianbrb:
## User Story
I want to know how much RDN / SVT I own so that I know if I can pay the PFS service

## Acceptance criteria
- I see my RDN / SVT balance in the UDC
- The SDK exposes it through observables

## Tasks
![grafik](https://user-images.githubusercontent.com/35398162/67951294-83e1a400-fbeb-11e9-81df-49f8a9334567.png)

christianbrb:
After the refinement:
- This is what André has showed with a nice error message and a link to a guide how to acquire RDN on the testnet and how to deposit it to the UDC

CC @taleldayekh

taleldayekh:
Current screen for reference.

![pfs](https://user-images.githubusercontent.com/43838780/68124310-61af9500-ff0f-11e9-9e1a-aa97a5e23037.png)

taleldayekh:
Current screen for reference.

![route](https://user-images.githubusercontent.com/43838780/68124401-97547e00-ff0f-11e9-97da-e9dcdb9d8030.png)

nephix:
I've removed dapp & design related acceptence criteria, because we have a separate ticket (#402) for this

kelsos:
The SDK part is merged on master, I will do the UI part integration by continuing on top of Thomas' PR.

kelsos:
The amount is now properly integrated with the SDK function and it is displayed to the user. I am closing this issue since all the related PRs have been merged.

PR #521, opened 2019-11-08
Adds udc balance fetching and IOU signing 

kelsos:
SDK part of #429 
SDK part of #463

### DIFF
--- a/raiden-ts/src/path/actions.ts
+++ b/raiden-ts/src/path/actions.ts
@@ -28,9 +28,9 @@ export const pfsListUpdated = createStandardAction('pfsListUpdated')<{
 
 export const persistIOU = createStandardAction('persistIOU')<
   {
-    signedIOU: Signed<IOU>;
+    iou: Signed<IOU>;
   },
   ServiceId
 >();
 
-export const clearIOU = createStandardAction('clearIOU')<{}, ServiceId>();
+export const clearIOU = createStandardAction('clearIOU')<undefined, ServiceId>();
--- a/raiden-ts/src/path/actions.ts
+++ b/raiden-ts/src/path/actions.ts
@@ -2,7 +2,6 @@ import { createStandardAction } from 'typesafe-actions';
 
 import { Address, UInt } from '../utils/types';
 import { Paths, PFS } from './types';
-import { BigNumberish } from 'ethers/utils';
 
 type PathId = {
   tokenNetwork: Address;
@@ -25,7 +24,7 @@ export const pfsListUpdated = createStandardAction('pfsListUpdated')<{
 export const udcBalanceFetch = createStandardAction('udcBalanceFetch')<{}>();
 
 export const udcBalanceUpdate = createStandardAction('udcBalanceUpdate')<{
-  balance: BigNumberish;
+  balance: UInt<32>;
 }>();
 
 export const udcBalanceFetchFailed = createStandardAction('udcBalanceFetchFailed').map(
--- a/raiden-ts/src/path/actions.ts
+++ b/raiden-ts/src/path/actions.ts
@@ -32,3 +32,5 @@ export const persistIOU = createStandardAction('persistIOU')<
   },
   ServiceId
 >();
+
+export const clearIOU = createStandardAction('clearIOU')<{}, ServiceId>();
--- a/raiden-ts/src/path/actions.ts
+++ b/raiden-ts/src/path/actions.ts
@@ -26,11 +26,11 @@ export const pfsListUpdated = createStandardAction('pfsListUpdated')<{
   pfsList: readonly Address[];
 }>();
 
-export const persistIOU = createStandardAction('persistIOU')<
+export const iouPersist = createStandardAction('iouPersist')<
   {
     iou: Signed<IOU>;
   },
   ServiceId
 >();
 
-export const clearIOU = createStandardAction('clearIOU')<undefined, ServiceId>();
+export const iouClear = createStandardAction('iouClear')<undefined, ServiceId>();
--- a/raiden-ts/src/path/actions.ts
+++ b/raiden-ts/src/path/actions.ts
@@ -1,12 +1,17 @@
 import { createStandardAction } from 'typesafe-actions';
 
-import { Address, UInt } from '../utils/types';
-import { Paths, PFS } from './types';
+import { Address, UInt, Signed } from '../utils/types';
+import { Paths, PFS, IOU } from './types';
 
 type PathId = {
   tokenNetwork: Address;
   target: Address;
   value: UInt<32>;
+};
+
+type ServiceId = {
+  tokenNetwork: Address;
+  serviceAddress: Address;
 };
 
 export const pathFind = createStandardAction('pathFind')<{ paths?: Paths; pfs?: PFS }, PathId>();
@@ -21,12 +26,9 @@ export const pfsListUpdated = createStandardAction('pfsListUpdated')<{
   pfsList: readonly Address[];
 }>();
 
-export const udcBalanceFetch = createStandardAction('udcBalanceFetch')<{}>();
-
-export const udcBalanceUpdate = createStandardAction('udcBalanceUpdate')<{
-  balance: UInt<32>;
-}>();
-
-export const udcBalanceFetchFailed = createStandardAction('udcBalanceFetchFailed').map(
-  (payload: Error) => ({ payload, error: true }),
-);
+export const persistIOU = createStandardAction('persistIOU')<
+  {
+    signedIOU: Signed<IOU>;
+  },
+  ServiceId
+>();
--- a/raiden-ts/src/path/actions.ts
+++ b/raiden-ts/src/path/actions.ts
@@ -2,6 +2,7 @@ import { createStandardAction } from 'typesafe-actions';
 
 import { Address, UInt } from '../utils/types';
 import { Paths, PFS } from './types';
+import { BigNumberish } from 'ethers/utils';
 
 type PathId = {
   tokenNetwork: Address;
@@ -20,3 +21,13 @@ export const pathFindFailed = createStandardAction('pathFindFailed').map(
 export const pfsListUpdated = createStandardAction('pfsListUpdated')<{
   pfsList: readonly Address[];
 }>();
+
+export const udcBalanceFetch = createStandardAction('udcBalanceFetch')<{}>();
+
+export const udcBalanceUpdate = createStandardAction('udcBalanceUpdate')<{
+  balance: BigNumberish;
+}>();
+
+export const udcBalanceFetchFailed = createStandardAction('udcBalanceFetchFailed').map(
+  (payload: Error) => ({ payload, error: true }),
+);
--- a/raiden-ts/src/path/epics.ts
+++ b/raiden-ts/src/path/epics.ts
@@ -22,7 +22,14 @@ import {
 } from 'rxjs/operators';
 import { fromFetch } from 'rxjs/fetch';
 import { ActionType, isActionOf } from 'typesafe-actions';
-import { BigNumber, bigNumberify, hexDataLength, hexlify, toUtf8Bytes } from 'ethers/utils';
+import {
+  BigNumber,
+  bigNumberify,
+  hexDataLength,
+  hexlify,
+  toUtf8Bytes,
+  verifyMessage,
+} from 'ethers/utils';
 import { Two, Zero } from 'ethers/constants';
 import { Event } from 'ethers/contract';
 
@@ -112,18 +119,21 @@ const updateIOU = (iou: IOU, price: UInt<32>): IOU => ({
   amount: iou.amount.add(price) as UInt<32>,
 });
 
+const packIOU = (iou: IOU) =>
+  concat([
+    encode(iou.one_to_n_address, 20),
+    encode(iou.chain_id, 32),
+    encode(MessageTypeId.IOU, 32),
+    encode(iou.sender, 20),
+    encode(iou.receiver, 20),
+    encode(iou.amount, 32),
+    encode(iou.expiration_block, 32),
+  ]);
+
 const signIOU$ = (iou: IOU, signer: Signer): Observable<Signed<IOU>> =>
-  from(signer.signMessage(
-    concat([
-      encode(iou.one_to_n_address, 20),
-      encode(iou.chain_id, 32),
-      encode(MessageTypeId.IOU, 32),
-      encode(iou.sender, 20),
-      encode(iou.receiver, 20),
-      encode(iou.amount, 32),
-      encode(iou.expiration_block, 32),
-    ]),
-  ) as Promise<Signature>).pipe(map(signature => ({ ...iou, signature })));
+  from(signer.signMessage(packIOU(iou)) as Promise<Signature>).pipe(
+    map(signature => ({ ...iou, signature })),
+  );
 
 const makeAndSignLastIOURequest$ = (sender: Address, receiver: Address, signer: Signer) =>
   defer(() => {
@@ -184,7 +194,14 @@ const prepareNextIOU$ = (
           }
           if (!response.ok)
             throw new Error(`PFS: last IOU request: code=${response.status} => body="${text}"`);
-          return decode(LastIOUResults, losslessParse(text)).last_iou;
+
+          const { last_iou: lastIou } = decode(LastIOUResults, losslessParse(text));
+          const signer = verifyMessage(packIOU(lastIou), lastIou.signature);
+          if (signer !== state.address)
+            throw new Error(
+              `PFS: last iou signature mismatch: signer=${signer} instead of us ${state.address}`,
+            );
+          return lastIou;
         }),
       )
   ).pipe(
--- a/raiden-ts/src/path/epics.ts
+++ b/raiden-ts/src/path/epics.ts
@@ -64,13 +64,10 @@ const makeIOU = (sender: Address, receiver: Address, chainId: UInt<32>, oneToNAd
     signature: HashZero as Signature,
   } as IOU);
 
-const updateIOU = (iou: IOU, price: UInt<32>, blockNumber: number) =>
+const updateIOU = (iou: IOU, price: UInt<32>) =>
   ({
     ...iou,
     amount: iou.amount.add(price),
-    expiration_block: iou.expiration_block.lt(bigNumberify(blockNumber).add(7 * 24 * 60 * 4))
-      ? bigNumberify(blockNumber).add(7 * 24 * 60 * 4)
-      : iou.expiration_block.add(50),
   } as IOU);
 
 const signIOU$ = (iou: IOU, signer: Signer) =>
@@ -147,7 +144,7 @@ const prepareNextIOU$ = (
         }),
       )
   ).pipe(
-    map(iou => updateIOU(iou, pfs.price, state.blockNumber)),
+    map(iou => updateIOU(iou, pfs.price)),
     mergeMap(iou => signIOU$(iou, deps.signer)),
   );
 };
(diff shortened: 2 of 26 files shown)


raiden-ts/src/path/actions.ts
```diff
@@ -20,3 +21,13 @@ export const pathFindFailed = createStandardAction('pathFindFailed').map(
 export const pfsListUpdated = createStandardAction('pfsListUpdated')<{
   pfsList: readonly Address[];
 }>();
+
+export const udcBalanceFetch = createStandardAction('udcBalanceFetch')<{}>();
```
andrevmatos: These actions usually have a `meta` param to help identify the related request/response/error ones. Shouldn't you identify it somehow? Or else any following response/error could be associated to this request.

raiden-ts/src/path/actions.ts
```diff
@@ -20,3 +21,13 @@ export const pathFindFailed = createStandardAction('pathFindFailed').map(
...
+
+export const udcBalanceUpdate = createStandardAction('udcBalanceUpdate')<{
+  balance: BigNumberish;
```
andrevmatos: Please, avoid using `BigNumberish` in core. It's a very broad type which should be used only on public interface and immediately validated to some of the specific ones. Prefer some of the `*Int` ones, as may fit on this case.

raiden-ts/src/path/actions.ts
kelsos: I saw the meta, but in the end here you just fetch the balance of the `UserDeposit` contract for your address, so I couldn't think of any useful information that would go to meta.

andrevmatos: Hmm, if you're only fetching the balance and there's no useful meta, there's a big chance you don't need a set of actions for this, just calling the async method directly where it's needed may be good enough.

kelsos: Do you have an example in the code I could reference for this? I am trying to think how this would happen. The async method would not be an epic then right? it would be something different.

kelsos: Something like the `getTokenInfo` I would guess?

andrevmatos: Sure, look at this:
https://github.com/raiden-network/light-client/blob/27c60e432ac9ab978ebf3b3a75aa72917f0a35f9/raiden-ts/src/path/utils.ts#L80-L82
Notice this happens inside a normal function, but defines an observable, and therefore could as well be used inside an epic or mergeMap chain as a normal observable, which calls an async function and returns a result to be used on the next step of the chain.

raiden-ts/src/path/epics.ts
```diff
@@ -318,3 +437,34 @@ export const pfsServiceRegistryMonitorEpic = (
...
+  action$: Observable<RaidenAction>,
+  state$: Observable<RaidenState>,
+  { userDepositContract }: RaidenEpicDeps,
+): Observable<ActionType<typeof udcBalanceUpdate | typeof udcBalanceFetchFailed>> => {
+  return action$.pipe(
+    filter(isActionOf(udcBalanceFetch)),
+    debounceTime(10e3),
+    withLatestFrom(state$),
+    mergeMap(([{}, state]) => {
+      return from(userDepositContract.balances(state.address)).pipe(
```
andrevmatos: Also, don't forget `functions`. I'm not sure ethers also expose as direct method, but even if it does, for consistency, it's better to call method from `contract.functions`

raiden-ts/src/types.ts
```diff
@@ -23,6 +24,8 @@ interface Info {
 export interface ContractsInfo {
   TokenNetworkRegistry: Info;
   ServiceRegistry: Info;
+  UserDeposit: Info;
+  OneToN: Info;
```
andrevmatos: I'd say there's no need to require `OneToN` info on this interface, which needs to be fullfilled by anyone instantiating a Raiden instance. Instead, it can be pulled in from `UserDepositContract.functions.one_to_n_address`, so one less source of inconsistency to need to check from there. Example:
https://github.com/raiden-network/light-client/blob/27c60e432ac9ab978ebf3b3a75aa72917f0a35f9/raiden-ts/src/path/utils.ts#L48-L51

raiden-ts/src/state.ts
```diff
@@ -130,5 +138,9 @@ export const initialState = makeInitialState({
     TokenNetworkRegistry: { address: AddressZero as Address, block_number: 0 },
     // eslint-disable-next-line @typescript-eslint/camelcase
     ServiceRegistry: { address: AddressZero as Address, block_number: 0 },
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    UserDeposit: { address: AddressZero as Address, block_number: 0 },
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    OneToN: { address: AddressZero as Address, block_number: 0 },
```
andrevmatos: idem

raiden-ts/src/types.ts
kelsos: Thanks, I missed that completely. I will adjust accordingly

raiden-ts/src/path/types.ts
```diff
@@ -68,3 +68,43 @@ export interface RaidenPFS {
...
+    sender: Address,
+    receiver: Address,
+    amount: UInt(32),
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    expiration_block: UInt(32),
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    one_to_n_address: Address,
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    chain_id: UInt(32),
+    signature: Signature,
```
andrevmatos: Careful. This is a required property, but in some places you initialize IOUs with type assertions without it. This is a problem, as you tell TypeScript to trust you about something, but it's not true.
Instead, I'd say you should use `Signed` type. For it, move `Signed` type from `messages/types.ts` to `utils/types.ts`, remove the `extends Message` constraint on the type, so it's valid for any type, and then use the distinction between `IOU` (not signed) and `Signed<IOU>` & `Signed(IOU)`.

raiden-ts/src/path/epics.ts
```diff
@@ -1,48 +1,154 @@
...
+const makeIOU = (sender: Address, receiver: Address, chainId: UInt<32>, oneToNAddress: Address) =>
+  ({
+    sender: sender,
+    receiver: receiver,
+    chain_id: chainId,
+    amount: Zero as UInt<32>,
+    one_to_n_address: oneToNAddress,
+    expiration_block: Zero as UInt<32>,
+    signature: HashZero as Signature,
+  } as IOU);
```
andrevmatos: As I commented on the `Signed` approach, this is not good. `HashZero` isn't a valid signature. Also,  the type assertion here can trick you sometimes. Instead, prefer putting `IOU` or `Signed<IOU>` in the return type of the function, and then just return the object. TypeScript will then make sure your object matches the return type or complain earlier otherwise.

kelsos: Thanks, I will change accordingly :)

andrevmatos: Also, as much as possible when creating objects, to create them ready. You can use this factory function by passing all needed parameters or create it in place (should be created in a single place, I suppose, right?), but try to make it ready. Zero isn't valid amount or expiration_block, and the thing with signature I already stated.

raiden-ts/src/path/epics.ts
```diff
@@ -1,48 +1,154 @@
...
+    concat([
+      encode(iou.one_to_n_address, 20),
+      encode(iou.chain_id, 32),
+      encode(MessageTypeId.IOU, 32),
+      encode(iou.sender, 20),
+      encode(iou.receiver, 20),
+      encode(iou.amount, 32),
+      encode(iou.expiration_block, 32),
+    ]),
+  ) as Promise<Signature>).pipe(map(signature => ({ ...iou, signature } as IOU)));
```
andrevmatos: Same: return type vs type assertion, prefer the former, or in this case, `Observable<Signed<IOU>>` or equivalent

raiden-ts/src/path/epics.ts
```diff
@@ -111,8 +217,13 @@ export const pathFindServiceEpic = (
                       pluck(0),
                     );
                 return pfs$.pipe(
-                  // TODO: use pfs.price to create & sign IOU
                   mergeMap(pfs =>
+                    pfs.price.isZero()
+                      ? of(undefined)
+                      : prepareNextIOU$(state, pfs, deps, httpTimeout, tokenNetwork),
+                  ),
+                  withLatestFrom(pfs$),
```
andrevmatos: Don't reuse it. Notice `pfs$` can be a cold observable which performs some [slow] network request. Instead, either return an explicit `of(pfs)` or put the body of `mergeMap` below also under/inside this mergeMap above, of `prepareNextIOU$`. There're tons of example of this pattern around, tell me if you have any question.

raiden-ts/src/path/types.ts
```diff
@@ -68,3 +68,43 @@ export interface RaidenPFS {
...
+  }),
+);
+
+// eslint-disable-next-line @typescript-eslint/interface-name-prefix
+export interface IOU extends t.TypeOf<typeof IOU> {}
+
+export interface RaidenIOU {
```
andrevmatos: I don't see you exposing the `IOU` anywhere. Why the public interface? Also, these public types are only needed if you plan to receive it, not only outputting it, and would need to be exposed in `src/index.ts` for easy access, but in this case I think should remove it.

raiden-ts/src/path/epics.ts
```diff
@@ -1,48 +1,154 @@
...
+  });
+
+const prepareNextIOU$ = (
+  state: RaidenState,
+  pfs: PFS,
+  deps: RaidenEpicDeps,
+  httpTimeout: number,
+  tokenNetwork: Address,
+) => {
+  const cachedIOU: IOU = get(state.iou, [tokenNetwork, pfs.address]);
```
andrevmatos:
```suggestion
  const cachedIOU: IOU | undefined = get(state.iou, [tokenNetwork, pfs.address]);
```

raiden-ts/src/path/epics.ts
```diff
@@ -1,48 +1,154 @@
...
+              method: 'GET',
+              headers: { 'Content-Type': 'application/json' },
+            },
+          ),
+        ),
+        timeout(httpTimeout),
+        mergeMap(async response => {
+          const text = await response.text();
+          if (response.status === 404) {
+            return makeIOU(
```
andrevmatos: Won't this break because of zero expiration block on makeIOU?

raiden-ts/src/path/epics.ts
```diff
@@ -1,48 +1,154 @@
...
+              state.address,
+              pfs.address,
+              bigNumberify(state.chainId) as UInt<32>,
+              deps.contractsInfo.OneToN.address,
```
andrevmatos: Here's where a memoized async getter over `UserDeposit` contract is better than an explicit `contractsInfo` entry

raiden-ts/src/path/epics.ts
```diff
@@ -1,48 +1,154 @@
...
+            {
+              method: 'GET',
+              headers: { 'Content-Type': 'application/json' },
+            },
+          ),
+        ),
+        timeout(httpTimeout),
+        mergeMap(async response => {
+          const text = await response.text();
+          if (response.status === 404) {
```
andrevmatos: You may need an explicit codec for the error case response. If possible/needed, make it private/contextual as it is used only here, ever.

raiden-ts/src/path/epics.ts
kelsos: You are right I should adjust the constant and update the makeIOU, thanks.

raiden-ts/src/raiden.ts
```diff
@@ -259,6 +263,26 @@ export class Raiden {
...
     });
 
+    this.getUDCBalance = async () =>
+      from(this.deps.userDepositContract.functions.balances(address))
+        .pipe(
+          map(balance => {
+            const owedAmount = Object.values(state.path.iou)
+              .reduce(
+                (acc, value) =>
+                  acc.concat(
```
nephix: **_Might_** make sense performance-wise here to use
```
Array.prototype.push.apply(arr1, arr2)
```
instead of `concat()` because `push()` is a lot faster. `arr1.push(...arr2)` seems like a better, shorter ES6-notation for this. I don't know to what dimensions `state.path.iou` can grow but [here's](https://dev.to/uilicious/javascript-array-push-is-945x-faster-than-array-concat-1oki) an article with benchmarks and explanations. Since we want to use the SDK on smaller IoT devices as well it might make sense to consider this

raiden-ts/src/raiden.ts
```diff
@@ -259,6 +263,26 @@ export class Raiden {
...
+                    Object.values(value).filter(value =>
+                      value.expiration_block.gte(state.blockNumber),
+                    ),
+                  ),
+                new Array<IOU>(),
+              )
+              .reduce((acc, iou) => acc.add(iou.amount), Zero);
```
nephix: Might be better to map-reduce instead of reducing twice

kelsos: Could you give me a basic example of how it would work? I tried a few approaches and I don't think I get it. 

One reduce is used to flatten an array of arrays to a single dimension array (instead of flatMap), and the other reduce is used to calculate the sum.

nephix: Hmm I think map isn't even necessary, my mistake. It could simply be
```
Object.values(state.path.iou)
   .flat()
   .filter(iou => iou.expiration_block.gte(state.blockNumber))
   .reduce((acc,iou) => ...)
```
(untested)

kelsos: Do you know if `flat` is poly filled when it is not supported I can see that it is implemented around 83% of the browsers.

I was trying flatMap didn't even think of flat (and flatMap didnt work)

nephix: Hmm I'm not sure if it's polyfilled, but looks like it's supported by all browsers that MetaMasks supports. I guess that's enough support for us? Should work just fine on nodejs too. Maybe we can bring this up internally to get clarification

kelsos: I will give it a try. It shouldn't take much time :)

raiden-ts/src/path/actions.ts
```diff
@@ -20,3 +25,12 @@ export const pathFindFailed = createStandardAction('pathFindFailed').map(
 export const pfsListUpdated = createStandardAction('pfsListUpdated')<{
   pfsList: readonly Address[];
 }>();
+
+export const persistIOU = createStandardAction('persistIOU')<
+  {
+    signedIOU: Signed<IOU>;
```
andrevmatos: Nitpick: this can be called only `iou`, as the type already implies it's signed

raiden-ts/src/path/actions.ts
```diff
@@ -20,3 +25,12 @@ export const pathFindFailed = createStandardAction('pathFindFailed').map(
...
+  },
+  ServiceId
+>();
+
+export const clearIOU = createStandardAction('clearIOU')<{}, ServiceId>();
```
andrevmatos: Prefer `undefined` as `payload` type, instead of empty object, as typesafe-actions has proper specialized typings for it

raiden-ts/src/path/epics.ts
```diff
@@ -1,48 +1,203 @@
...
+        value: UInt(32),
+      }),
+    }),
+  ]),
+);
+interface PathError extends t.TypeOf<typeof PathError> {}
+
+type IOUPayload = {
+  serviceAddress: Address;
+  signedIOU?: Signed<IOU>;
```
andrevmatos: Idem

raiden-ts/src/path/epics.ts
```diff
@@ -1,48 +1,203 @@
...
+        to: Address,
+        value: UInt(32),
+      }),
+    }),
+  ]),
+);
+interface PathError extends t.TypeOf<typeof PathError> {}
+
+type IOUPayload = {
+  serviceAddress: Address;
```
andrevmatos: Do we really need a new type for it? Isn't `serviceAddress` the same as `iou.receiver`? If so, you can replace the whole type by just the `Signed<IOU>` object

raiden-ts/src/path/epics.ts
```diff
@@ -1,48 +1,203 @@
...
+  signedIOU?: Signed<IOU>;
+};
+
+class InvalidPFSRequestError extends Error {
```
andrevmatos: Nitpick: this would be the 2nd class in whole SDK. I've been avoiding this pattern, specially for errors, as errors should always be an exceptional case, not an expected one. Can this be refactored without the error class? Maybe an specific object or flag?

andrevmatos: This is at your discretion though, if you think it's worth it.

raiden-ts/src/path/epics.ts
kelsos: You are right. I forgot about that. Thanks :)

raiden-ts/src/path/epics.ts
kelsos: Like we discussed, the main issue is that I have to perform a secondary action in some cases along with the failure. Addind a class extending error is the only way I can think of passing the iou to the error handling. But I am open to suggestions :+1:

raiden-ts/src/path/epics.ts
```diff
@@ -1,48 +1,198 @@
...
+            {
+              method: 'GET',
+              headers: { 'Content-Type': 'application/json' },
+            },
+          ),
+        ),
+        timeout(httpTimeout),
+        mergeMap(async response => {
+          const text = await response.text();
+          const oneToNAddress = memoize(
```
andrevmatos: This needs to be defined outside function, or else `memoize` is getting re-created every time and will just memory-leak instead of helping cache the async call

kelsos: So moving it to top-level should be enough right

raiden-ts/src/path/types.ts
```diff
@@ -68,3 +68,32 @@ export interface RaidenPFS {
...
+  }),
+);
+
+// eslint-disable-next-line @typescript-eslint/interface-name-prefix
+export interface IOU extends t.TypeOf<typeof IOU> {}
+
+export const LastIOUResults = t.readonly(
+  t.type({
+    // eslint-disable-next-line @typescript-eslint/camelcase
+    last_iou: IOU,
```
andrevmatos: Isn't this `Signed`? How to verify the last IOU returned by PFS is valid?

raiden-ts/src/path/epics.ts
```diff
@@ -1,48 +1,198 @@
...
+              pfs.address,
+              pfs.price,
+              bigNumberify(state.chainId) as UInt<32>,
+              await oneToNAddress(deps.userDepositContract),
+              state.blockNumber,
+            );
+          }
+          if (!response.ok)
+            throw new Error(`PFS: last IOU request: code=${response.status} => body="${text}"`);
+          return decode(LastIOUResults, losslessParse(text)).last_iou;
```
andrevmatos: You need to also verify signature is valid and ours

raiden-ts/src/path/types.ts
kelsos: Yes it is signed an yes we need to verify the signature

raiden-ts/src/path/actions.ts
```diff
@@ -20,3 +25,12 @@ export const pathFindFailed = createStandardAction('pathFindFailed').map(
 export const pfsListUpdated = createStandardAction('pfsListUpdated')<{
   pfsList: readonly Address[];
 }>();
+
+export const persistIOU = createStandardAction('persistIOU')<
```
andrevmatos: Nitpick: can these actions be named as `${domain}{verb}` for consistency? e.g. `iouPersist` & `iouClear`

kelsos:
Thank you for the input :)

Issue #402, opened 2019-10-18
Extend the payment stepper to select PFS and path


christianbrb:
## Description
After #386 and #392 we have got two modal dialogs before the payment stepper. 

Extend the payment stepper to include the PFS and route selection instead of the two modal dialogs.

## Acceptance criteria
- Show the most important piece of information in every step
- I see a list of the PFS
- I have a stepper on top

## Tasks
- [ ] 

CC @roy-castrillo @taleldayekh

taleldayekh:
Brainstorming w/ Roy

![IMG_4459](https://user-images.githubusercontent.com/43838780/68203044-41421200-ffc5-11e9-9b22-9e71bfdd7e65.jpeg)

nephix:
As the designs for this are done, I'm going ahead and implement the dapp part of it

christianbrb:
Looks like it is done and can be closed

PR #523, opened 2019-11-12
Feature/payment steps select pfs path

nephix:
Adds new payment stepper #402

### DIFF
--- a/raiden-dapp/src/views/TransferSteps.vue
+++ b/raiden-dapp/src/views/TransferSteps.vue
@@ -72,36 +72,51 @@
           </v-stepper-content>
 
           <v-stepper-content step="3">
-            <v-row justify="center" align="center">
-              <v-col cols="12">
-                <p>
-                  {{ $t('transfer.steps.confirm-transfer.total-amount') }}
-                </p>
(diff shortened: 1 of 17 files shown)


nephix:
@kelsos sorry for the large pull request, but most of the changes are just refactoring and tests.

raiden-dapp/src/views/TransferSteps.vue
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    if (this.step === 3 && this.selectedRoute && this.selectedPfs) {
+      this.transfer();
+    }
+  }
+
+  get token(): Token {
+    const { token: address } = this.$route.params;
+    return this.$store.state.tokens[address] || ({ address } as Token);
```
kelsos: I noticed in my PR that when you have parameters in the URL and this causes an error during loading due to the fact that the default token we use doesn't have any decimals. 

It's not part of this PR but can you try refreshing the payment screen to verify that your don't get any console.errors?

nephix: Thanks 🙏 
Sure will check it out tomorrow and create a follow up ticket if needed

Issue #388, opened 2019-10-14
I want to deposit tokens to the UDC


christianbrb:
## User Story
As Chris, I want to deposit tokens to the UDC so that I am able to pay for getting routes.

## Acceptance criteria
- I see the amount of SVT / RDN that I am going to mint & deposit
- I am going to deposit the amount of tokens I have minted
- I see a button to mint & deposit to the UDC
- Minting and depositing is done in a single step
- The SDK exposes a function deposit to the UDC
- The UDC deposit is stored in an observable

![grafik](https://user-images.githubusercontent.com/35398162/67953273-16377700-fbef-11e9-85a4-0f83b4ddbb62.png)

christianbrb:
Refinement:
- We need a design for it
- Put it where the user needs to use a paid PFS
- Check, if the UDC deposit is made per PFS services or globally ? @palango

taleldayekh:
The UDC deposit is made once and the amout deposited can be used with however many PFS (and monitoring) services you wish until the UDC is "drained". @christianbrb

kelsos:
Taking care of the SDK part: https://github.com/raiden-network/light-client/issues/389#issuecomment-555083544

kelsos:
All PRs related to this issue have been merged to master and this issue is considered complete so I am closing.

PR #548, opened 2019-11-18
Exposes UDC token address and deposit function

kelsos:
Implements the SDK part of #388

### DIFF
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -1,7 +1,7 @@
 import { Signer } from 'ethers';
 import { Wallet } from 'ethers/wallet';
 import { AsyncSendable, Web3Provider, JsonRpcProvider } from 'ethers/providers';
-import { Network, BigNumber, BigNumberish } from 'ethers/utils';
+import { Network, BigNumber, BigNumberish, bigNumberify } from 'ethers/utils';
 import { Zero } from 'ethers/constants';
 
 import { MatrixClient } from 'matrix-js-sdk';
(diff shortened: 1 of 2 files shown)


raiden-ts/src/raiden.ts
```diff
@@ -1011,6 +1021,53 @@ export class Raiden {
...
+   */
+  public async depositToUDC(amount: BigNumberish): Promise<Hash> {
+    const depositAmount = bigNumberify(amount);
+
+    if (depositAmount.isZero()) throw new Error('Please deposit a positive amount.');
+
+    const { userDepositContract, address } = this.deps;
+
+    const tokenAddress = await this.userDepositTokenAddress();
+    const serviceToken = CustomTokenFactory.connect(tokenAddress, this.deps.signer);
```
andrevmatos: `CustomToken` is needed only for minting. As we don't mint on this function, please, use `HumanStandardToken` for consistency

raiden-ts/src/raiden.ts
```diff
@@ -1011,6 +1021,53 @@ export class Raiden {
...
+    const balance = await serviceToken.functions.balanceOf(address);
+
+    if (balance.lt(amount)) throw new Error(`Insufficient token balance (${balance}).`);
+
+    const approveTx = await serviceToken.functions.approve(
+      userDepositContract.address,
+      depositAmount,
+    );
+    const approveReceipt = await approveTx.wait();
```
andrevmatos: If you're confident checks above (i.e. balance) are enough to ensure the tx will go through, a nice optimization is to move `wait`ing for tx to be mined after deposit tx is subtmitted (you can then wait on `Promise.all()` txs). This way, both txs are queued, and can even be mined on the same block, as ethereum node is smart enough to queue correctly by nonce. This is an optimization I plan to implement for `channelDeposit` as well, but this could be a nice demo of this approach. Worse case you just spend a little more of gas, but the speed gains in the trivial case may be worth it if the unhappy case is unlikely/impossible.

kelsos: You can never be sure. But I can give it a try and see how it works.

kelsos:
```typescript
  const [approveReceipt, depositReceipt] = await Promise.all([approveTx.wait(), depositTx.wait()])
``` 
This is what you had in mind right?

PR #550, opened 2019-11-19
Update Light Client to raiden develop

andrevmatos:
- [x] raiden-contracts v0.33.3
- [x] ~~update messages fields, packing & signing~~[split PR]
- [x] ~~IOU fixes~~[split PR]
- [ ] ~~update transport~~[split PR]


raiden-ts/src/path/epics.ts
```diff
@@ -61,37 +61,28 @@ const oneToNAddress = memoize(
 
 /**
  * Codec for PFS API returned error
+ *
+ * May contain other fields like error_details, but we don't care about them (for now)
  */
 const PathError = t.readonly(
```
kelsos: Thanks for removing I was thinking to remove them at some point and then I completely forgot it :)

raiden-ts/src/path/epics.ts
```diff
@@ -127,55 +118,61 @@ const makeAndSignLastIOURequest$ = (sender: Address, receiver: Address, signer:
...
+            mergeMap(payload =>
+              fromFetch(
+                `${pfs.url}/api/v1/${tokenNetwork}/payment/iou?${new URLSearchParams(
+                  payload,
+                ).toString()}`,
+                {
+                  method: 'GET',
+                  headers: { 'Content-Type': 'application/json' },
+                },
+              ).pipe(timeout(httpTimeout)),
```
kelsos: Thank you for taking the time to fix the timeout issue.

raiden-ts/src/path/epics.ts
andrevmatos: This wouldn't usually be a problem, but I've stumbled upon `error_details` which didn't match the provided schema. Seems it's error-dependent, so either we kept it as `t.any` or just remove it, as we don't inspect this property.

raiden-ts/src/path/epics.ts
andrevmatos: I've hit it as well, so made sense.

raiden-ts/src/path/epics.ts
kelsos: Removing is better than having dead code. I was thinking that before merging and then I missed it so thanks for taking care of it :)

raiden-ts/src/path/epics.ts
```diff
@@ -254,9 +251,12 @@ export const pathFindServiceEpic = (
                   mergeMap(pfs =>
                     pfs.price.isZero()
                       ? of({ pfs, iou: undefined })
-                      : prepareNextIOU$(state, pfs, deps, httpTimeout, tokenNetwork).pipe(
-                          map(iou => ({ pfs, iou })),
-                        ),
+                      : prepareNextIOU$(
+                          pfs,
+                          tokenNetwork,
+                          cached$.pipe(pluck(0)) /* cached state$ */,
```
kelsos: I am mostly curious about the reasoning behind this change also adding everything on the state. was the problem that passing the state the other way would result in the state being old/out of date?

andrevmatos: It wasn't on my tests because it was quick, but could be. Notice we use it mostly to get the blockNumber, and it's fetched after signing and requesting lastIOU, which can take a long time, so in the end, the blockNumber could be outdated. This change ensures reactive state is always kept up to date.

andrevmatos:
As requested, I'm going to split this PR in smaller ones.

PR #583, opened 2019-11-24
New start screen

taleldayekh:


raiden-dapp/src/components/AppCore.vue
```diff
@@ -1,13 +1,25 @@
 <template>
-  <tokens />
+  <NoTokens v-if="!tokens.length" />
```
kelsos:
```suggestion
  <no-tokens v-if="!tokens.length" />
```

In all the components we use the components in kebab case.
We should also add a lint rule to enforce this.

raiden-dapp/src/components/NoTokens.vue
```diff
@@ -0,0 +1,34 @@
...
+import { Component, Mixins } from 'vue-property-decorator';
+import NavigationMixin from '@/mixins/navigation-mixin';
+
+@Component({})
+export default class NoTokens extends Mixins(NavigationMixin) {}
+</script>
+
+<style lang="scss" scoped>
+.new-token {
+  text-align: center;
```
kelsos: Can you move the button and header parts here?
as `&__button ` and `&__header`?

raiden-dapp/src/components/AppCore.vue
taleldayekh: Done

raiden-dapp/src/components/NoTokens.vue
taleldayekh: Done

PR #592, opened 2019-11-26
Fix config.logger

andrevmatos:
It couldn't be fully disabled because even setting it to `''` still outputed the `console.group`, even if without content.

Now, setting `config.logger` to empty string or (if left as default of `undefined) `NODE_ENV` environment variable to something other than `develop`, `redux-logger` will be completely disabled.
One can also set it to some of the `console`'s logging level to enable and force it to a specific logging level.

Additionally, if the logs are disabled  on either way, also reflect this on matrix logging, which could polute the output (actually, set to filter only critical `error`s, but most of the normal logs are debug).

TODO for the future is replacing our internal direct usage of some `console`'s logging functions with a proper library like [loglevel](https://github.com/pimterry/loglevel), used by matrix.

### DIFF
--- a/raiden-ts/src/config.ts
+++ b/raiden-ts/src/config.ts
@@ -37,7 +37,14 @@ export const RaidenConfig = t.readonly(
     }),
     t.partial({
       matrixServer: t.string,
-      logger: t.keyof({ ['']: null, debug: null, log: null, info: null, warn: null, error: null }),
+      logger: t.keyof({
+        ['']: null,
+        trace: null,
+        debug: null,
+        info: null,
+        warn: null,
+        error: null,
+      }),
       pfs: t.union([Address, t.string, t.null]),
     }),
   ]),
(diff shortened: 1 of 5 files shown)


kelsos:
@andrevmatos I really like the idea of having a proper logging solution for this SDK. My experience so far is that these logging libraries make things easier to customize and turn off when needed.

nephix:
That would be great for the tests which currently produce a lot of output

kelsos:
@nephix we had a discussion with @andrevmatos on this before and the optimal would be if we could somehow manage to show detailed logging only for failing tests. In a failing test the detailed logging might actually help but in the successful ones it is just noise.

nephix:
Would be great, just not sure if possible because you would have to know before running the actual test if it's going to fail or not, to enable logging?

kelsos:
Yes I did a quick check at some point to see if that is possible with jest and it seemed to be an open issue of people asking for the same thing but without something being done about it.

PR #594, opened 2019-11-26
Adds no logging serve script

kelsos:


nephix:
Is `NODE_ENV=nologging` also used somewhere? Internally by vue-cli? Couldn't find anything related after a quick search

kelsos:
@nephix no not really. It just had to be `!="development"` and obviously not `production` the target is to have no logs so I did some _"creative thinking"_...

PR #599, opened 2019-11-26
Enforces kebab-case in component names

kelsos:


kelsos:
@nephix rules are good, especially the ones that are automatically enforced an fixable

LefterisJP:
I actually learned something. Had no idea `that-this-is-kebab-case`

kelsos:
@LefterisJP I knew it was called something but I didn't remember. I actually had to check vue's documentation for that.

Issue #408, opened 2019-10-22
I don't want to be bothered by routes if I am just doing direct payments


christianbrb:
## User Story
As Chris I don't want to be bothered by routes if I am just doing direct payments so that the dApp is easier and cheaper to use.

## Acceptance criteria
- The Fee dialogue is skipped
- The Select PFS dialogue is skipped
- The Select Route dialogue is skipped
- The user is informed, that direct payment is possible (How is the user is informed)
- The overlay doesn't rely on the component
- Do the SDK communication on the host component
- Out of scope: Availability of the node (#218)

andrevmatos:
Fixed by #600

Issue #591, opened 2019-11-25
Make actions serializable


andrevmatos:
## Description

We need serializable actions. Usually, it's not expected actions to need to be serialized, nor go in state, but it's useful for one specific usecase: _pending_ actions that have a **confirmation** counterpart, and we need to be able to recover in case client is closed between both.

The action creator implementation should use `io-ts`, our codec/validation library of choice, and be self-contained (no multiple places of declarations to add a new action). It must also be type-safe and implement the [FSA](https://github.com/redux-utilities/flux-standard-action) pattern (actions being plain objects containg only `type`, `payload`, `meta` & `error` properties).

From this, it'd be interesting if the Acion Creator itself was the codec instance, so members can be accessed directly, but it may not be practical to do it over the callable function needed for the action creator. If this is the case, the codec can be defined as a `codec` property.

It may replace or extend our [typesafe-actions](https://github.com/piotrwitek/typesafe-actions) dependency, may have an interface similar to it (to minimize changes), but ideally replace standalone utilities (ActionType, `getType`, `isActionOf`, etc) with methods/member functions in the action creator object itself.

A generic `Error` serialization codec is needed for the error actions, like [serialize-error](https://github.com/sindresorhus/serialize-error) but typesafe and compatible with `io-ts`. No need for strong guarantees on the types, as they're mostly informative in the SDK.

## Acceptance criteria
- Actions can be serialized
- Actions can be deserialized/validated
- Errors payloads can also be [de]serialized
- Actions are plain objects in the FSA pattern
- Action Creators can be created passing io-ts codecs for `payload` & `meta` members
- The action codec can be accessed, but relevant properties/methods are exposed through utility members.
- An efficient type-guard is exposed as a member arrow function (no `this` unbind) of the AC which checks only `type` literal instead of `codec.is` which validates recursively
- Codec, payload, meta and literal `type` types are exposed as well; a standalone `ActionType` generic can also be exposed to calculate this in a standalone way
- All our current action creators are converted to use this implementation/factory

## Tasks
- [ ]

christianbrb:
@andrevmatos Thanks for pointing this out :)
Some additional information: https://github.com/reduxjs/redux/blob/master/docs/faq/Actions.md#why-should-type-be-a-string-or-at-least-serializable-why-should-my-action-types-be-constants

nephix:
[...]

PR #600, opened 2019-11-26
Proper fees dialogs behavior if direct route available, single free PFS or single free route

### LINKED ISSUE #596: Can't use a PFS set in config
## Steps to Reproduce
1. Get the `raiden` SDK instance on browser's dev tool
2. Set a config with `raiden.updateConfig({ pfs: 'https://pfs-goerli-with-fee.services-test.raiden.network' })`
3. Try to perform a transfer

## Expected Result
- Skip PFS selection dialog and call `findRoutes` without specifying a `pfs` if `raiden.config.pfs !== undefined`
- Can fetch routes and transfer using this set PFS

## Actual Result
image

andrevmatos:
Fix #596 

- There's a new method available, `Raiden.directRoute`, similar to `findRoutes`, which can get earlier if a direct route is available
  - This can be used to skip PFS fee & mediation fee dialogs
  - It resolves to `undefined` if no direct route is found; rejects only if validation fails (e.g. token address), not on route not found like `findRoutes`
- `findPFS` will throw only if pfs is disabled in config (`config.pfs=null`), `directRoute` before can skip this error if direct transfer is possible; it uses set `config.pfs`, or fetches from `ServiceRegistry` only if `undefined` (default)
- Should skip PFS fee dialog if `findPFS` finds a single free (price=0) PFS
- Should skip mediation fee dialog if `findRoutes` finds a single free route

So, general dApp workflow should be like:
- call `Raiden.directRoute`
  - if there's a direct route: use it as `paths` for `transfer` in last point directly
  - else (resolves to undefined)
    - call `findPFS`
      - if throws, show error, means no direct route and PFS disabled
      - if returns a single free PFS, auto-confirm (skip) PFS fee dialog
      - else, show dialog and let the user select and confirm payment for a PFS
    - call `findRoutes` with selected `pfs` above
      - if there's a single free route, auto-confirm (skip) mediation fee dialog
      - else, show dialog and let user select and confirm a route and its fee
- call `transfer` with selected `paths`

### DIFF
--- a/raiden-dapp/src/services/raiden-service.ts
+++ b/raiden-dapp/src/services/raiden-service.ts
@@ -4,7 +4,7 @@ import { RootState } from '@/types';
 import { Web3Provider } from '@/services/web3-provider';
 import { BalanceUtils } from '@/utils/balance-utils';
 import { DeniedReason, Progress, Token, TokenModel } from '@/model/types';
-import { BigNumber } from 'ethers/utils';
+import { BigNumber, BigNumberish } from 'ethers/utils';
 import { Zero } from 'ethers/constants';
 import { exhaustMap, filter, first } from 'rxjs/operators';
 import asyncPool from 'tiny-async-pool';
@@ -293,6 +293,15 @@ export default class RaidenService {
 
   noPfsSelected(): boolean {
     return this.raiden.config.pfs === undefined;
+  }
+
+  /* istanbul ignore next */
+  async directRoute(
+    token: string,
+    target: string,
+    value: BigNumberish
+  ): Promise<RaidenPaths | undefined> {
+    return await this.raiden.directRoute(token, target, value);
   }
 
   /* istanbul ignore next */
--- a/raiden-dapp/src/services/raiden-service.ts
+++ b/raiden-dapp/src/services/raiden-service.ts
@@ -291,10 +291,6 @@ export default class RaidenService {
     return raidenPFS;
   }
 
-  noPfsSelected(): boolean {
-    return this.raiden.config.pfs === undefined;
-  }
-
   /* istanbul ignore next */
   async directRoute(
     token: string,
--- a/raiden-dapp/src/views/TransferSteps.vue
+++ b/raiden-dapp/src/views/TransferSteps.vue
@@ -320,6 +320,27 @@ export default class TransferSteps extends Mixins(
 
     if (typeof this.token.decimals !== 'number') {
       this.navigateToHome();
+      return;
+    }
+
+    const directRoutes = await this.$raiden.directRoute(
+      address,
+      this.target,
+      BalanceUtils.parse(this.amount, this.token.decimals)
+    );
+
+    if (directRoutes) {
+      const [route] = directRoutes;
+
+      this.selectedRoute = {
+        key: 0,
+        fee: Zero,
+        displayFee: '0',
+        path: [...route.path],
+        hops: 0
+      };
+
+      this.transfer();
     }
   }
 
--- a/raiden-dapp/src/views/TransferSteps.vue
+++ b/raiden-dapp/src/views/TransferSteps.vue
@@ -174,12 +174,16 @@
       v-if="step === 1"
       :value="pfsFeesConfirmed"
       class="confirmation-overlay"
+      :class="{ 'v-overlay--dark': freePfs && pfsFeesConfirmed }"
     >
       <spinner v-if="!pfsFeesPaid" />
       <checkmark v-else class="confirmation-overlay__checkmark" />
 
-      <h2 v-if="!pfsFeesPaid">
+      <h2 v-if="!pfsFeesPaid && !freePfs">
         {{ this.$t('transfer.steps.request-route.in-progress') }}
+      </h2>
+      <h2 v-else-if="freePfs">
+        {{ this.$t('transfer.steps.request-route.searching-for-route') }}
       </h2>
       <h2 v-else>{{ this.$t('transfer.steps.request-route.done') }}</h2>
     </v-overlay>
@@ -212,7 +216,7 @@
 <script lang="ts">
 import { Component, Mixins } from 'vue-property-decorator';
 import { RaidenPFS } from 'raiden-ts';
-import { BigNumber } from 'ethers/utils';
+import { BigNumber, bigNumberify } from 'ethers/utils';
 
 import { BalanceUtils } from '@/utils/balance-utils';
 import { Token, Route, emptyDescription, StepDescription } from '@/model/types';
@@ -253,6 +257,7 @@ export default class TransferSteps extends Mixins(
   routes: Route[] = [];
   pfsFeesConfirmed: boolean = false;
   pfsFeesPaid: boolean = false;
+  freePfs: boolean = false;
   showMintDeposit: boolean = false;
   mediationFeesConfirmed: boolean = false;
   processingTransfer: boolean = false;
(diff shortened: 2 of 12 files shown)


christianbrb:
This would fix #408

kelsos:
@andrevmatos I should be done with the dApp part. I did try it by hand also and it seems to work properly. I am not sure if you also want to have a look. 

I would like to re-evaluate what happens if for example you have free pfs and there is no route. Do you go back to the previous `transfer` screen or you go to the select PFS step. But these are mostly UX matters. We can try the scenarios mentioned and then maybe @taleldayekh can help us plan the user flow in these cases better.

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -338,6 +338,15 @@ export default class RaidenService {
     return this.raiden.config.pfs === undefined;
```
andrevmatos: I think you may remove the `noPfsSelected` now, you shouldn't need to inspect it, right?

raiden-dapp/src/views/TransferSteps.vue
```diff
@@ -353,6 +378,13 @@ export default class TransferSteps extends Mixins(
             path
           } as Route)
       );
+
+      const [route] = this.routes;
+
+      if (route && route.fee.isZero()) {
+        this.selectedRoute = route;
+        this.transfer();
```
nephix: Maybe go to step 3 first and let the user confirm the transfer. Otherwise from UX he would see the 3 steps at the top and then move directly from 2nd step to the transfer screen. Very minor

raiden-dapp/tests/unit/views/transfer-steps.spec.ts
```diff
@@ -240,4 +247,47 @@ describe('TransferSteps.vue', () => {
     expect(transferDone).toBeCalledTimes(0);
     expect(wrapper.vm.$data.error).toEqual('failure');
   });
+
+  test('skip to transfer if a direct route is available', async () => {
```
nephix: `'Should...'`

raiden-dapp/tests/unit/views/transfer-steps.spec.ts
```diff
@@ -240,4 +247,47 @@ describe('TransferSteps.vue', () => {
...
+    expect.assertions(2);
+    $raiden.directRoute.mockResolvedValueOnce([freeRoute] as RaidenPaths);
+    $raiden.transfer.mockResolvedValueOnce(undefined);
+    const wrapper = createWrapper({});
+    await flushPromises();
+    expect(wrapper.vm.$data.step).toBe(1);
+    expect($raiden.transfer).toHaveBeenCalledTimes(1);
+  });
+
+  test('skip to transfer if a free route is available', async () => {
```
nephix: `'Should...'`

raiden-dapp/src/services/raiden-service.ts
kelsos: I am on it :) I will add an extra commit removing it

raiden-dapp/tests/unit/views/transfer-steps.spec.ts
```diff
@@ -240,4 +247,47 @@ describe('TransferSteps.vue', () => {
...
+    });
+
+    await flushPromises();
+    wrapper.find('.action-button__button').trigger('click');
+    await flushPromises();
+    expect(wrapper.vm.$data.step).toBe(1);
+    expect($raiden.transfer).toHaveBeenCalledTimes(1);
+  });
+
+  test('skip pfs selection if free pfs is found', async () => {
```
nephix: `'Should...'`

raiden-ts/src/path/actions.ts
```diff
@@ -14,7 +14,10 @@ type ServiceId = {
   serviceAddress: Address;
 };
 
-export const pathFind = createStandardAction('pathFind')<{ paths?: Paths; pfs?: PFS }, PathId>();
+export const pathFind = createStandardAction('pathFind')<
```
nephix: `typesafe-actions` v4 seem to have a new way of creating standard actions. Do you see any benefit of migrating to the new style when we create new actions or touch old ones?

raiden-ts/tests/e2e/raiden.spec.ts
```diff
@@ -794,17 +794,37 @@ describe('Raiden', () => {
   });
 
   describe('findPFS', () => {
-    test('fail config.pfs not in auto mode', async () => {
-      expect.assertions(2);
+    test('fail config.pfs disabled', async () => {
```
nephix: Any merit to applying `'Should...'` also to the SDK?

raiden-dapp/src/views/TransferSteps.vue
kelsos: We already discussed with @taleldayekh that this flow can get improved. The plan is to evaluate all the possible scenarios and figure out the proper flow.

raiden-dapp/tests/unit/views/transfer-steps.spec.ts
kelsos: When it comes to jasmine's `it` `Should` sounds like the natural next step. Is that a common convention for jest test naming too?

nephix: That's what we have always used in the past, I guess it has not so much to do with jest or jasmine but it's one of many naming conventions. Maybe we can decide on one in the test refactor ticket and just leave them here as they are? Here are more "patterns": https://dzone.com/articles/7-popular-unit-test-naming

PS: `test('Should fail when x y z')` also sounds pretty natural :)

kelsos: @andrevmatos it would be nice to have a consistent naming strategy in the tests accross the SDK in general it would be nice to also have your input here.

andrevmatos: On the SDK, `test` doesn't use `should`, I usually just name the test in a descriptive way, like `test('fail when amount is negative', () => ...`

kelsos: I don't mind doing this way either and since I am touching all the tests anyway it would not be much @nephix are you ok with changing the tests in the dapp to follow the same scheme?

raiden-ts/src/path/actions.ts
andrevmatos: Hmm, which new way? From [what I can see](https://github.com/piotrwitek/typesafe-actions#2-fsa-compliant-actions), this is still how it's done. Anyway,  I'd say we shouldn't touch this now, and consider #591 when we need to migrate to anything else (probably a custom implementation based on typesafe-actions and io-ts by then)

raiden-dapp/tests/unit/views/transfer-steps.spec.ts
nephix: Sure sounds good!

raiden-ts/src/path/actions.ts
nephix: We had a dependency PR https://github.com/raiden-network/light-client/pull/474#issuecomment-549709382 that tried to bump the version to 5.x which deprecates the way actions are created in 4.x (like we do currently). The "old" way still works if we use `createStandardAction` from `deprecated`

andrevmatos: Hmm, right.. didn't see the changes in new release, we may move forward then to the serializable approach. Thank you for pointing this out.

PR #612, opened 2019-11-27
BigNumbers are de|encoded as strings

andrevmatos:
Adapt requirements by https://github.com/raiden-network/raiden/pull/5390

With this, we could remove the `lossless-json` dependency and go directly with JSON utilities, as all numbers should be encoded and decoded to/from strings, but to stay in the safe side if any field changes in the future or was left behind, better safe than sorry and still use lossless library to do it.


andrevmatos:
I tested by hand and changes here are compatible with RC2, with direct and mediated transfers.

nephix:
Good job @andrevmatos

PR #636, opened 2019-11-28
Added update dialog and invalidate service worker if there's an update

### LINKED ISSUE #598: Caching issues, if a new version of the dApp gets deployed
## Steps to Reproduce
1. Connect to https://lightclient.raiden.network/
2. Use the dApp
3. Introduce some visible changes in the dApp code
4. Deploy a new version of the dApp to github pages
5. Connect to https://lightclient.raiden.network/ with the same browser as used in 1.

## Expected Result
- The changes introduced in 3. are visible

## Actual Result
- The changes introduced in 3. are not visible
- They are visible after hitting STRG+F5 (clearing the cache for this page) and a reconnecting

nephix:
Fixes #598

Tested it locally by 
1. Making a production build
2. Start local http server
3. Navigate to dapp
4. Make an update (e.g. change button color)
5. Make new production build
6. Start local http server

out

### DIFF
--- a/raiden-dapp/src/components/UpdateSnackbar.vue
+++ b/raiden-dapp/src/components/UpdateSnackbar.vue
@@ -1,0 +1,54 @@
+<template>
+  <span v-if="updateAvailable">
+    <blurred-overlay :show="updateAvailable" />
+    <v-snackbar v-model="updateAvailable" :timeout="0" color="primary">
+      {{ $t('update.available') }}
+      <v-btn dark text :loading="isUpdating" @click="update">
+        {{ $t('update.update') }}
+      </v-btn>
+    </v-snackbar>
+  </span>
+</template>
+
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator';
+import BlurredOverlay from '@/components/BlurredOverlay.vue';
+
+@Component({ components: { BlurredOverlay } })
+export default class UpdateSnackbar extends Vue {
+  isUpdating: boolean = false;
+  updateAvailable: boolean = false;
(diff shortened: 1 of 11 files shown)


raiden-dapp/src/registerServiceWorker.ts
```diff
@@ -0,0 +1,45 @@
...
+      console.log(
+        'App is being served from cache by a service worker.\n' +
+          'For more details, visit https://goo.gl/AFskqB'
+      );
+    },
+    registered(registration: ServiceWorkerRegistration) {
+      // Check periodically for updates every minute
+      setInterval(() => {
+        registration.update();
+      }, 1000 * 60);
```
nephix: This will check every minute while the dapp is running whether there's a new service worker available. I first had this at `1000`ms but that would be pretty bad e.g. if a user is running a transfer and we release an update, he would be blocked from continuing and forced to refresh the dapp.

raiden-dapp/src/components/UpdateSnackbar.vue
```diff
@@ -0,0 +1,51 @@
...
+      {{ $t('update.available') }}
+      <v-btn dark text :loading="isUpdating" @click="update">
+        {{ $t('update.update') }}
+      </v-btn>
+    </v-snackbar>
+  </span>
+</template>
+
+<script lang="ts">
+/* istanbul ignore file */
```
nephix: I've excluded this from code coverage, because it's mostly using Web API and testing it would be pretty difficult and would yield very little insight for us IMO

kelsos: It makes absolute sense. Thanks for doing that.

PR #637, opened 2019-11-28
Refactor/remove dead code

### LINKED ISSUE #611: Remove unused Token.vue component with it's tests
## Description

After the implementation of #583 the Token.vue component along with the tests and maybe some other related code is not used anymore.  

## Acceptance criteria
- No unused code should be in the repo

## Tasks
- [ ] Remove unused compoents and their tests.

kelsos:
- Removes Tokens.vue with tests
- Removes ConfirmationDialog.vue with tests
- Removes old Home.vue and moves AppCore.vue as the new home as one of them was completely redundant
- Removes `leaveTokenNetwork` from raiden-service.ts

Closes #611


raiden-dapp/src/services/raiden-service.ts
```diff
@@ -371,5 +328,3 @@ export class RaidenInitializationFailed extends Error {}
 export class FindRoutesFailed extends Error {}
 
 export class PFSRequestFailed extends Error {}
-
-export class MintTokenFailed extends Error {}
```
nephix: Nice catch!

kelsos: Btw I know I started doing this `Error` thing in the first place but in a lot of cases it doesn't make that much sense. And where it makes we could also use a separate try/catch in order to have the same result. We should certainly re-evaluate and reduce that.

Issue #371, opened 2019-10-08
I want to get a warning if I try to open a channel with 0.0 tokens


christianbrb:
## User Story
As Chris, I want to get a warning if I try to open a channel with 0.0 tokens so that I am able to use the channel when I have opened it.

kelsos:
@christianbrb what does this mean technically, that you are still allowed to open a channel without deposit but you get a warning that the channels will be unusable until you deposit?

christianbrb:
Yes :-)

christianbrb:
Closed via #647

PR #647, opened 2019-11-29
Invalidate amount input for zero values

### LINKED ISSUE #609: Don't allow Zero as a Transfer amount
## Steps to Reproduce
1. Don't set a transfer amount in the transfer screen
2. Press continue

## Expected Result
- Continue to be disabled
- Needs to be done also in the transfer steps

## Actual Result
- PFS errors because the amount is 0

nephix:
Fixes #609


nephix:
Good point, I will make it configurable

christianbrb:
@nephix @kelsos I liked it how is has been before [the last commit](https://github.com/raiden-network/light-client/pull/647/commits/150ccccf57cac5bc75df5fe6f6e123f8f36074b1). This would have fixed https://github.com/raiden-network/light-client/issues/371

nephix:
@christianbrb ok thanks for your input. I will reverse the change then unless @kelsos has some objections

PR #650, opened 2019-12-02
Derived subkey

### LINKED ISSUE #172: Add derived subkey support
## Abstract

While fully relying on ETH-JSON-RPC node-side **private keys** has the advantage of offloading private key (PK) security to the ethereum node, on the most common usecases (e.g. Metamask, Parity) it has the downside of making for a poor UX, as every signature requires an interactive prompt and user action. On current implementation, at least 2 signatures are required at first startup time for Matrix login and **up to 6** are required for a single payment sending. This even worse when using a hardware wallet (e.g. Metamask's Trezor support), as each signature will need multiple approval clicks on UI and on the physical device.
As most of the time these are signatures of opaque byte strings or even hashes, there's no point in requiring users to see it anyway.
image

### What is a subkey?

Here we define a subkey (child key) as a new private key **deterministically and secretly derived** from the main/parent user key/account. An ETH private key is 256bits/32 bytes number/string. A possible approach is to use ~~the last 32 bytes~~ the `keccak` or `sha256` hash (32 bytes) of the signature (65 bytes) of a constant (possibly per-network) string as private key.
Pseudo-code Example: 
```javascript
message = `raiden_${network.name || network.chainId}`;
subkey = web3.sha3(web3.personal.sign(message, web3.eth.accounts[0]));
// or eth_sign equivalent, depending on provider's implementation of EIP191
subwallet = new Wallet(subkey, provider);
```
Optionally, the to-be-signed constant string could contain BIG WARNINGS (which will be invariably shown to the user in Metamask sign prompt), alerting users to NOT sign this message anywhere outside a trusted Raiden client, to prevent key steal by phishing:
image

EDIT: updated to the current proposal of using the hash of the signature

## Proposal

At startup/load time, the user may be given the choice to use some eth-provider key (e.g. Metamask always expose a single key, the selected one) **directly** (as is today) or a generated subkey.
On the subkey case:
- a single signature is required at startup time to generate a local private key, which will then be used as Raiden key through the session, without needing any user interaction for signing.
- this key should not be persisted (only the address on `localStorage` object, but also no evidence of relationship between main and subkey), and the user will be given the initial prompt again on every startup.
- the UI may offer the user information about how this key is generated, steps to reproduce outside of Raiden, and possibly allow the user to download it in password-protected JSON format, for backup or usage in an external wallet (although they'll always be able to re-generate it from the main key)
- the UI may offer actions to easily transfer small amounts of ETH from the main account to this subkey, to pay for gas when needed.
- investigate: the user may not need to transfer tokens to the subkey, as possibly the main account could perform the on-chain actions related to depositing on a channel **on behalf of the subkey**, allowing the main account to actually hold the user token's stash and the normal provider transaction prompt to be used to send the deposit tx.
- the UI should offer actions to withdraw back to the main account any tokens the subkey may hold, like the ones coming from any channel close or (in the future) withdraw and receiving.

This approach is already used in other dapps which need a big number of signatures.

## Compatibility

This proposal is fully compatible with Raiden protocol, as from the other node's perspective, it's just a normal account interacting with the network, with additional perk of being able to keep the bulk of token stash/ether on a separate, not often used main account.
Additionally, it makes possible to use Raiden Light Client with hardware wallets like Trezor, as a single startup signing prompt, plus eventual on-chain deposit tx prompts are the only intervention needed from the main account.

## Security

As the key is not persisted anywhere, but derived from the main key on each session's startup and kept only on memory, there's very little risk of steal. If the application isn't trusted or its runtime was compromised somehow, user OS/browser/environment or the account itself was already compromised and there's little value on the interactive prompts anyway, while positively making the main user account use so rare it'd raise red flags if any unrequested prompt is shown.

## Implementation complexity

Surprisingly very little, as the SDK already has support for different key/signers (like ether's `JSONRPCSigner` and local `Wallet`). It would be just a matter of having the subkey generation code on `Raiden.create` (to be used as Raiden `signer`) and keeping the provider's `main_signer` around and implement giving the **alternative** of using it on behalf of subkey as deposit source account.

## Acceptance criteria
- The user can choose to use either the provider's account on Raiden light client directly or a derived subkey of it on every startup, and both accounts/options are usable and seen as different Ethereum accounts for Raiden.
- The user should be able to easily transfer ETH back and forth the main account and this subkey, to pay for gas when needed
- The user should be able to withdraw any available tokens from the subkey to the main account
- The user should be able to deposit on subkey's channels (or possibly even create them) from the main account, to avoid the need to transfer tokens to it.
- The user should be able to know how the subkey derives from the main account, have instructions on how to do it outside Raiden dApp, and be able to download the encrypted JSON subkey for backup.

## Tasks
- [ ]

andrevmatos:
Fix SDK part of #172

### DIFF
--- a/raiden-ts/src/helpers.ts
+++ b/raiden-ts/src/helpers.ts
@@ -65,7 +65,7 @@ async function genSubkey(network: Network, main: Signer) {
   const url = globalThis?.location?.origin ?? 'unknown';
   const message = `=== RAIDEN SUBKEY GENERATION ===
 
-Network: ${getNetworkName(network)}
+Network: ${getNetworkName(network).toUpperCase()}
 Raiden dApp URL: ${url}
 
 WARNING: ensure this signature is being requested from Raiden dApp running at URL above by comparing it to your browser's url bar.
--- a/raiden-ts/src/helpers.ts
+++ b/raiden-ts/src/helpers.ts
@@ -1,26 +1,26 @@
+import { Signer, Wallet, Contract } from 'ethers';
 import { Network, toUtf8Bytes, sha256 } from 'ethers/utils';
-import { Signer, Wallet } from 'ethers';
 import { JsonRpcProvider } from 'ethers/providers';
 import { Observable, from } from 'rxjs';
 import { filter, map, scan, concatMap, pluck } from 'rxjs/operators';
 import { findKey, transform, pick } from 'lodash';
 
-import { RaidenState } from '../state';
-import { ContractsInfo } from '../types';
-import { raidenSentTransfer } from '../transfers/utils';
-import { SentTransfer, SentTransfers, RaidenSentTransfer } from '../transfers/state';
-import { channelAmounts } from '../channels/utils';
-import { RaidenChannels, RaidenChannel, Channel } from '../channels/state';
-import { pluckDistinct } from '../utils/rx';
-import { Address, PrivateKey, isntNil, Hash } from '../utils/types';
-import { getNetworkName } from '../utils/ethers';
+import { RaidenState } from './state';
+import { ContractsInfo, RaidenEpicDeps } from './types';
+import { raidenSentTransfer } from './transfers/utils';
+import { SentTransfer, SentTransfers, RaidenSentTransfer } from './transfers/state';
+import { channelAmounts } from './channels/utils';
+import { RaidenChannels, RaidenChannel, Channel } from './channels/state';
+import { pluckDistinct } from './utils/rx';
+import { Address, PrivateKey, isntNil, Hash } from './utils/types';
+import { getNetworkName } from './utils/ethers';
 
-import ropstenDeploy from '../deployment/deployment_ropsten.json';
-import rinkebyDeploy from '../deployment/deployment_rinkeby.json';
-import goerliDeploy from '../deployment/deployment_goerli.json';
-import ropstenServicesDeploy from '../deployment/deployment_services_ropsten.json';
-import rinkebyServicesDeploy from '../deployment/deployment_services_rinkeby.json';
-import goerliServicesDeploy from '../deployment/deployment_services_goerli.json';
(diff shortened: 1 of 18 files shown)


raiden-ts/src/helpers.ts
```diff
@@ -72,24 +104,33 @@ export const getSigner = async (account: string | number | Signer, provider: Jso
...
   } else if (PrivateKey.is(account)) {
     // private key
     signer = new Wallet(account, provider);
+    address = signer.address as Address;
   } else {
     throw new Error('String account must be either a 0x-encoded address or private key');
   }
 
-  return signer;
+  if (subkey) {
```
nephix: Can't we skip all the stuff above if `subkey === true`?

raiden-ts/src/helpers.ts
```diff
@@ -53,16 +54,47 @@ export const getContracts = (network: Network): ContractsInfo => {
...
  * @param account - an account used for signing
  * @param provider - a provider
+ * @param subkey - Whether to generate a subkey
  * @returns a [[Signer]] or [[Wallet]] that can be used for signing
  */
-export const getSigner = async (account: string | number | Signer, provider: JsonRpcProvider) => {
+export const getSigner = async (
+  account: string | number | Signer,
+  provider: JsonRpcProvider,
+  subkey?: boolean,
```
nephix: `subKey`

raiden-ts/src/helpers.ts
andrevmatos: No, `main` key still needs to be generated/instantiated as usual. The `subkey` is for after we have a main `Signer` already, however we got it, and then derive it, put the main on `main`, and replace the _actual_ with the subkey's.

raiden-ts/src/helpers.ts
andrevmatos: No, `subkey` was the keyword used everywhere, it's not two words, but a single one representing the `subkey` flag.

Issue #218, opened 2019-07-24
Check availability of target as soon as a valid address is inserted.


kelsos:
## Description

Currently, the `getAvailability` is only checked as soon as the user presses transfer.
Check `getAvailability` as soon as the address is input, and require/validate that the available is true.

## Acceptance criteria
- Check availability as soon as a valid address is inserted.
- Show availability color-coded in the UI
- If it is available the border of the Identicon becomes green
- If it is offline
    - the border of the Identicon becomes gray
    - A validation message is shown "Target is offline"
    - The "Continue" button is disabled
- A channel cannot be opened if the channel partner is offline

## Tasks
- [ ]

christianbrb:
Is `getAvailability` checking the matrix presence?

If yes #150 could be closed?

kelsos:
Yes this will be checking the presence in matrix.

taleldayekh:
- Need to go through the user flow of the app and see at what steps the user inputs an address
- Need to decide what system or color-coding to use for distinguishing available targets from non-available ones

andrevmatos:
Related to #600, #408: the same way I don't want to be bothered by paying PFS dialog & routes if there's just direct transfer, I don't want to be presented the PFS fees dialog if the target is offline, which can be done beforehand through this issue.

christianbrb:
From @andrevmatos 
> How about 3-4s after last change in transfer in status, so if signatures keep being prompted (there’s activity) it’s delayed a little longer, and if it doesn’t change in 3-4s, may mean the transfer is getting stuck or slow, and then hide it

PR #653, opened 2019-12-03
Add availability check for addresses entered in address input

nephix:
Tackles #218


raiden-dapp/src/components/AddressInput.vue
```diff
@@ -146,13 +162,37 @@ export default class AddressInput extends Mixins(BlockieMixin) {
...
     } else if (
       !AddressUtils.isAddressLike(value) &&
       AddressUtils.isDomain(value)
     ) {
       this.resolveEnsAddress(value);
+    } else if (
+      AddressUtils.checkAddressChecksum(value) &&
+      !Object.keys(this.presences).includes(value)
+    ) {
+      this.isAddressAvailable = false;
```
nephix: @kelsos these seem to be unreachable lines, according to codecov and I'm not sure how to reach them at all. Any suggestions? The main problem I have is that mocking state in VueJS is really a pain

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -52,7 +52,10 @@ export default class RaidenService {
 
   async ensResolve(name: string): Promise<string> {
     try {
-      return await this.raiden.resolveName(name);
+      const address = await this.raiden.resolveName(name);
+      const { available } = await this.raiden.getAvailability(address);
+      this.store.commit('updatePresence', { [address]: available });
```
kelsos: Is there any specific reason why you update the presense here instead of the `getAvailability` function

nephix: Nice catch! Actually it's not necessary here to update the state. I will get rid of it and test it quickly if everything still works. The state positive (read: `available=true`) updates should all be handled here: https://github.com/raiden-network/light-client/pull/653/files#diff-ddfc85a29ab30ce1822e4b860e4305d2R121

nephix:
All done, I did let the CI rerun just to make sure everything works with latest changes from `master`

raiden-dapp/src/components/AddressInput.vue
```diff
@@ -146,13 +161,37 @@ export default class AddressInput extends Mixins(BlockieMixin) {
...
-    } else if (AddressUtils.checkAddressChecksum(value)) {
-      this.input(value);
     } else if (
       !AddressUtils.isAddressLike(value) &&
       AddressUtils.isDomain(value)
     ) {
       this.resolveEnsAddress(value);
+    } else if (
+      AddressUtils.checkAddressChecksum(value) &&
+      !Object.keys(this.presences).includes(value)
```
andrevmatos:
```suggestion
      !(value in this.presences)
```
Nitpick: more performant, as doesn't require creating an array

raiden-dapp/src/model/types.ts
```diff
@@ -8,6 +8,10 @@ export interface Token {
   readonly name?: string;
 }
 
+export interface Presence {
+  [address: string]: boolean;
```
andrevmatos: Please, don't store only the boolean, but the entire presences object, as it contains other useful information, like last time the user was seen online.

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -314,6 +323,17 @@ export default class RaidenService {
   async getUDCCapacity(): Promise<BigNumber> {
     return this.raiden.getUDCCapacity();
   }
+
+  async getAvailability(address: string): Promise<boolean> {
+    try {
+      const { available } = await this.raiden.getAvailability(address);
+      return available;
+    } catch (e) {
+      this.store.commit('updatePresence', { [address]: false });
```
andrevmatos: If you store the whole presences object, and check for its presence as suggested above, you can just omit this, as object not here indicates presence is unknown, and this request can only ever fail on first `getAvailability` request, as following ones are always cached.
```suggestion
```

raiden-dapp/src/model/types.ts
```diff
@@ -8,6 +8,10 @@ export interface Token {
   readonly name?: string;
 }
 
+export interface Presence {
```
andrevmatos: Nitpick
```suggestion
export interface Presences {
```

raiden-dapp/src/store/index.ts
```diff
@@ -126,6 +131,9 @@ const store: StoreOptions<RootState> = {
         value => value.state === ChannelState.open
       );
       return orderBy(openChannels, ['capacity'], ['desc'])[0];
+    },
+    presences: (state: RootState): Presence => {
+      return state.presences;
```
andrevmatos: Why the trivial getter? `$store.getter.presences` === `$store.state.presences`
Getters are only useful if they do something useful

raiden-dapp/src/model/types.ts
nephix: I feel like if the information isn't accessed and not part of the reqs it doesn't make sense to include it, as it can easily be extended later on, if ever needed

raiden-dapp/src/services/raiden-service.ts
nephix: I made this because of a `matrixPresenceUpdateFailed` exception (not sure of exact name right now) which gets thrown when a target is offline by the SDK

raiden-dapp/src/store/index.ts
nephix: Will remove

raiden-dapp/src/model/types.ts
nephix: Ok

raiden-dapp/src/services/raiden-service.ts
andrevmatos: This only happens if no user matching given address is found. Therefore, if address isn't in the Presences mapping, it may as well mean user is offline, so you can just swallow the error and not set anything.

raiden-dapp/src/model/types.ts
andrevmatos: Ok, I'd have written the requirements differently, to also include this as part of the new feature, but you're right, it's not there and thus it's acceptable to only include the bare minimum data you need, although it'd be more complete the other way.

raiden-dapp/src/services/raiden-service.ts
nephix: Then we would need to do the request again every time the user tries it with the same address, instead of having the info saved up in state. I guess the point of having the info in state was that the request does not happen every time?

I think it could be useful in the future as well, if we are going to persist state and display a "list of friends" or something like that, and we can quickly determine if everyone in the list is online or offline

andrevmatos: Well, they requesting again every time (the user inputs the address) is actually good, as it does a more complete search for candidate users. Again, it's not cached only if there was an error, which happens only if no user could be found matching given address (which shouldn't be common, and mostly indicate an address which didn't join yet or ever Raiden Network).

Better thinking, this isn't that crucial, I suggested more so we could keep the whole presences object (as I had suggested, but you rightfully argued it'd be better left for the future, when really needed), as constructing a "false" object is harder than just the boolean, so in that case it'd be better to just leave the entry out.
But, for the current state of the PR, this is not important. Thank you.

PR #655, opened 2019-12-03
Adds the ability to dismiss the transfer progress

### LINKED ISSUE #426: I want to be able to continue using the dApp while sending a long running transfer
## User Story
As Chris,I want to be able to continue using the dApp while sending a long running transfer so that I don't have to wait hours until the dApp is reusable.

Known drawbacks: 
- Signing messages, which come up after the payment progress window is closed.
- It might lock more tokens, than needed

## Acceptance criteria
- Just the hide button + the text
- The user does not get notified if a transfer fails or succeeds

## Wireframe

 grafik

kelsos:
Closes #426

### DIFF
--- a/raiden-dapp/src/components/RaidenDialog.vue
+++ b/raiden-dapp/src/components/RaidenDialog.vue
@@ -1,0 +1,33 @@
+<template>
+  <v-card class="raiden-dialog">
+    <v-btn icon class="raiden-dialog__close" @click="close()">
+      <v-icon>mdi-close</v-icon>
+    </v-btn>
+    <v-card-text>
+      <slot></slot>
+    </v-card-text>
+  </v-card>
+</template>
+
+<script lang="ts">
+import { Component, Emit, Vue } from 'vue-property-decorator';
+
+@Component({})
+export default class RaidenDialog extends Vue {
+  @Emit()
+  close() {}
+}
+</script>
+
+<style scoped lang="scss">
+@import '../scss/colors';
+.raiden-dialog {
+  background-color: $card-background;
+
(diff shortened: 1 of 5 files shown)


raiden-dapp/src/components/TransferProgressDialog.vue
```diff
@@ -0,0 +1,124 @@
+<template>
+  <v-card class="transfer-progress-dialog">
```
nephix: I guess at some point we should create a `RaidenDialog` component that takes a component tree as children and displays it

kelsos: It would make sense yes :) I will try making something as part of the PR.

nephix: Oh I didn't mean to extend this PR it was just in general :) i think we will have a ticket for this soon. But if you think it can be done quickly why not :)

kelsos:
@nephix I am not sure maybe we can have to separate change logs. But I will have a look at generating a changelog retrospectively for everything we did up to this point tomorrow so we can skip that at the moment.

raiden-dapp/src/components/RaidenDialog.vue
```diff
@@ -0,0 +1,33 @@
...
+      <slot></slot>
+    </v-card-text>
+  </v-card>
+</template>
+
+<script lang="ts">
+import { Component, Emit, Vue } from 'vue-property-decorator';
+
+@Component({})
+export default class RaidenDialog extends Vue {
```
nephix: 🔥

PR #656, opened 2019-12-04
Aligning content according to screen size

taleldayekh:
The content on the welcome screen is now aligned nicely for different screen sizes. On smaller screens the Raiden logo appears on top of the app name and on larger screens it appears to the left of the app name.

**Smaller Screens**
Screenshot 2019-12-04 at 12 27 44

**Larger Screens**
Screenshot 2019-12-04 at 12 29 52

### DIFF
--- a/raiden-dapp/src/components/SplashScreen.vue
+++ b/raiden-dapp/src/components/SplashScreen.vue
@@ -104,6 +104,7 @@ export default class Loading extends Vue {
   display: flex;
   height: 100%;
   padding-left: 10px;
+  white-space: nowrap;
   @media only screen and (max-width: 1263px) {
     justify-content: center;
     padding: 30px 0px 0px 0px;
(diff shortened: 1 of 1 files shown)


raiden-dapp/src/components/SplashScreen.vue
```diff
@@ -1,29 +1,31 @@
...
+      <v-col cols="8" xl="4" lg="5" md="8" sm="8">
+        <div class="splash-screen__app-name display-3">
+          {{ name }}
         </div>
-        <div class="font-weight-light text-center splash-screen__matrix_sign">
+      </v-col>
+      <v-col cols="8">
+        <div class="splash-screen__text font-weight-light text-center">
+          {{ $t('splash-screen.disclaimer') }}
+          <br />
```
kelsos: Do we really need the <br/> here? is it used to add space?

taleldayekh: I used them to add some space between the two paragraphs instead of having two divs for each and adding padding. Should I revert to the previous way?

taleldayekh:
> Thank you very much It looks good. I only have a request. Could we please change the double `<br />`? one approach would be to wrap the second line in a tag and use css to give it the space it requires.

Sure thing, will do that!

PR #664, opened 2019-12-04
Introduce stylelint & fix errors for dApp

andrevmatos:
Part of #607

PR #685, opened 2019-12-05
Enforce nesting rules in stylelint for dApp

### LINKED ISSUE #607: Update scss in components to use & instead of repeating containing class name
## Description

New styles follow the `&` notation to include the containing class and follow a nested structure.
Styles that were in the past were using a different flat style.

This causes an inconsistency on the templates

## Acceptance criteria
- All styles follow the nested structure where the sub styles start with &
- [Linting tool is in place](https://github.com/raiden-network/light-client/issues/607#issuecomment-559055830)
- All rules are in place
- Rules are enforced in the CI

## Tasks
- [ ] 

## Additional
- Check if there is a way to enforce that automatically with some tool

andrevmatos:
Fix #607 , continuation of #664

### DIFF
--- a/raiden-dapp/.stylelintrc
+++ b/raiden-dapp/.stylelintrc
@@ -1,14 +1,24 @@
 {
- "plugins": ["stylelint-scss"],
- "extends": "stylelint-config-recommended-scss",
- "rules": {
-   "no-empty-source": null,
-   "selector-pseudo-element-no-unknown": [
-     true,
-     {
-       "ignorePseudoElements": ["v-deep"]
-     }
-   ],
-   "scss/selector-nest-combinators": "always"
- }
+  "plugins": [
+    "stylelint-scss",
+    "@namics/stylelint-bem"
+  ],
+  "extends": "stylelint-config-recommended-scss",
+  "rules": {
+    "no-empty-source": null,
+    "selector-pseudo-element-no-unknown": [
+      true,
+      {
+        "ignorePseudoElements": ["v-deep"]
+      }
+    ],
+    "scss/selector-nest-combinators": "always",
+    "scss/selector-no-redundant-nesting-selector": true,
+    "plugin/stylelint-bem-namics": {
+      "patternPrefixes": [],
+      "helperPrefixes": []
+    },
+    "selector-class-pattern": "^(?:(?![^&](__|--))[\\s\\S])*$",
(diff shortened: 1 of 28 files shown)


raiden-dapp/src/components/Confirmation.vue
```diff
@@ -76,15 +76,17 @@ $background-color: #323232;
...
+    font-size: 16px;
+    line-height: 21px;
+    text-align: center;
+    border-radius: 29px;
+    margin-left: 15px;
+    margin-right: 15px;
+  }
 }
 
 .confirmation__buttons__cancel {
```
kelsos: Another part of this issue was to fix the structure for consistency and use the & like the newer css rules so this should be nested under 
```scss
.confirmation {
  &__buttons {
    &__cancel {}
  }
}
```

Could you also do that?

raiden-dapp/.stylelintrc
```diff
@@ -8,6 +8,7 @@
      {
        "ignorePseudoElements": ["v-deep"]
      }
-   ]
+   ],
+   "scss/selector-nest-combinators": "always"
  }
```
nephix: Might make sense to apply a rule that checks if correct BEM naming is used as class names

andrevmatos: Done

andrevmatos:
@kelsos enforcing nesting within BEM classes was quite hard. It doesn't seem to be very common, to the point I couldn't find how to do it through any of the common plugins out there. I managed to get a solution using regexes, which fail if any class or nested selector contains `__` or `--` not precceded by `&`, so only nested BEM compositions can contain `__` or `--`. This worked on most of the cases, with the exception of boolean functions like `:not` containing bem classes as parameters, but it was only some cases which could be disabled.
I'm not 100% confortable with the solution, but it helped enough to fix all occurrences on current codebase, which were fixed. Please, anything else should go in another issue & PR, I think this is enough for this issue.

andrevmatos:
There was a lot of things that needed to me moved around. I tried as much as possible to keep the same styles, just reorganizing them, but I see some build-time warnings about CSS. Please, double-check visually if everything is in place.

kelsos:
@andrevmatos thank you very much for your work. I will have a look in the morning to verify that everything is in place and then I will go on and merge.

Issue #706, opened 2019-12-09
Replace publishReplay operator usage in epics with deps ReplaySubject


andrevmatos:
## Description

One common pattern in epics which has contributed with cognitive complexity + 2 levels of nesting in observables chain is the usage of `publishReplay` RxJS operator to "cache" last seen value of a given observable (`action$`, `state$`, or some of their derivatives, e.g. `getPresences$`).

This makes the code hard to follow, can add some performance penalty as multiple subscriptions/ReplaySubjects are created even for values are used multiple times. 

The proposal is to have a `RaidenEpicsDeps` new member property, `latest$: ReplaySubject<{ action, state, presences, pfsList, ... }>` being a single source of a `replay(1)`/cached version of the respective reactive values. Also, a new, custom `pluckDistinct` operator combines `pluck` and `distinctUntilChanged`, to easily map this subject to any of its members reactively.

Also, `actionOutput$`, `stateOutput$` & `config$` deps members can be re-defined in function of `latest$`, so less dependencies, less code to break and an unified reactive source of truth for the latest values of these important mappings and runtime data.

This should be done in multiple PRs, as the new approach can be made in parallel to the current solution, and split PRs to implement changes gradually.

This change is purely internal, and shouldn't impact and of the interfaces or current behavior of the state machine, just simplify and make it more performant.

Will help on #658, #659 and #660 

## Acceptance criteria
- 

## Tasks
- [ ]

christianbrb:
Closing. It is covered within #658, #659 and #660

PR #707, opened 2019-12-09
Introduce RaidenEpicDeps.latest$ with cached/latest/replay(1) version of some runtime data

andrevmatos:
First part of #706

### DIFF
--- a/raiden-ts/src/epics.ts
+++ b/raiden-ts/src/epics.ts
@@ -1,4 +1,4 @@
-import { Observable, from, of } from 'rxjs';
+import { Observable, from, of, combineLatest } from 'rxjs';
 import {
   catchError,
   filter,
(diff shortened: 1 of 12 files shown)


raiden-ts/src/epics.ts
```diff
@@ -1,25 +1,56 @@
...
+/**
+ * This function maps cached/latest relevant values from action$ & state$
+ *
+ * @param action$ - Observable of RaidenActions
+ * @param state$ - Observable of RaidenStates
+ * @returns NEVER, to watch and dispose subject's subscription
+ */
+const getLatest$ = (action$: Observable<RaidenAction>, state$: Observable<RaidenState>) =>
+  combineLatest(
+    action$,
```
kelsos: `combineLatest` is giving some deprecation warnings for a while now. Checking the JSDoc I see the following:
/** @deprecated Pass arguments in a single array instead `combineLatest([a, b, c])` */

PR #711, opened 2019-12-10
Refactors transfer epic to reduce congnitive complexity

### LINKED ISSUE #659: Reduce complexity on raiden-ts/src/transfers/epics.ts
## Description

According to the code climate reports [raiden-ts/src/transfers/epics.ts](https://github.com/raiden-network/light-client/blob/master/raiden-ts/src/transfers/epics.ts) is one of the often modified classes that has a high maintenance cost due to complexity.

The purpose of this issue is to identify the possible improvements that can be done on raiden.ts in order to reduce the complexity and improve the maintainability of the class.

## Acceptance criteria
- 

## Tasks
- [x] Reduce nesting on MakeAndSignUnlock
- [x] Reduce nesting on MakeAndSignLockExpired
- [x] Reduce nesting on MakeAndSignWithdrawConfirmation
- [x] Reduce nesting on TransferSignedRetryMessageEpic
     - [x] Research, if nesting in PublishReplay can be reduced
- [x] Check if state$.pipe(publishReplay(....$.pipe( can become a utility function
- [x] Split out promise like observable utility function
- [ ] Checking the validity of messages in a function
- [x] Check, if the message is the same
- [x] Switch statement in TransferAndGenerateSignEnvelopeMessageEpic
- [x] Check, if Asserts can be used
- [ ] Adjust the duplication threshold (a little bit up)
- [x] Adjust tests

kelsos:
Closes #659

### DIFF
--- a/raiden-ts/src/transfers/epics.ts
+++ b/raiden-ts/src/transfers/epics.ts
@@ -146,10 +146,10 @@ function makeAndSignTransfer$(
   // assume paths are valid and recipient is first hop of first route
   // compose metadata from it, and use first path fee
   const metadata: Metadata = {
-      routes: action.payload.paths.map(({ path }) => ({ route: path })),
-    },
-    fee = action.payload.paths[0].fee,
-    recipient = action.payload.paths[0].path[0];
+    routes: action.payload.paths.map(({ path }) => ({ route: path })),
+  };
+  const fee = action.payload.paths[0].fee;
+  const recipient = action.payload.paths[0].path[0];
 
   const channel: Channel | undefined = state.channels[action.payload.tokenNetwork][recipient];
   // check below shouldn't fail because of route validation in pathFindServiceEpic
@@ -157,13 +157,13 @@ function makeAndSignTransfer$(
   if (!channel || channel.state !== ChannelState.open) throw new Error('not open');
 
   const lock: Lock = {
-      amount: action.payload.value.add(fee) as UInt<32>, // fee is added to the lock amount
-      expiration: bigNumberify(state.blockNumber + revealTimeout * 2) as UInt<32>,
-      secrethash: action.meta.secrethash,
-    },
-    locks: Lock[] = [...(channel.own.locks || []), lock],
-    locksroot = getLocksroot(locks),
-    token = findKey(state.tokens, tn => tn === action.payload.tokenNetwork)! as Address;
+    amount: action.payload.value.add(fee) as UInt<32>, // fee is added to the lock amount
+    expiration: bigNumberify(state.blockNumber + revealTimeout * 2) as UInt<32>,
+    secrethash: action.meta.secrethash,
+  };
+  const locks: Lock[] = [...(channel.own.locks || []), lock];
+  const locksroot = getLocksroot(locks);
+  const token = findKey(state.tokens, tn => tn === action.payload.tokenNetwork)! as Address;
 
   console.log(
     'Signing transfer of value',
@@ -245,11 +245,11 @@ function makeAndSignUnlock$(
 ) {
   const secrethash = action.meta.secrethash;
   if (!(secrethash in state.sent)) throw new Error('unknown transfer');
-  const transfer = state.sent[secrethash].transfer[1],
-    channel: Channel | undefined = get(state.channels, [
-      transfer.token_network_address,
-      transfer.recipient,
-    ]);
+  const transfer = state.sent[secrethash].transfer[1];
+  const channel: Channel | undefined = get(state.channels, [
+    transfer.token_network_address,
+    transfer.recipient,
+  ]);
   // shouldn't happen, channel close clears transfers, but unlock may already have been queued
   if (!channel || channel.state !== ChannelState.open || !channel.own.balanceProof)
     throw new Error('channel gone, not open or no balanceProof');
@@ -262,8 +262,8 @@ function makeAndSignUnlock$(
     // don't forget to check after signature too, may have expired by then
     if (transfer.lock.expiration.lte(state.blockNumber)) throw new Error('lock expired');
 
-    const locks: Lock[] = (channel.own.locks || []).filter(l => l.secrethash !== secrethash),
-      locksroot = getLocksroot(locks);
+    const locks: Lock[] = (channel.own.locks || []).filter(l => l.secrethash !== secrethash);
+    const locksroot = getLocksroot(locks);
 
     const message: Unlock = {
       type: MessageType.UNLOCK,
@@ -329,11 +329,11 @@ function makeAndSignLockExpired$(
 ) {
   const secrethash = action.meta.secrethash;
   if (!(secrethash in state.sent)) throw new Error('unknown transfer');
-  const transfer = state.sent[secrethash].transfer[1],
-    channel: Channel | undefined = get(state.channels, [
-      transfer.token_network_address,
-      transfer.recipient,
-    ]);
+  const transfer = state.sent[secrethash].transfer[1];
+  const channel: Channel | undefined = get(state.channels, [
+    transfer.token_network_address,
+    transfer.recipient,
+  ]);
   if (!channel || channel.state !== ChannelState.open || !channel.own.balanceProof)
     throw new Error('channel gone, not open or no balanceProof');
 
@@ -345,8 +345,8 @@ function makeAndSignLockExpired$(
     if (transfer.lock.expiration.gte(state.blockNumber)) throw new Error('lock not yet expired');
     else if (state.sent[secrethash].unlock) throw new Error('transfer already unlocked');
 
-    const locks: Lock[] = (channel.own.locks || []).filter(l => l.secrethash !== secrethash),
-      locksroot = getLocksroot(locks);
+    const locks: Lock[] = (channel.own.locks || []).filter(l => l.secrethash !== secrethash);
+    const locksroot = getLocksroot(locks);
 
     const message: LockExpired = {
       type: MessageType.LOCK_EXPIRED,
@@ -547,9 +547,9 @@ const transferSignedRetryMessage = (
   state$: Observable<RaidenState>,
   action: ActionType<typeof transferSigned>,
 ) => {
-  const secrethash = action.meta.secrethash,
-    signed = action.payload.message,
-    send = messageSend({ message: signed }, { address: signed.recipient });
+  const secrethash = action.meta.secrethash;
+  const signed = action.payload.message;
+  const send = messageSend({ message: signed }, { address: signed.recipient });
   // emit Send once immediatelly, then wait until respective messageSent, then completes
   return retrySendUntil$(
     send,
@@ -571,6 +571,7 @@ const transferSignedRetryMessage = (
  *
  * @param action$ - Observable of transferSigned actions
  * @param state$ - Observable of RaidenStates
+ * @param latest$ - RaidenEpicDeps latest
  * @returns Observable of messageSend actions
  */
 export const transferSignedRetryMessageEpic = (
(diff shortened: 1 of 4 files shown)


raiden-ts/src/transfers/epics.ts
```diff
@@ -852,6 +842,33 @@ export const transferSecretRequestedEpic = (
     }),
   );
 
+const secretReveal = (
```
andrevmatos: Functions returning observables may have the `$` suffix, for consistency

kelsos: Thanks :) I will change it :+1:

raiden-ts/src/transfers/epics.ts
```diff
@@ -80,101 +81,223 @@ function nextNonce(balanceProof?: SignedBalanceProof): UInt<8> {
...
+    }),
+  );
+}
+
 /**
  * Create an observable to compose and sign a LockedTransfer message/transferSigned action
- * As it's an async observable which depends on state and may return an action which changes it,
- * the returned observable must be subscribed in a serialized context that ensures non-concurrent
- * write access to the channel's balance proof (e.g. concatMap)
+As it's an async observable which depends on state and may return an action which changes it,
```
andrevmatos: Can you re-add the asterisks?

raiden-ts/src/transfers/epics.ts
```diff
@@ -80,101 +81,223 @@ function nextNonce(balanceProof?: SignedBalanceProof): UInt<8> {
...
-
-      // assume paths are valid and recipient is first hop of first route
-      // compose metadata from it, and use first path fee
-      const metadata: Metadata = {
-          routes: action.payload.paths.map(({ path }) => ({ route: path })),
-        },
-        fee = action.payload.paths[0].fee,
-        recipient = action.payload.paths[0].path[0];
+    mergeMap(([state, { revealTimeout }]) =>
+      makeAndSignTransfer$(state, action, revealTimeout, deps),
```
andrevmatos: Maybe it would be better to pass the whole `config` to the inner function,a s `state` & `action`, and unpack there, instead of a specific `revealTimeout`

raiden-ts/src/transfers/epics.ts
```diff
@@ -449,24 +521,50 @@ export const transferGenerateAndSignEnvelopeMessageEpic = (
...
-        ),
-      );
+>> => {
+  const withdrawCache = new LruCache<string, Signed<WithdrawConfirmation>>(32);
+  const latestState$ = deps.latest$.pipe(pluckDistinct('state'));
+  return action$.pipe(
+    filter(isActionOf([transfer, transferUnlock, transferExpire, withdrawReceiveRequest])),
+    concatMap(action => {
+      switch (action.type) {
+        case getType(transfer): {
```
andrevmatos: Nitpick: the brackets in the `case` are unneeded and make it less readable IMO

raiden-ts/src/transfers/epics.ts
```diff
@@ -449,24 +521,50 @@ export const transferGenerateAndSignEnvelopeMessageEpic = (
...
+          return makeAndSignTransfer(latestState$, action, deps);
+        }
+        case getType(transferUnlock): {
+          return makeAndSignUnlock(latestState$, action, deps);
+        }
+        case getType(transferExpire): {
+          return makeAndSignLockExpired(latestState$, action, deps);
+        }
+        default: {
```
andrevmatos: Maybe make it explicit on `getType` instead of `default`? We may then get TS to complain if not all cases are handled, which helps on type safety

kelsos: Sure, it seems more reasonable yes :)

raiden-ts/src/transfers/epics.ts
```diff
@@ -476,125 +574,87 @@ export const transferGenerateAndSignEnvelopeMessageEpic = (
...
+
+  return retrySendUntil$(send, action$, state$, unlockProcessedOrChannelClosed);
+};
+
 /**
  * Handles a transferUnlocked action and retry messageSend until transfer is gone (completed with
- * success or error).
- * transferUnlocked for pending Unlock's may be re-emitted on startup for pending transfer, to
- * start retrying sending the message again until stop condition is met.
+success or error).
```
andrevmatos: `*`

raiden-ts/src/transfers/epics.ts
```diff
@@ -449,24 +521,50 @@ export const transferGenerateAndSignEnvelopeMessageEpic = (
...
+      }
     }),
   );
+};
+
+const transferSignedRetryMessage = (
+  action$: Observable<RaidenAction>,
+  state$: Observable<RaidenState>,
+  action: ActionType<typeof transferSigned>,
+) => {
```
nephix: Seems to return an observable, would be great to add a descriptive return type. Should also be `transferSignedRetryMessage$` since it returns an observable, right?

raiden-ts/src/transfers/epics.ts
```diff
@@ -476,125 +574,87 @@ export const transferGenerateAndSignEnvelopeMessageEpic = (
  *
  * @param action$ - Observable of transferSigned actions
  * @param state$ - Observable of RaidenStates
+ * @param latest$ - RaidenEpicDeps latest
  * @returns Observable of messageSend actions
  */
 export const transferSignedRetryMessageEpic = (
```
nephix: `transferSignedRetryMessageEpic$`?

raiden-ts/src/transfers/epics.ts
```diff
@@ -476,125 +574,87 @@ export const transferGenerateAndSignEnvelopeMessageEpic = (
...
+      transferSignedRetryMessage(action$, latest$.pipe(pluckDistinct('state')), action),
     ),
   );
 
+const transferUnlockRetryMessage$ = (
+  action$: Observable<RaidenAction>,
+  state$: Observable<RaidenState>,
+  action: ActionType<typeof transferUnlocked>,
+  state: RaidenState,
+) => {
```
nephix: Would be great to have a descriptive return type

raiden-ts/src/transfers/epics.ts
```diff
@@ -476,125 +574,87 @@ export const transferGenerateAndSignEnvelopeMessageEpic = (
...
+success or error).
+transferUnlocked for pending Unlock's may be re-emitted on startup for pending transfer, to
+start retrying sending the message again until stop condition is met.
  *
  * @param action$ - Observable of transferUnlocked actions
  * @param state$ - Observable of RaidenStates
+ * @param deps - RaidenEpicDeps
  * @returns Observable of messageSend actions
  */
 export const transferUnlockedRetryMessageEpic = (
```
nephix: `transferUnlockedRetryMessageEpic$`

raiden-ts/src/transfers/epics.ts
```diff
@@ -476,125 +574,87 @@ export const transferGenerateAndSignEnvelopeMessageEpic = (
...
+      transferUnlockRetryMessage$(action$, latest$.pipe(pluckDistinct('state')), action, state),
     ),
   );
 
+const expiredRetryMessages$ = (
+  action$: Observable<RaidenAction>,
+  state$: Observable<RaidenState>,
+  action: ActionType<typeof transferExpired>,
+  state: RaidenState,
+) => {
```
nephix: Descriptive return type

raiden-ts/src/transfers/epics.ts
```diff
@@ -603,64 +663,57 @@ export const transferUnlockedRetryMessageEpic = (
  *
  * @param action$ - Observable of transferExpired actions
  * @param state$ - Observable of RaidenStates
+ * @param latest$ - RaidenEpicDeps latest
  * @returns Observable of messageSend actions
  */
 export const transferExpiredRetryMessageEpic = (
```
nephix: `transferExpiredRetryMessageEpic$`

raiden-ts/src/transfers/epics.ts
```diff
@@ -603,64 +663,57 @@ export const transferUnlockedRetryMessageEpic = (
...
-      ),
+  action$.pipe(
+    filter(isActionOf(transferExpired)),
+    withLatestFrom(latest$.pipe(pluckDistinct('state'))),
+    mergeMap(([action, state]) =>
+      expiredRetryMessages$(action$, latest$.pipe(pluckDistinct('state')), action, state),
     ),
   );
 
+function autoExpire$(state: RaidenState, blockNumber: number, action$: Observable<RaidenAction>) {
```
nephix: What's our strategy for documenting internal functions? Should we still add JSDoc or are inline comments enough?

JSDoc seems like a good overall strategy, because you get a quick summary of what the function does and can also export it and move it around, without having to write all the doc again. But that's just my opinion

Currently it's a bit hard to understand what `autoExpire$` does, without digging into the code

raiden-ts/src/transfers/epics.ts
```diff
@@ -852,6 +861,33 @@ export const transferSecretRequestedEpic = (
     }),
   );
 
+const secretReveal$ = (
+  state: RaidenState,
+  action: ActionType<typeof transferSecretRequest>,
+  signer: Signer,
+) => {
```
nephix: Descriptive return type

raiden-ts/src/transfers/epics.ts
```diff
@@ -861,45 +897,22 @@ export const transferSecretRequestedEpic = (
  *
  * @param action$ - Observable of transferSecretRequest actions
  * @param state$ - Observable of RaidenStates
- * @param signer - RaidenEpicDeps signer
+ * @param deps - RaidenEpicDeps
+ * @param deps.signer - RaidenEpicDeps signer
+ * @param deps.latest$ - RaidenEpicDeps latest$
  * @returns Observable of transferSecretReveal|messageSend actions
  */
 export const transferSecretRevealEpic = (
```
nephix: `transferSecretRevealEpic$`

raiden-ts/src/transfers/epics.ts
kelsos: I am not sure if the epics should also follow the same naming scheme (@andrevmatos could what do you think about that). My basic reasoning at least for the epics it is that since we are using redux-observable it is kind of implied that the Epics return an Observable/stream of actions.

nephix: Seems like an odd convention then 🤷‍♂

raiden-ts/src/transfers/epics.ts
kelsos: I am not against that but it would be better to enforce return types if we want to be consistent.

raiden-ts/src/transfers/epics.ts
kelsos: I am not sure how we should document this extracted methods. The reason for that is that they are actually documented as methods but at the caller level. So maybe we could actually link to the actual method above with something like `core logic for ***epic`

kelsos:
@andrevmatos I will finish with the return types and some JSDoc so I will probably require a quick look to the additions just to be sure.

raiden-ts/src/transfers/epics.ts
```diff
@@ -68,6 +67,9 @@ import {
   withdrawSendConfirmation,
 } from './actions';
 import { getLocksroot, getSecrethash, makeMessageId } from './utils';
+import { Signer } from 'ethers';
```
andrevmatos: For consistency, could you please move dependencies imports to the top?

raiden-ts/src/transfers/epics.ts
```diff
@@ -68,6 +67,9 @@ import {
...
+import { pluckDistinct } from '../utils/rx';
+import { RaidenConfig } from '../config';
```
andrevmatos: And upper level imports on top of friends (`./`)

raiden-ts/src/transfers/epics.ts
kelsos: Sure, but is there a way to enforce this by a rule with eslint maybe? it seems hard to remember and enforce by hand.

nephix: Yes we had this rule in another project but I forgot the name

nephix: @kelsos the rule is `eslint(import/order)` [here](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md)

Issue #657, opened 2019-12-04
Reduce complexity on raiden-ts/src/raiden.ts


kelsos:
## Description

According to the code climate reports [raiden-ts/src/raiden.ts](https://github.com/raiden-network/light-client/blob/master/raiden-ts/src/raiden.ts) is one of the often modified classes that has a high maintenance cost due to complexity. 

The purpose of this issue is to identify the possible improvements that can be done on raiden.ts in order to reduce the complexity and improve the maintainability of the class.

## Acceptance criteria
- 

## Tasks
- [x] Split Observable mappings out of the constructor 
- [x] Single line declarations + linting
- [x] Split small functions out of create factory
- [x] Writing new tests

PR #714, opened 2019-12-11
Refactor raiden.ts

nephix:
Addresses #657


raiden-ts/src/raiden.ts
```diff
@@ -309,6 +242,92 @@ export class Raiden {
...
+      },
+    );
+
+  /**
+   * Initializes the [[transfers$]] observable
+   *
+   * @param state$ - Observable of the current RaidenState
+   * @returns observable of sent and completed Raiden transfers
+   */
+  private initTransfersObservable = (
```
andrevmatos: Could these functions become `static` methods instead of private member functions? For consistency, the member functions are more useful when we need it to be a member: dynamic definition, memoization, etc.
Or maybe even move them to relevant utils or so (like `raidenSentTransfer`)

nephix: Do we need it to be static though, as it would expose it also to end users/developers and would be available as `Raiden.initTransfersObservable()` but it seems pretty useful for the outside world. Other than that I'm fine with it 👍

andrevmatos: I think `static private` is a thing, but not sure. Don't take it hard anyway, just nitpicking, can be just private as is. Thank you

nephix: I have moved them instead to `src/helper.ts` and `src/state.ts`, as suggested in #717

raiden-ts/src/raiden/helpers.ts
```diff
@@ -0,0 +1,179 @@
...
+  return signer;
+};
+
+/**
+ * Initializes the [[transfers$]] observable
+ *
+ * @param state$ - Observable of the current RaidenState
+ * @returns observable of sent and completed Raiden transfers
+ */
+export const initTransfersObservable = (
```
kelsos: @andrevmatos should this also follow the `name$` pattern?

andrevmatos: It'd be nice yes to kepe the pattern, `initTransfers$` as a function which returns an observable

PR #719, opened 2019-12-12
Introduce assert condition, use for validation in raiden.ts

andrevmatos:
After #718 
Introduce the utility and demonstrate how it can be used to narrow types in public interface.

It may not be this simple to be used in epics, since they usually require returning the fail action containing already contextual data (e.g. `action.meta`), while the `assert` just throws a generic error and therefore doesn't contain the context information for the *Failed action. 
In these cases, `if` + `return actionFailed(..., action.meta)` is still the way to go.

Part of #657

### DIFF
--- a/raiden-ts/src/utils/types.ts
+++ b/raiden-ts/src/utils/types.ts
@@ -3,7 +3,7 @@ import * as t from 'io-ts';
 import { BigNumber, bigNumberify, getAddress, isHexString, hexDataLength } from 'ethers/utils';
 import { Two, Zero } from 'ethers/constants';
 import { memoize } from 'lodash';
-import { isLeft } from 'fp-ts/lib/Either';
+import { Either, Right } from 'fp-ts/lib/Either';
 import { ThrowReporter } from 'io-ts/lib/ThrowReporter';
 
 /* A Subset of DOM's Storage/localStorage interface which supports async/await */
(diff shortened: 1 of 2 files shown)


raiden-ts/src/utils/types.ts
```diff
@@ -13,18 +13,37 @@ export interface Storage {
...
+ */
+export class AssertionError extends Error {}
+
+/**
+ * Type-safe assertion function (TS3.7)
+ *
+ * @param condition - Condition to validate as truthy
+ * @param msg - Message to throw if condition is falsy
+ */
+export function assert(condition: any, msg?: string): asserts condition {
```
nephix: At some point we should use this everywhere and enforce error codes, that the dApp and other implementers can expect and react on 👍

andrevmatos: Yes, I agree. We're coding for the minimum amount of errors to occur, and when they do, it's usually either an indicator or a wrong call (e.g. malformed address) or something trully unexpected, when most of what can be done is show it to the user and continue. But surely we could be more specific if any dev using the SDK wants to code specifically for some of these issues.

nephix: We actually have an issue for the dApp for this, and having documented error codes would be very beneficial because if we return messages we assume that we know how to explain to whatever end-user what went wrong.

With an error code the end-developer has much more control on how to react (e.g. write their own error messages) on errors, that's what I meant

Issue #654, opened 2019-12-03
'vue-i18n/no-raw-text' rule should not error on development


kelsos:
## Description

Currently, the `vue-i18n/no-raw-text`rule is considered an error based on our eslint configuration. The rule is useful in production and it helps ensure that we have no accidental text entries that are not in the en.json localization file. The problem with this approach is that it hinders prototyping since it would block the compilation.

```
Module Error (from ./node_modules/eslint-loader/index.js):
error: raw text '
          Transfering Tokens
        ' is used (vue-i18n/no-raw-text) at src/components/TransferProgressDialog.vue:8:25:
   6 |     <v-card-text>
   7 |       <v-row>
>  8 |         <v-col cols="6">
     |                         ^
   9 |           Transfering Tokens
  10 |         </v-col>
  11 |       </v-row>


```

## Acceptance criteria
- This is still an error in production, but it is a warning on development.
- Works fine in both environments

The reasoning behind this is that we are still aware of the problem but it doesn't stop the prototyping by blocking the template compilation when developing. 

In many cases, it makes sense to create the actual resources after getting the basics with some placeholder text.

## Tasks
- [ ]

PR #741, opened 2019-12-13
no-raw-test rule will only error on production

nephix:
Tackles #654


raiden-dapp/.eslintrc.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
     }],
     'vue/v-bind-style': ['error', 'shorthand'],
     'vue/v-on-style': ['error', 'shorthand'],
-    'vue-i18n/no-raw-text': ['error', {
+    'vue-i18n/no-raw-text': [process.env.NODE_ENV === 'production' ? 'error' : 'warn', {
```
kelsos: Ok now I know why it didn't work for me. I used `warning` instead of `warn`.

PR #754, opened 2019-12-17
Providing link to the getting started guide

taleldayekh:
The start screen now has a link to the getting started guide in our README so that any user who wants that information easily can access it when starting the dApp.


taleldayekh:
The linter is giving me `Using $t on 'v-html' directive can lead to XSS attack`. Any suggestions for that?

PR #759, opened 2019-12-17
Serializable actions

### LINKED ISSUE #591: Make actions serializable

andrevmatos:
Fix #591 
- [x] actions creators based on `io-ts`, typesafe
  - [x] supports optional `payload`, `meta` & `error` flag, as per `FSA` pattern
  - [x] Includes useful properties and helper funtions, to replace all functionality we currently use from `typesafe-actions`:
   - [x] `action.type` string literal, for inspection & narrowing
   - [x] `action.error` boolean literal, for error actions
   - [x] `action.codec` for direct codec access
   - [x] `action.is` member typeguard function
   - [x] `ActionType` type helper
   - [x] `isActionOf` type guard supporting single and arrays of action creators
- [x] `Error` codec (so Error objects can be [de]serialized)
- [x] Implement in all our actions, replacing `typesafe-actions`
- [x] Tests

### DIFF
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -14,7 +14,6 @@ import {
   publishReplay,
   switchMap,
 } from 'rxjs/operators';
-import { isActionOf, ActionType } from 'typesafe-actions';
 import { findKey, get, isEmpty, negate } from 'lodash';
 
 import { BigNumber, hexlify, concat } from 'ethers/utils';
@@ -48,6 +47,7 @@ import {
 import { SignatureZero, ShutdownReason } from '../constants';
 import { chooseOnchainAccount, getContractWithSigner } from '../helpers';
 import { Address, Hash, UInt, Signature } from '../utils/types';
+import { isActionOf } from '../utils/actions';
 import { fromEthersEvent, getEventsStream, getNetwork } from '../utils/ethers';
 import { encode } from '../utils/data';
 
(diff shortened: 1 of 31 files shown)


raiden-ts/src/channels/epics.ts
```diff
@@ -63,7 +63,7 @@ export const initNewBlockEpic = (
   {}: Observable<RaidenAction>,
   {}: Observable<RaidenState>,
   { provider }: RaidenEpicDeps,
-): Observable<ActionType<typeof newBlock>> =>
+): Observable<newBlock> =>
```
kelsos: Looks way nicer now :+1:

PR #763, opened 2019-12-18
Fixes wrong structure in .codeclimate.yml

kelsos:
It appears that the structure was completely wrong. now it should work fine :)


kelsos:
Thanks :+1: I did run the cli tool for the config verification but it didn't catch it.

PR #827, opened 2020-01-02
Improve config state

### LINKED ISSUE #826: Improve config state
## Description

Small improvements for config state:
- [x] Make `config.logger` able to specify `redux-logger`'s separate configs for each member of the logger group: `prevState`, `action`, `error`, nextState`, so one can chose exactly what to be shown and fine-control log spamming.
- [x] `state.config` shouldn't contain the whole `RaidenConfig`, but instead just the `overwrites`, a `Partial<RaidenConfig>` which is applied on top of default config. This way, the stored state is smaller and less prone to upgrade issues, and also allows to modify defaults on the codebase and get it reflected on already initialized states, keeping only explicitly set values between versions.

andrevmatos:
Fix #826

---
This change is [Reviewable](https://reviewable.io/reviews/raiden-network/light-client/827)

### DIFF
--- a/raiden-ts/src/config.ts
+++ b/raiden-ts/src/config.ts
@@ -26,6 +26,14 @@ import { getNetworkName } from './utils/ethers';
  * - subkey - When using subkey, this sets the behavior when { subkey } option isn't explicitly set
  *            in on-chain method calls. false (default) = use main key; true = use subkey
  */
+const logLevels = t.keyof({
+  ['']: null,
+  trace: null,
+  debug: null,
+  info: null,
+  warn: null,
+  error: null,
+});
 export const RaidenConfig = t.readonly(
   t.intersection([
     t.type({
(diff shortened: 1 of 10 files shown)


raiden-ts/src/config.ts
```diff
@@ -40,33 +48,35 @@ export const RaidenConfig = t.readonly(
...
+      ]),
       pfs: t.union([Address, t.string, t.null]),
       subkey: t.boolean,
     }),
   ]),
 );
 export interface RaidenConfig extends t.TypeOf<typeof RaidenConfig> {}
 
+export const PartialRaidenConfig = t.readonly(
+  t.intersection([t.partial(RaidenConfig.type.types['0'].props), RaidenConfig.type.types['1']]),
```
kelsos: Is that the best way we can have this? I am guessing if I get the code checked on an IDE I can probably see what everything is, but without tooling it is not clear what `types['0'].props` and `types['1']` is supposed to be?

andrevmatos: Unfortunatelly, I didn't find a way to apply `t.partial` to an already set `t.Mixed`, we need to apply it to the members. The way I did is typesafe and properly typed on an IDE, but by looking at the code, one just needs to know that this is the codec version of `Partial<RaidenConfig>`.
You see above `RaidenConfig` is an intersection of 2 objects: the required part, and the optional/partial part. These are the ones referenced there, by making the required part also optional, and intersecting it with the already partial part.

PR #838, opened 2020-01-07
Reduces the code duplication in channels and transfers reducers

### LINKED ISSUE #835: Reduce code duplication on raiden-ts/src/channels/reducer.ts and raiden-ts/src/transfers/reducer.ts
## Description

While reducing complexity on the channels and transfers reducers I come across some obvious patterns of code duplication. 

This task is about removing the duplication of code from the functions that too similar.

## Acceptance criteria
- Remove duplicate code by extracting functions or unifying code where it makes sense

## Tasks
- [ ]

kelsos:
Closes #835

### DIFF
--- a/raiden-ts/src/channels/reducer.ts
+++ b/raiden-ts/src/channels/reducer.ts
@@ -73,47 +73,26 @@ function channelOpenFailureReducer(
   return unset(path, state);
 }
 
-function channelDepositSuccessReducer(
+function channelUpdateOnchainBalanceStateReducer(
   state: RaidenState['channels'],
-  action: channelDeposit.success,
+  action: channelDeposit.success | channelWithdrawn,
 ): RaidenState['channels'] {
   const path = [action.meta.tokenNetwork, action.meta.partner];
   let channel: Channel | undefined = get(path, state);
   if (!channel || channel.state !== ChannelState.open || channel.id !== action.payload.id)
     return state;
-  if (action.payload.participant === action.meta.partner)
-    channel = {
-      ...channel,
-      partner: {
-        ...channel.partner,
-        deposit: action.payload.totalDeposit,
-      },
-    };
-  else
-    channel = {
-      ...channel,
-      own: {
-        ...channel.own,
-        deposit: action.payload.totalDeposit,
-      },
-    };
-  return set(path, channel, state);
-}
 
-function channelWithdrawReducer(
-  state: RaidenState['channels'],
-  action: channelWithdrawn,
-): RaidenState['channels'] {
-  const path = [action.meta.tokenNetwork, action.meta.partner];
-  let channel: Channel | undefined = get(path, state);
-  if (!channel || channel.state !== ChannelState.open || channel.id !== action.payload.id)
-    return state;
+  const key = channelWithdrawn.is(action) ? 'withdraw' : 'deposit';
+  const total = channelWithdrawn.is(action)
+    ? action.payload.totalWithdraw
+    : action.payload.totalDeposit;
+
   if (action.payload.participant === action.meta.partner)
     channel = {
       ...channel,
       partner: {
         ...channel.partner,
-        withdraw: action.payload.totalWithdraw,
+        [key]: total,
       },
     };
   else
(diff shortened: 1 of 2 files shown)


raiden-ts/src/channels/reducer.ts
```diff
@@ -73,55 +73,34 @@ function channelOpenFailureReducer(
...
 
-function channelWithdrawReducer(
-  state: RaidenState['channels'],
-  action: channelWithdrawn,
-): RaidenState['channels'] {
-  const path = [action.meta.tokenNetwork, action.meta.partner];
-  let channel: Channel | undefined = get(path, state);
-  if (!channel || channel.state !== ChannelState.open || channel.id !== action.payload.id)
-    return state;
+  const key = channelWithdrawn.is(action) ? 'withdraw' : 'deposit';
```
kelsos: @andrevmatos is that OK or maybe should I explicitly check for each action and throw if not?

andrevmatos: One **never** throws in reducers. At most, if some condition isn't met, it returns unchanged `state`. But you can safely assume the union is strict there, don't worry.

raiden-ts/src/channels/reducer.ts
```diff
@@ -73,71 +73,39 @@ function channelOpenFailureReducer(
...
+  const total = channelWithdrawn.is(action)
+    ? action.payload.totalWithdraw
+    : action.payload.totalDeposit;
+
   if (action.payload.participant === action.meta.partner)
     channel = {
       ...channel,
       partner: {
```
andrevmatos: Interesting approach on the `key`. Would it work for `partner`/`own` as well?

kelsos: Good point, I will adjust accordingly :)

raiden-ts/src/channels/reducer.ts
```diff
@@ -198,12 +159,12 @@ const channels = createReducer(initialState.channels)
   .handle(channelOpen.request, channelOpenRequestReducer)
```
andrevmatos: Please, can you explicitly define the types of `createReducer`s to e.g. `Reducer<RaidenState['channels'], RaidenAction>` (on all `createReducer` usage, adapting the `S` type)? I've noticed the intersection type can get quite heavy on tsc/tsserver, and is part of the slowdown you saw. I'm looking for a proper fix, but this workaround is compatible and good enough for now (it is a `Reducer` afterall).

raiden-ts/src/transfers/reducer.ts
```diff
@@ -324,13 +290,13 @@ function withdrawReceiveSuccessReducer(
 export const transfersReducer: Reducer<RaidenState, RaidenAction> = createReducer(initialState)
```
andrevmatos: Like this ^

raiden-ts/src/channels/reducer.ts
kelsos: Sure :+1:

raiden-ts/src/transfers/reducer.ts
```diff
@@ -202,52 +187,33 @@ function transferExpireSuccessReducer(
...
+  if (transferProcessed.is(action)) {
+    key = 'transferProcessed';
+  } else if (transferUnlockProcessed.is(action)) {
+    key = 'unlockProcessed';
+  } else if (transferExpireProcessed.is(action)) {
+    key = 'lockExpiredProcessed';
+  } else if (transferRefunded.is(action)) {
+    key = 'refund';
+  } else {
+    return state;
```
andrevmatos: You don't need the `else` if you check all possibilities.

kelsos: The problem is that for `switch` to work I have to work with the `action.type` instead of the `is(action)`. The const/ternary approach will be more confusing due to the multiple chained ternary operators.

Finally, I am aware that else is not logically needed, but without the `else` there `tsc` doesn't consider the checks exhaustive so it complains about using the variable before assigning any value to it.

andrevmatos: If you wanna stick with if/else, you can safely replace the last `else if` with just `else`, so to make it exhaustive. But this was just nitpicking, don't mind it. Thank you =)

PR #844, opened 2020-01-08
Adding new content and some editing

taleldayekh:


docs/sdk-docs/installing-dapp/README.md
```diff
@@ -0,0 +1,40 @@
...
+```bash
+git clone --recurse-submodules https://github.com/raiden-network/light-client.git
+```
+
+## Build the SDK
+
+Change to the `raiden-ts` directory, install the dependencies and build the SDK.
+
+```bash
+cd ../raiden-ts
```
nephix: Not `cd light-client/raiden-ts`?

docs/sdk-docs/installing-dapp/README.md
```diff
@@ -0,0 +1,40 @@
...
+npm install
+npm run build
```
nephix: I never ran `npm run build`, @kelsos is this necessary?

docs/sdk-docs/installing-dapp/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+## Install the dApp dependencies
+
+Change to the `raiden-dapp` directory and install its dependencies.
+
+```bash
+cd ../raiden-dapp
+npm install --save raiden-ts
```
nephix: Isn't `npm install` enough? `--save` adds it to `package.json` where it already is

docs/sdk-docs/installing-sdk/README.md
```diff
@@ -0,0 +1,24 @@
...
+```bash
+npm install raiden-ts
+```
+
+After installing the SDK you can simply import Raiden to your JavaScript or TypeScript project.
+
+```typescript
+import { Raiden } from 'raiden-ts';
+
+# async factory
```
nephix: `#` is not valid JavaScript

docs/sdk-docs/prerequisites/README.md
```diff
@@ -0,0 +1,9 @@
+# Prerequisites
+
+#### For building the SDK and installing the dApp you'll need:
+- Node.js 10+ installed on your system.
+- A Web3 enabled browser, i.e. Chrome with the MetaMask extension.
```
nephix: Maybe link to MetaMask for convenience

docs/sdk-docs/installing-dapp/README.md
kelsos: This assumes that you are already in the directory where you cloned the project. Maybe add the information about moving to the directory where the project was cloned after the `git clone` command

docs/sdk-docs/installing-dapp/README.md
taleldayekh: Yes, your right I've always used only npm install and it works fine.

docs/sdk-docs/installing-dapp/README.md
kelsos: It was at some point, I still run it but maybe just to be sure @taleldayekh can create a new environment and go through the steps one by one to make sure that they are needed and up to date.

taleldayekh: I can confirm that it works without `npm run build` :)

docs/sdk-docs/installing-dapp/README.md
kelsos: This was required at least once in the past due to npm link shenanigans. I am not sure if it is need anymore or if it is required only the first time you clone the project. Or the first time you start in a new environment. In my experience so far, if you set up the link once then it pretty much works normally. But if you try a new system that never had the project issues might appear.

A good way to check would be in a VM that never worked with the light client to ensure that there is no leftover setup. 

These instructions target someone that never worked with the SDK before so we have to make sure that they work in such an environment.

docs/sdk-docs/installing-dapp/README.md
nephix: I usually just do `npm install` and I'm all set for local development. I think™ `npm run build` is necessary for publishing to npm

docs/sdk-docs/installing-sdk/README.md
kelsos: Yes, we have a few parts with python comments also in the rest of the docs. It' is not only here.
All these code snippets should be updated. We should create an issue about that.

docs/sdk-docs/installing-dapp/README.md
nephix: I would make it 5yo-proof and write down the exact steps :)

Besides that even if you are already in the cloned project, `cd ../raiden-ts` would not work

docs/sdk-docs/installing-dapp/README.md
kelsos: I would still prefer it if we can verify in a clean VM that this actually works as intended when there is no previous state left. 

I had a lot of cases where something worked on my development environment properly but started failing when someone new tried to run the dApp locally in a new environment.

docs/sdk-docs/installing-dapp/README.md
taleldayekh: I will leave this one then. I usually clone the repo and install the all dependencies  from scratch each time I start something new on the LC, dunno if that is the same as installing it on another machine or VM.

nephix: I never ran it that way, but yes we should create a ticket and go through the steps on a clean machine and verify that they work (or not)

kelsos:
@taleldayekh Any change you can open an issue about verifying that the guide works as expected in a fresh environment? If you need any help with that I am able to provide assistance. 

You can try all the steps that @nephix mentioned and assuming that they work properly you can also adjust the guide accordingly. The most probable scenario is that our guide is a bit outdated but I am a little cautious since I have been bitten by such changes in the past where my existing environment skewed the results.

nephix:
I have created an issue #845

Issue #833, opened 2020-01-06
Token not shown on Select Token after disconnect


christianbrb:
## Steps to Reproduce
1. Connect to the TTT token on Goerli (https://goerli.etherscan.io/token/0x3ed0daedc3217615bde34fedd023bc81ae49251b)
2. Disconnect from TTT token
3. Open the dApp
4. Click on "Connect new token"
5. Search for TTT on Select Token screen

## Expected Result
- I am able to select the TTT Token with the address 0x3ed0daedc3217615bde34fedd023bc81ae49251b

## Actual Result
- I don't find any TTT Token 

## Acceptance Criteria
* Try and reproduce
* State backup

## Task
*

christianbrb:
Cleared the application data and the token is back :)

@andrevmatos Maybe related to #834 ?

nephix:
TTT's token address is now `0x31aA9D3E2bd38d22CA3Ae9be7aae1D518fe46043`. I remember I had similar behavior when I didn't sync the light client for a while and the address of a token network has changed

kelsos:
@nephix like I mentioned in the https://github.com/raiden-network/light-client/pull/870#discussion_r365137792 the token being updated in the token registry cannot happen.

You can register a token, in the network registry but after the TokenNetwork is deployed you cannot do any modification. The symbol is not a unique identifier for the token, only the address is. I can have  multiple TTT tokens deployed within the same TokenNetworkRegistry with different addresses, they are not the same token, they are different tokens with the same symbol (the symbol and name are not unique). If there is only one entry of the TTT token in each different state this could mean that the TokenNetworkRegistry in the two states is different, which means that we actually have two different Smart Contract deployments (two different networks).

@andrevmatos what happens when the SDK and dApp go through a contract update? The dApp will get updated along with the SDK to use the newer contracts but it can't interact with the previous contracts and state and channels? how do we handle it?

nephix:
That's kind of what I meant, I just couldn't get the wording right. Thanks for clarifying @kelsos

nephix:
At the time this can't be reproduced with the provided steps and possible solutions for this need to be discussed

andrevmatos:
The SDK's state localStorage key includes the Token Network Registry address in the name, which is the "identifier" for a network version/deployment. When it gets updated, the old state is kept there but are not accessed. For SDK, it's a new state/new account, which gets initialized, scanned, populated from scratch.
[...]

PR #870, opened 2020-01-09
Monitors only tokens of interest

### LINKED ISSUE #834: Optimize ethers events polling for several tokens
While working on https://github.com/raiden-network/light-client/issues/551#issuecomment-570220222 , I noticed that, despite the bundled script working, it was very slow. After several hours of debugging, it became clear that one issue slipped through our view: Goerli currently contains hundreds of tokens registered, mainly because of the Scenario Player, and although we're relatively optimized, we still perform at least **2 logs request per token per block**, to check if new channels were opened on each token network with us, and this load, applied on a single server, starts accumulating and slowing down every request. On MetaMask on browser, this isn't visible because Metamask requests aren't show in `Network` tab in DevTools, and are also spread accross a cloud of servers, this isn't wise to bang the service like that, and when pointing it to a single (e.g. local or hosted) server, the issue is more evident.

This can be trivially optimized to 1 request per token per block, but we need to optimize further. We shouldn't keep checking for new channels on all these token networks, as most of them aren't of interest. Instead, we should check all token networks once (at startup), and **only** on networks in which at any point there was any activity for this user (e.g. a channel opened), poll for new channels. The token list will still continue to be available, and at any point, the user can request a channel to be opened on any of them, or a token network to be scanned for channels.

Tasks:
- [x] `ChannelOpened` filter should make a single request, and filter participants on pipe
- [x] Token networks are only scanned for channels at startup
- [x] Only Token Networks of interest (e.g. with have/had channels with us) are actively monitored for new channels
- [x] The user can still act on these "passive" token networks, by requesting a channel to be opened or to scan for new channels again.

andrevmatos:
Fix #834 
- At first state initialization, all found token networks are scanned for past and future channels with us
  - If any found, we consider this a TokenNetwork of **interest**, and start monitoring it for channels from beginning (and keep monitoring on each new block on current and future sessions)
  - Otherwise, we ignore the TokenNetwork and don't monitor it (which is expensive)
- A new `Raiden.monitorToken` public method is exposed, which allows user to trigger scanning and monitoring a token network from beginning (even if state is already initialized). Even if no channel is found, this token network is now also considered of interest, and also monitored throughout current and future sessions
- All on-chain registered token networks can still be retrieved through `Raiden.getTokenList` method (which can be a bit more slow, since it needs to fetch all logs from registry, and should be cached)
- Requesting a channel to be opened in an unmonitored token network causes it to also become of interest and be monitored from now on.

TODO:
- [ ] adapt dApp to fetch and cache tokens for a new connection from `getTokenList` instead of internal state which depended on `tokenMonitored` events

PS: already initialized instances don't benefit from the optimization, as all currently registered token networks are considered of interest on old state. One needs to clean state (if no transfers were made on any possibly open channel) or manually edit localStorage on browser and remove all entries from `tokens` property, to trigger it to be populated again with only networks of interest.

---
This change is [Reviewable](https://reviewable.io/reviews/raiden-network/light-client/870)

### DIFF
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -1,4 +1,4 @@
-import { Observable, from, of, EMPTY, merge, interval } from 'rxjs';
+import { Observable, from, of, EMPTY, merge, interval, defer, concat as concat$ } from 'rxjs';
 import {
   catchError,
   filter,
@@ -11,12 +11,12 @@ import {
   groupBy,
   exhaustMap,
   first,
-  publishReplay,
-  switchMap,
+  take,
+  mapTo,
 } from 'rxjs/operators';
 import { findKey, get, isEmpty, negate } from 'lodash';
 
-import { BigNumber, hexlify, concat } from 'ethers/utils';
+import { BigNumber, hexlify, concat, defaultAbiCoder } from 'ethers/utils';
 import { Event } from 'ethers/contract';
 import { HashZero, Zero } from 'ethers/constants';
 import { Filter } from 'ethers/providers';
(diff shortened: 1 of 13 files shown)


raiden-ts/src/channels/epics.ts
```diff
@@ -62,52 +62,81 @@ export const initNewBlockEpic = (
...
-            ),
-            // monitor previously monitored tokens
-            from(Object.entries(state.tokens)).pipe(
-              map(([token, tokenNetwork]) =>
-                tokenMonitored({ token: token as Address, tokenNetwork }),
+    take(1),
+    mergeMap(state => {
+      const encodedAddress = defaultAbiCoder.encode(['address'], [address]);
+      // if tokens are already initialized, use it
+      if (!isEmpty(state.tokens))
```
nephix: What happens when token addresses in the token registry change (like in #833)?

kelsos: @nephix Token addresses in the registry don't change. If you have two tokens with the same symbol and different address they are different tokens. The only unique identifier of a token is the address. In the case of TTT appearing with different addresses between different states, my guess would be that this is a different deployment of the smart contracts which would mean that this is a different token registry, which means that this is a different "Network".

nephix: Hm ok that makes sense, thanks for clarifying 🙏

nephix:
Please create a change log entry

raiden-ts/src/channels/reducer.ts
```diff
@@ -20,17 +20,16 @@ import {
...
+  ({}, { payload }) => payload.blockNumber,
+);
 
 // state.tokens specific reducer, handles only tokenMonitored action
-function tokens(state: RaidenState['tokens'] = initialState.tokens, action: RaidenAction) {
-  if (isActionOf(tokenMonitored, action))
-    return set([action.payload.token], action.payload.tokenNetwork, state);
-  else return state;
-}
+const tokens = createReducer(initialState.tokens).handle(
```
kelsos: :D nice addition :+1:

PR #872, opened 2020-01-10
Sort module imports and group imports from same module

nephix:
Follow up to https://github.com/raiden-network/light-client/pull/711#discussion_r358809398


raiden-dapp/.eslintrc.js
```diff
@@ -59,5 +64,13 @@ module.exports = {
     'vue-i18n': {
       localeDir: './src/locales/*.json',
     },
+    'import/resolver': {
+      alias: {
+        map: [
+          ['@', path.resolve('src')],
```
nephix: This is to support
```
import Blah from '@/components/Narf'
```
in our vue components

raiden-ts/tests/.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
+	"rules": {
+		"import/order": "off",
+		"import/no-unresolved": "off"
+	}
+}
```
nephix: This turns off sorting rules for the SDK tests as some importing orders are important for the tests to succeed. There's also some mapping from `raiden-ts` imports to `src` going on which is easy in jest but not so trivial for eslint

raiden-ts/package.json
```diff
@@ -44,6 +44,7 @@
     "@types/events": "^3.0.0",
     "@types/isomorphic-fetch": "0.0.35",
     "@types/jest": "^24.0.25",
+    "@types/lodash": "^4.14.149",
```
andrevmatos: This is going to fail. It was in dependencies because some SDK module is exporting a Memoized function, and if the type is not accessible, it fails. You must either keep it in dependencies or find where this Memoized export is, and cast it to the normal function type, to prevent the emitted `.d.ts` depending on the lodash's types.

raiden-ts/package.json
```diff
@@ -74,7 +76,6 @@
     "vuepress": "^1.2.0"
   },
   "dependencies": {
-    "@types/lodash": "^4.14.149",
```
nephix: I moved this to devDependencies unless someone has a good reason for it to be in dependencies. I saw that this package also takes up quite a big of space in the dApp bundle. My assumption is that it's bundled into the dApp because it's not a dev dep

raiden-ts/package.json
nephix: Ok will revert

andrevmatos: I found it: `transport/utils.ts` and `utils/types.ts`. You can safely cast it to the normal function type, as the "memoized" part of the type isn't used (just at runtime).

nephix: Lets do it in another PR with more context

nephix:
I can do it on Tuesday or on the weekend, no big deal :) thanks guys

PR #909, opened 2020-01-16
Updated error screen to an error modal

taleldayekh:


raiden-dapp/src/components/ErrorDialog.vue
```diff
@@ -0,0 +1,99 @@
+<template>
+  <v-overlay v-if="description !== ''" class="overlay" absolute="true">
```
nephix: I think we have a `BlurredOverlay` component already, that could be used whenever an overlay is needed. That way it's more consistent. It's already used in e.g.:

> raiden-dapp/src/components/UpdateSnackbar.vue

taleldayekh: Yes, that sounds good, I'll take a look at it.

Issue #690, opened 2019-12-06
Address alignment (showed not input addresses)


christianbrb:
## Description
All addresses should appear in the same format on all screens

## Acceptance criteria (discussed not final)
- The address should be shown in a shortenend format everywhere
- The address should be copied, if I click on it
- A tool tip should be shown, if I hover over the address (`Click to copy`)

## Tasks
- [ ]

christianbrb:
@nephix Great improvement :)

One small issue:
- We now have got two copy functions within the header address. Can you remove the blue one?

![grafik](https://user-images.githubusercontent.com/35398162/72533283-303f3800-3875-11ea-843b-eee776c2dcac.png)

nephix:
I kept this as an optional parameter `icon` of the component, to make it obvious to the user that the address can quickly be copied in the header

christianbrb:
I would naturally navigate with the mouse to the address if I want to copy it.  The interface looks cleaner if there is less information/elements.

nephix:
Ok will remove

PR #911, opened 2020-01-16
Added address display component

nephix:
#690

### DIFF
--- a/raiden-dapp/src/components/TokenOverlay.vue
+++ b/raiden-dapp/src/components/TokenOverlay.vue
@@ -73,12 +73,7 @@
                     }}
                   </v-list-item-title>
                   <v-list-item-subtitle class="token-list__token-address">
-                    <v-tooltip bottom>
-                      <template #activator="{ on }">
-                        <span v-on="on">{{ token.address | truncate }}</span>
-                      </template>
-                      <span>{{ token.address }}</span>
-                    </v-tooltip>
+                    <address-display :address="token.address" />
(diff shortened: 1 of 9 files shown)


raiden-dapp/src/components/TokenOverlay.vue
```diff
@@ -124,6 +120,15 @@ export default class TokenOverlay extends Mixins(
   allTokens!: Token[];
   tokens!: TokenModel[];
 
+  handleTokenClick(tokenAddress: string) {
+    const { token } = this.$route.params;
+    if (token === tokenAddress) {
+      this.cancel();
```
kelsos: @nephix I am curious about this change, could you describe to be what it does and what problem it solves, please?

kelsos: Does it close the overlay if you try to navigate to the token you already are?

nephix: It was a bug that I found when trying to click/copy the token address in the TokenOverlay component. The `to=...` declaration [here](https://github.com/raiden-network/light-client/pull/911/files#diff-e66fa633bbb3802562485bba974d68eaL53) did not prevent the event bubbling when trying to copy the address so I had to remove it. 

And yes, the 2nd issue was that it redirected even if you have clicked on the token you are currently on (instead of just closing the overlay)

kelsos: Good finds, thank you :+1:

Issue #840, opened 2020-01-08
Activate git.legal


christianbrb:
## Description
We need to check the compliance of our licenses. Activate the [git.legal](https://docs.codeclimate.com/docs/git-legal) plug-in in the [CodeClimate.yml file](https://github.com/raiden-network/light-client/blob/master/.codeclimate.yml).

## Acceptance criteria
- 

## Tasks
- [ ]

christianbrb:
It is getting executed and no issues have been found:

eg. https://codeclimate.com/github/raiden-network/light-client/builds/1424
![grafik](https://user-images.githubusercontent.com/35398162/72429854-27753600-3791-11ea-9d13-1cb06ba83af4.png)

kelsos:
@christianbrb we can try to include a throwaway PR that tries to introduce a GPL dependency just to verify that this will stop us.

christianbrb:
@kelsos That's a nice idea :)

christianbrb:
We could use: https://github.com/zhukov/webogram

Or any other JS library: https://github.com/search?l=JavaScript&q=gpl&type=Repositories

kelsos:
That looks sufficient for the purposes of our test, thank you

PR #917, opened 2020-01-20
Don't merge this is a test!

kelsos:
Test for git.legal as per discussion on #840

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -14,10 +14,7 @@ checks:
       threshold: 100
 plugins:
     git-legal:
-      enabled: true
-      config:
-        allow_affero_copyleft: false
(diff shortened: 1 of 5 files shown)


kelsos:
@christianbrb I found out that this plugin assumes that the `package.json` is located in the root of the project.

Only if I move the `package.json` to the root dir and run the codeclimate CLI locally I get an error:

```
== package.json (1 issue) ==
26: Library `telegram-api-js` is licensed under a non-compliant license: `GNU General Public License 3.0` [git-legal]
```

Unfortunately, it doesn't seem to support multi-project repositories. We probably need to find an alternative since this does not cover our use-case. I am not aware of any bots/services performing license checking on PRs though.

PR #918, opened 2020-01-20
Adds matrix filters to avoid fetching unneeded events

### LINKED ISSUE #746: Use filters and lazy members loading to optimize matrix syncing
## Description

Although our usage of Matrix is already quite optimized through `matrix-js-sdk` state keeping, busy/crowded rooms like `discovery` and `path_finding` may include quite a sizeable state and number of events we don't really care for.

Matrix specification includes support for [filters](https://matrix.org/docs/spec/client_server/r0.6.0#api-endpoints) and [lazy members loading](https://matrix.org/docs/spec/client_server/r0.6.0#lazy-loading-room-members), to allow reducing the size of the sync payload needed to both perform initial syncing and delta updates.
We should make use of it to avoid syncing any state and events possible from these rooms.

Initial/first fetching of users/presences isn't a problem, as it uses `search_user_directory` functionality, which doesn't require room state. But later presence events needs to be investigated, as while they're not specific to a room, they are only sent for "users of interest", which are mainly "co-members" in rooms, so we need to check if filtering out events for these rooms could impact our ability to receive global presence events for users on these rooms as well.

## Acceptance criteria
- State and events from global rooms aren't ever fetched
- State and events from other rooms do
- We don't lose presence events from co-members in the global rooms

## Tasks
- [ ]

kelsos:
Closes #746 

- [x] Create filter before the first sync to filter out broadcast room events.
- [x] Ensure that presence monitoring works properly for the addresses we are interested in.
- [x] Tests

### DIFF
--- a/raiden-ts/src/transport/epics.ts
+++ b/raiden-ts/src/transport/epics.ts
@@ -132,6 +132,9 @@ function createFilter(matrix: MatrixClient, roomIds: string[]): Observable<Filte
       ephemeral: {
         not_types: ['m.receipt'],
       },
+      state: {
+        lazy_load_members: true,
+      },
     })),
     map((roomFilter: RoomFilterJson) => ({
       room: roomFilter,
--- a/raiden-ts/src/transport/epics.ts
+++ b/raiden-ts/src/transport/epics.ts
@@ -57,6 +57,7 @@ import {
   RoomFilterJson,
   FilterDefinition,
   Filter,
+  User,
 } from 'matrix-js-sdk';
 import matrixLogger from 'matrix-js-sdk/lib/logger';
 
@@ -203,6 +204,16 @@ function searchAddressPresence$(matrix: MatrixClient, address: Address) {
         } catch (err) {
           continue;
         }
+
+        const profile = matrix.getUser(user.user_id);
+        if (!profile) {
+          const newUser = new User(user.user_id);
+          newUser.setDisplayName(user.display_name);
+          matrix.store.users[user.user_id] = newUser;
+        } else {
+          profile.setDisplayName(user.display_name);
+        }
+
         yield user.user_id;
       }
     }),
--- a/raiden-ts/src/transport/epics.ts
+++ b/raiden-ts/src/transport/epics.ts
@@ -102,7 +102,7 @@ const userRe = /^@(0x[0-9a-f]{40})[.:]/i;
  */
 async function createFilter(config: RaidenConfig, server: string, matrix: MatrixClient) {
   const { pfsRoom, discoveryRoom } = config;
-  const transport = server.split('://')[1];
+  const transport = getServerName(server);
 
   const filteredRooms = [];
 
--- a/raiden-ts/src/transport/epics.ts
+++ b/raiden-ts/src/transport/epics.ts
@@ -11,6 +11,7 @@ import {
   merge,
   defer,
   concat,
+  AsyncSubject,
 } from 'rxjs';
 import {
   catchError,
@@ -55,6 +56,7 @@ import {
   RoomMember,
   RoomFilterJson,
   FilterDefinition,
+  Filter,
 } from 'matrix-js-sdk';
 import matrixLogger from 'matrix-js-sdk/lib/logger';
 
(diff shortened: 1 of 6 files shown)


raiden-ts/src/transport/epics.ts
```diff
@@ -83,6 +92,47 @@ import { getPresences$, getRoom$, roomMatch, globalRoomNames } from './utils';
...
+ * Creates and returns a matrix filter. The filter reduces the size of the initial sync by
+ * filtering out broadcast rooms, emphemeral messages like receipts etc.
+ *
+ * @param config - The {@link RaidenConfig} provides the broadcast room aliases for pfs and discovery.
+ * @param server - The matrix server used e.g., https://transport04.raiden.network is part of the alias.
+ * @param matrix - The {@link MatrixClient} instance used to create the filter.
+ */
+async function createFilter(config: RaidenConfig, server: string, matrix: MatrixClient) {
+  const { pfsRoom, discoveryRoom } = config;
+  const transport = server.split('://')[1];
```
andrevmatos: `getServerName`

kelsos: Thanks, I was pretty sure something like this existed but I had trouble locating it.
Now I am trying to figure out the whole presence thing :)

raiden-ts/src/transport/epics.ts
```diff
@@ -83,6 +92,47 @@ import { getPresences$, getRoom$, roomMatch, globalRoomNames } from './utils';
...
+
+  const filteredRooms = [];
+
+  if (pfsRoom) {
+    const { room_id: roomId } = await matrix.getRoomIdForAlias(`#${pfsRoom}:${transport}`);
```
andrevmatos: Each `getRoomIdForAlias` is an http request, which is slow. Both should be done in parallel, when needed. Besides this, I believe the global rooms's `joinRoom` returns roomId somehow, so you could use it to save a request for each, although it may require putting global joinRoom logic here as well

raiden-ts/src/transport/epics.ts
```diff
@@ -433,7 +483,13 @@ export const initMatrixEpic = (
           }),
           switchMapTo(matrix$),
           delay(1e3), // wait 1s before starting matrix, so event listeners can be registered
-          mergeMap(matrix => matrix.startClient({ initialSyncLimit: 0 })),
+          withLatestFrom(config$),
+          mergeMap(async ([matrix, config]) =>
+            matrix.startClient({
+              initialSyncLimit: 0,
+              filter: await createFilter(config, server, matrix),
```
andrevmatos: Nitpick: createFilter and startClient are two separate async calls. Put each on different mergeMap so, in case of unsubscribe/complete between both, the second isn't unconditionally called as with plain composed async calls

kelsos: Thanks that is a really good idea :)

raiden-ts/src/transport/epics.ts
```diff
@@ -83,6 +95,94 @@ import { getPresences$, getRoom$, roomMatch, globalRoomNames } from './utils';
...
+    .pipe(
+      withLatestFrom(config$),
+      mergeMap(([matrix, config]) =>
+        joinGlobalRooms(config, server, matrix).pipe(
+          mergeMap(roomIds => createFilter(matrix, roomIds)),
+          map(filter => [matrix, filter] as const),
+        ),
+      ),
+      mergeMap(([matrix, filter]) =>
+        matrix.startClient({
```
andrevmatos: This could go above, inside `joinGlobalRooms`'s pipe, replacing the `map as const` there. so you don't need to pass `matrix` down

raiden-ts/src/transport/epics.ts
```diff
@@ -83,6 +95,94 @@ import { getPresences$, getRoom$, roomMatch, globalRoomNames } from './utils';
...
+/**
+ * Creates and returns a matrix filter. The filter reduces the size of the initial sync by
+ * filtering out broadcast rooms, emphemeral messages like receipts etc.
+ *
+ * @param matrix - The {@link MatrixClient} instance used to create the filter.
+ * @param roomIds - The ids of the rooms to filter out during sync.
+ * @returns Observable of the {@link Filter} that was created.
+ */
+function createFilter(matrix: MatrixClient, roomIds: string[]): Observable<Filter> {
+  return of(roomIds).pipe(
```
andrevmatos: All of this `of + map + map` is just composing a single object, it'd be clearer to have it defined directly (or in variables, if you wanna make steps explicit). The `mergeMap + from` can then be replaced by a `defer`. This makes all these operators a single `defer`, which is more concise.

andrevmatos: Splitting on different operators/*Maps is useful only for async chains

PR #953, opened 2020-01-24
Updates environment variable for jest-junit

kelsos:
This should fix the reporting for the tests on CircleCI. This should make it again possible to see the tests that failed in CI without having to look at the logs.


kelsos:
And if you check CircleCI the artifact and report storing works properly now:
You can see the TESTS tab showing the following:
image

Before when the variable was not the proper one the same screen would show this instead.
image

PR #978, opened 2020-02-03
Fixes validation issues and adds debouncing to the input validation

### LINKED ISSUE #885: Continue Button stays disabled on transfer screen
## Steps to Reproduce
1. Enter a target (eg. 0x2d1bC0F9e6B755155cB379546C1546E0353E6b19)
2. Enter an amount

## Expected Result
- Continue button gets enabled

## Actual Result
- Continue button stays disabled

## Workaround
- Delete the target and paste it again
- The continue button gets enabled
- Afterward it also works if I delete the amount and enter it

kelsos:
Closes #885

### DIFF
--- a/raiden-dapp/src/components/AddressInput.vue
+++ b/raiden-dapp/src/components/AddressInput.vue
@@ -17,17 +17,6 @@
       @input="valueChanged"
       @change="valueChanged"
     >
-      <template #append>
-        <div class="address-input__status__paste-button">
-          <v-btn text @click="paste()">
-            <span
-              class="address-input__status__paste-button__text text-capitalize"
-            >
-              {{ $t('address-input.paste-button') }}
-            </span>
-          </v-btn>
-        </div>
-      </template>
       <template #prepend-inner>
         <div
           v-if="value && isChecksumAddress(value)"
@@ -285,8 +274,6 @@ export default class AddressInput extends Mixins(BlockieMixin) {
       this.$refs.address.validate();
     }
   }
-
-  paste() {}
 }
 </script>
 
@@ -435,17 +422,6 @@ export default class AddressInput extends Mixins(BlockieMixin) {
   &--untouched {
     caret-color: white !important;
     color: white !important;
-  }
-
-  &__status {
-    &__paste-button {
-      display: none;
-      margin-top: -6px;
-
-      &__text {
-        color: $primary-color;
-      }
-    }
   }
 }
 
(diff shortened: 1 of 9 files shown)


raiden-dapp/src/components/AddressInput.vue
```diff
@@ -100,41 +117,186 @@ export default class AddressInput extends Mixins(BlockieMixin) {
...
+  private checkAvailability(
+    value: ValidationResult
+  ): Observable<ValidationResult> {
+    return iif(
+      () => 'error' in value,
+      of(value),
+      of(value).pipe(
+        tap(() => (this.busy = true)),
+        switchMap(value => {
+          return iif(
```
andrevmatos: `iif`, as `defer`, is relevant if you wanna define an observable through an expression which should be evaluated only at subscription time. `switchMap`'s callback (like `mergeMap`) is called and subscribed immediately when value flows through *Map's pipeline, therefore, there's no need of an `iif` or `defer` inside a merge or switchMap, and you can plainly use an `if` or a conditional ternary operator to choose the observable that should be merged to the output.

raiden-dapp/src/components/AddressInput.vue
```diff
@@ -100,41 +117,186 @@ export default class AddressInput extends Mixins(BlockieMixin) {
...
+            () => value.value in this.presences,
+            of(this.presences[value.value]),
+            defer(() => from(this.$raiden.getAvailability(value.value)))
+          );
+        }),
+        tap(available => {
```
andrevmatos: `tap` and `map` below would run serially and would consist of only synchronous operators, I'd say there's no need for them to be split, you can just put the `tap` statements in the beginning of the `map` callback

raiden-dapp/src/components/AddressInput.vue
```diff
@@ -100,41 +117,186 @@ export default class AddressInput extends Mixins(BlockieMixin) {
...
+          this.busy = false;
```
andrevmatos: Caution with these on/off patterns inside the pipeline. It's not that critical here, but you have async operations between them, you could end up with a value stuck at `true` because of an exception, or some funny behaviors if somehow parallel runs interleave the statements. Some patterns with increasing/decreasing integers and/or `finalize`/`using` operator combinations could be safer. Not crucial here, as the side-effect is just some spinner, but it's something important to keep in mind in this kind of scenario.

nephix:
Can you please also add a changelog entry

raiden-dapp/src/components/AddressInput.vue
kelsos: Sounds reasonable I will modify accordingly thank you for the input :)

Issue #930, opened 2020-01-22
I want to see the full address on hover over 


christianbrb:
## User Story
As Chris I want to see the full address on hover over so that I see the full address before copying it.

![grafik](https://user-images.githubusercontent.com/35398162/72904644-84439400-3d2f-11ea-8430-bd09c761c143.png)

Show centered
``` 
Click to copy
0xB808877b5Ec745183bd60Ce8c89aBa83011D56f6
```
@czepluch Thanks for the feedback during the demo :)

PR #979, opened 2020-02-04
Display complete address when hovering one

nephix:
Implements #930

### DIFF
--- a/raiden-dapp/src/components/AddressDisplay.vue
+++ b/raiden-dapp/src/components/AddressDisplay.vue
@@ -47,10 +47,9 @@ export default class AddressDisplay extends Vue {
 
   copy(event: MouseEvent) {
     event.stopPropagation();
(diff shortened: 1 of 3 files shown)


raiden-dapp/src/components/AddressDisplay.vue
```diff
@@ -29,13 +31,30 @@ export default class AddressDisplay extends Vue {
   copied: boolean = false;
   private timeout: number = 0;
 
+  selectAddress(node: Node): void {
```
kelsos: Interesting, when I was looking for a way to copy text all the articles I could find suggested a hidden text-area. I don't remember seeing this. 

I would guess this works on every modern browser right?

nephix: Yes worked in Firefox and Chrome, probably in Safari too but Metamask doesnt ;)

It would only require a workaround for IE but we dont need to support it 🎉

Issue #692, opened 2019-12-06
Align modal screens


christianbrb:
## Description
Currently the modal screens have got different styles. They should get aligned.

## Acceptance criteria
- 

## Tasks
- [ ]

christianbrb:
## Collection of "modal" screens
without error screens -> covered in #689

### 1. 
![grafik](https://user-images.githubusercontent.com/35398162/72416636-757d4000-3777-11ea-96cc-9b907c3bcf2e.png)

### 2. 
![grafik](https://user-images.githubusercontent.com/35398162/72416817-cc831500-3777-11ea-9d6b-c404bf4250d9.png)

### 3. 
![grafik](https://user-images.githubusercontent.com/35398162/72417439-2c2df000-3779-11ea-9339-495d7a9bd580.png)

### 4. 
![grafik](https://user-images.githubusercontent.com/35398162/72417985-40262180-377a-11ea-822c-0148719c43fc.png)

### 5. 
![grafik](https://user-images.githubusercontent.com/35398162/72418680-96479480-377b-11ea-824f-316c9b83e71d.png)

### 6.
![grafik](https://user-images.githubusercontent.com/35398162/72418783-cbec7d80-377b-11ea-9a36-977c63131bec.png)

nephix:
Looks like it got finished in #977, thanks a lot @taleldayekh 🙏

taleldayekh:
All modals are now aligned 🥳

PR #1004, opened 2020-02-07
Adds information about the current udc deposit progress

### LINKED ISSUE #640: Explain waiting and signing in UDC minting and deposit dialogue to the user
## Description
Tell the user in the UDC minting and deposit dialogue, that he has to
- Describe the action, which is happening while the spinner is turning
- This text should be put on every step to the button
    1. Mint SVT tokens
    2. Allow UDC to access SVT tokens
    3. Deposit SVT tokens to the UDC

grafik

## Acceptance criteria
- Have a message at the bottom explaining what is happening when the spinner is shown.

## Tasks
- [ ]

kelsos:
Closes #640 

Also applies modifications to MintDepositDialog for the #692  

[skip ci]

### DIFF
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -20,7 +20,7 @@ import { ServiceRegistryFactory } from './contracts/ServiceRegistryFactory';
 import { CustomTokenFactory } from './contracts/CustomTokenFactory';
 import { UserDepositFactory } from './contracts/UserDepositFactory';
 
-import { ContractsInfo, Deposit, OnChange, RaidenEpicDeps } from './types';
+import { ContractsInfo, EventTypes, OnChange, RaidenEpicDeps } from './types';
 import { ShutdownReason } from './constants';
 import { RaidenState, getState } from './state';
 import { RaidenConfig, makeDefaultConfig, PartialRaidenConfig } from './config';
@@ -933,7 +933,7 @@ export class Raiden {
    */
   public async depositToUDC(
     amount: BigNumberish,
-    onChange?: OnChange<Deposit, { txHash: string }>,
+    onChange?: OnChange<EventTypes, { txHash: string }>,
     { subkey }: { subkey?: boolean } = {},
   ): Promise<Hash> {
     assert(!subkey || this.deps.main, "Can't send tx from subkey if not set");
@@ -960,7 +960,7 @@ export class Raiden {
     if (!approveReceipt.status) throw new Error('Approve transaction failed.');
 
     onChange?.({
-      type: Deposit.APPROVED,
+      type: EventTypes.APPROVED,
       payload: {
         txHash: approveTx.hash as Hash,
       },
@@ -975,7 +975,7 @@ export class Raiden {
     if (!depositReceipt.status) throw new Error('Deposit transaction failed.');
 
     onChange?.({
-      type: Deposit.DEPOSITED,
+      type: EventTypes.DEPOSITED,
       payload: {
         txHash: depositTx.hash as Hash,
       },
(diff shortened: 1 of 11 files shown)


raiden-ts/src/raiden.ts
```diff
@@ -925,13 +925,15 @@ export class Raiden {
...
    * @param amount - The amount to deposit on behalf of the target/beneficiary.
+   * @param depositing - Called before the deposit function is called
    * @param options - tx options
    * @param options.subkey - By default, if using subkey, main account is used to send transactions
    *    Set this to true if one wants to force sending the transaction with the subkey
    * @returns transaction hash
    */
   public async depositToUDC(
     amount: BigNumberish,
+    depositing?: () => void,
```
andrevmatos: I think having a callback for a specific step is not a good idea, this is going to be too specific and require multiple callbacks for multi-step methods. I'd say it should instead be like `onChange?: (event { type: 'APPROVED', payload: { txHash: string } }) => void` or some other type-safe and well-described event-like callback.

kelsos: I just did it because it was the simplest thing to do and our discussion on how to approach that is still ongoing but the point is valid indeed. Thanks for the suggestion it sounds really good.

kelsos:
@nephix @taleldayekh @andrevmatos I am removing the fullscreen from the `raiden-dialog` and creating another issue to restructure the channels list component/sub-components. I did start some work on it but it would make the PR way bigger so I am gonna create another issue and do it in a new PR.

raiden-ts/src/raiden.ts
kelsos: @andrevmatos could you have a look to the newest implementation

raiden-ts/src/types.ts
```diff
@@ -49,3 +49,15 @@ export interface RaidenEpicDeps {
   userDepositContract: UserDeposit;
   main?: { signer: Signer; address: Address };
 }
+
+export interface ChangeEvent<T, P> {
```
andrevmatos: Then you can just add the requirement, which makes it also compatible with the string literals from enums
```suggestion
export interface ChangeEvent<T extends string, P> {
```

raiden-ts/src/types.ts
```diff
@@ -49,3 +49,15 @@ export interface RaidenEpicDeps {
...
+  readonly type: T;
+  readonly payload: P;
+}
+
+export type OnChange<T, P> = (event: ChangeEvent<T, P>) => void;
+
+export enum Deposit {
+  APPROVED,
```
andrevmatos: May I suggest using string literals for enums? Easier to work on JS and more consistent with the events/actions as well
```suggestion
  APPROVED = 'APPROVED',
```

raiden-ts/src/types.ts
```diff
@@ -49,3 +49,15 @@ export interface RaidenEpicDeps {
...
+  DEPOSITED,
```
andrevmatos:
```suggestion
  DEPOSITED = 'DEPOSITED',
```

raiden-ts/src/raiden.ts
```diff
@@ -925,15 +925,15 @@ export class Raiden {
...
+   * @param onChange - callback providing notifications about state changes
    * @param options - tx options
    * @param options.subkey - By default, if using subkey, main account is used to send transactions
    *    Set this to true if one wants to force sending the transaction with the subkey
    * @returns transaction hash
    */
   public async depositToUDC(
     amount: BigNumberish,
-    depositing?: () => void,
+    onChange?: OnChange<Deposit, { txHash: string }>,
```
andrevmatos: Nice!

raiden-ts/src/types.ts
```diff
@@ -49,3 +49,15 @@ export interface RaidenEpicDeps {
...
 }
+
+export interface ChangeEvent<T, P> {
+  readonly type: T;
+  readonly payload: P;
+}
+
+export type OnChange<T, P> = (event: ChangeEvent<T, P>) => void;
+
+export enum Deposit {
```
andrevmatos: Maybe it'd be useful to name it not only an enum for Deposit steps, but for overall callback events types, then it'd be already a beginning for general callbacks for the other multistep operations as well

PR #1005, opened 2020-02-07
Bumps version 0.2.2 -> v0.3.0

kelsos:


kelsos:
I am not sure about it does it make sense to have the two versions tied?

andrevmatos:
I'd say they don't need to be absolutely tied, specially regarding patch, but minor ones, as it includes relatively big changes on the dApp as well, it'd be interesting to sync them on minor & major when it makes sense

PR #1007, opened 2020-02-07
Adds new modal design to close and settle confirmation for channels

taleldayekh:


raiden-dapp/tests/unit/components/channel-list.spec.ts
```diff
@@ -119,181 +107,134 @@ describe('ChannelList.vue', () => {
...
+    test('settle is called successfully', async () => {
+      // @ts-ignore
+      await wrapper.vm.settle();
+      jest.advanceTimersByTime(2000);
+
       expect(raiden.settleChannel).toHaveBeenCalledTimes(1);
       expect(raiden.settleChannel).toHaveBeenCalledWith(
-        TestData.settlableChannel.token,
-        TestData.settlableChannel.partner
+        TestData.mockChannelArray[0].token,
```
kelsos: Was there any specific reason for this change? Did the tests not pass when using the specific channel?

raiden-dapp/tests/unit/components/channel-list.spec.ts
```diff
@@ -119,181 +107,134 @@ describe('ChannelList.vue', () => {
...
   describe('close a channel', () => {
-    test('close the channel when the user confirms', async () => {
-      raiden.closeChannel = jest.fn().mockReturnValue(null);
+    beforeEach(() => {
+      raiden.closeChannel = jest.fn();
+    });
+
+    test('display confirmation dialog when close is pressed', async () => {
       wrapper.find('#channel-278').trigger('click');
-      await wrapper.vm.$nextTick();
```
kelsos: Is there any specific reason that the call to the function is removed here? `$nextTick()` and `$nextTick` are not the same. One is calling the function and the other is the function.

taleldayekh: No, that must be an error on my side, I will revert it.

raiden-dapp/tests/unit/components/channel-list.spec.ts
taleldayekh: I believe I made that change to not have to use `settlableChannel` tokens for all tests. Because it felt weird to grab the values for settlable channel tokens for deposit and close as well. Let me know if that was a faulty move from my side.

kelsos: The basic problem with this change is that while the tests pass, they are semantically wrong. The element you are using to test the settle function is an open channel. 

Before analyzing my thought, I will start with a small nitpick. I think that `TestData.settleableChannel` gives me more information about the channel compared to `mockChannelArray[0]` that is the reason I used that instead, but that is a matter of personal opinion. 

What I see as a problem in this change is that the test is semantically wrong. `mockChannelArray[0]` has an open channel that you are using to test settle. If this was the actual SDK this would not be possible. In this case, the behavior we are mocking is not the same as the one of the SDK.

I would say with a certain degree of certainty that the `settleableChannel` was not used in the deposit. The integration tests would not allow you to press deposit in a settleable channel. The open channel was used instead in these tests, to keep the semantics as close to the real SDK behavior as possible.

Also, the deposit test was using the `openChannel` but it seems that it was also replaced with `mockChannelArray[0]` from what I can see. 

image

taleldayekh: Thank you for clearifying @kelsos . That was more than I knew/considered when making the changes. :) I will adapt the tests accordingly, no problem.

Issue #613, opened 2019-11-27
Implement confirmation blocks


andrevmatos:
## Description

Currently, our blockchain interface through `ethers` treats any event as final, which could not be true if a reorg occurs. Raiden also don't see an event as confirmed before a given number of blocks (currently, default of 5).

Fortunately, our Rx/redux architecture provides a strong ground to support it in a selective manner, as some events may require confirmation and some don't.

This can use #591 between pending and confirmation events actions. Could also make use of #357 to update users about confirmation progress.

TODO: expand implementation description

## Acceptance criteria
- Blockchain actions are emitted twice: on mining and upon confirmation, iff it was not removed due to a reorg
- On confirmation checks, need to take into account that the block in which the tx was registered could have changed since first mining, if a reorg took place
- Pending actions are serialized to state and cleaned upon confirmation; on restarts, pending actions can still be confirmed
- Try to take confirmation into account only when needed, as it slows SDK, e.g:
  - deposit actions should wait for confirmation, as partners aren't going to accept transfers anyway before it, but approves don't, as deposit will only be mined/confirmed if deposit is registed.
  - In the future (when we can receive/mediate), Secrets could be registered/detected as soon as they're seen on-chain, so we "know" the secret when/if they're revealed, even if the block is reorg'ed, but should wait/ensure confirmation if need to take actions depending on it being registered on-chain
- number of confirmation blocks is configurable

## Tasks
- [ ]

christianbrb:
@andrevmatos to describe the design / architecture / tasks

PR #1009, opened 2020-02-10
Implement tx confirmation for channelOpen

andrevmatos:
Part of #613 
Todo in a later PR, is to extend this method to all on-chain transactions.
First time the on-chain transaction is seen, it gets dispatched with `payload.confirmed` property present but set as `undefined`, meaning pending. Reducer will persist/serialize it to `state.pendingTx` (thanks to #759).

Then, on each block, the possibly confirmed actions (when more than `config.confirmationBlocks` have passed since first/pending action) gets fetched with `getTransactionReceipt`, and if the transaction is still present on-chain, it gets re-emitted with `confirmed=true`, which also removes it from the pendintTx state.
If it isn't present, on each block until `2 * confirmationBlocks`, check again. If it still isn't there/confirmed after this timeout, consider the transaction as having been removed by a chain reorg, and re-emit it, now with `confirmed=false`, which also removes it from `pendingTx` and rejects any pending promise.
.
The confirmable actions always being emitted twice (first as pending, later as confirmed or removed) enables each part of the codebase to chose which actions/events to react to, be it first time seeing a tx event, or upon confirmation: e.g. the transfer epics can react early to `channelClose` and avoid accepting new transfers or mediating (in the future), while accepting deposits as increasing the channel's capacity is done only after it has been confirmed.

The `confirmationEpic` reacting on state ensures this algorithm is safe upon shutdown and later client restart, even with to-be-confirmed actions being able to be picked up again after much time (as long as receipt can be fetched by eth client, which may not be the case without full or even archiving node with enormous block ranges, but we'd consider this as an unexpected edge case).

### DIFF
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -891,6 +891,33 @@ export const channelSettleableEpic = (
     }),
   );
 
+function checkPendingAction(
+  action: ConfirmableAction,
+  provider: RaidenEpicDeps['provider'],
+  blockNumber: number,
+  confirmationBlocks: number,
+): Observable<ConfirmableAction> {
+  return defer(() => provider.getTransactionReceipt(action.payload.txHash)).pipe(
+    map(receipt => {
+      if (receipt?.confirmations !== undefined && receipt.confirmations >= confirmationBlocks) {
+        return {
+          ...action,
+          // beyond setting confirmed, also re-set blockNumber,
+          // which may have changed on a reorg
+          payload: { ...action.payload, txBlock: receipt.blockNumber!, confirmed: true },
+        };
+      } else if (action.payload.txBlock + 2 * confirmationBlocks < blockNumber) {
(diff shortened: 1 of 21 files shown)


raiden-ts/src/channels/epics.ts
```diff
@@ -878,3 +890,56 @@ export const channelSettleableEpic = (
...
+    // exhaust will ignore blocks while concat$ is busy
+    exhaustMap(([blockNumber, pendingTxs, confirmationBlocks]) =>
+      concat$(
+        ...pendingTxs
+          // only txs/confirmable actions which are more than confirmationBlocks in the past
+          .filter(a => a.payload.txBlock + confirmationBlocks <= blockNumber)
+          .map(a =>
+            defer(() => provider.getTransactionReceipt(a.payload.txHash)).pipe(
+              map(receipt => {
+                if (
```
kelsos: nitpick: any chance we could move the map a to function to avoid nesting?

andrevmatos: Good catch. Will do, thanks!

PR #1011, opened 2020-02-11
Adds new modal design to open channel

taleldayekh:


raiden-dapp/src/components/ChannelDepositDialog.vue
```diff
@@ -2,7 +2,7 @@
   <raiden-dialog
     class="channel-deposit"
     :visible="visible"
-    :fullscreen="true"
+    :fullscreen="false"
```
nephix: Have you checked if that looks as intended? There can be weird behavior if there is a parent element in the DOM with `position: relative;`

raiden-dapp/src/components/ConfirmationDialog.vue
```diff
@@ -1,95 +1,67 @@
...
-        @click="confirm()"
-      >
-        {{ positiveAction }}
-      </v-btn>
-    </v-row>
-  </v-row>
+  <raiden-dialog
+    class="confirmation"
+    :visible="visible"
+    :fullscreen="false"
```
nephix: Have you checked if that looks as intended? There can be weird behavior if there is a parent element in the DOM with `position: relative;`

raiden-dapp/src/components/ConfirmationDialog.vue
```diff
@@ -1,95 +1,67 @@
...
+    <v-card-text>
+      <v-row align="center" justify="center">
+        <span>
+          <slot></slot>
+        </span>
+      </v-row>
+    </v-card-text>
+
+    <v-card-actions>
+      <v-row align="center" justify="center">
```
nephix: Does it need the `v-row` and `v-col` here?

raiden-dapp/src/components/ConfirmationDialog.vue
```diff
@@ -1,95 +1,67 @@
...
+        <v-col>
+          <action-button
+            :id="`confirm-${identifier}`"
+            :enabled="true"
+            :text="positiveAction"
```
nephix: Since we only have one button, we treat click on X and outside of the dialog as cancel?

raiden-dapp/src/components/ConfirmationDialog.vue
```diff
@@ -1,95 +1,67 @@
...
+            @click="confirm()"
```
nephix: Should the confirm button be `sticky`?

raiden-dapp/src/components/ConfirmationDialog.vue
kelsos: I would say that seems like the most reasonable approach.

raiden-dapp/src/components/ConfirmationDialog.vue
```diff
@@ -1,95 +1,67 @@
...
   @Prop({ required: true })
   positiveAction!: string;
+  @Prop({ required: true })
+  identifier!: number;
+
   @Emit()
-  public cancel() {}
+  cancel() {}
   @Emit()
   public confirm() {}
```
kelsos: nitpick: since we are removing the public property from cancel maybe we should remove it from confrirm too.

raiden-dapp/src/components/ConfirmationDialog.vue
taleldayekh: It should be set to `true`. I think it's a leftover from my previous PR.

raiden-dapp/src/components/ConfirmationDialog.vue
taleldayekh: Apparently not, it seems to work perfectly fine without it. :) I should look over the other dialogs as well then and make sure I'm not using redundant `v-row` and `v-col`.

raiden-dapp/src/components/ConfirmationDialog.vue
taleldayekh: I wanted to try how it feels/looks to have the small button for everything that is not the main dApp screen.

raiden-dapp/src/components/ConfirmationDialog.vue
taleldayekh: The other ones seem to be giving me some weird behavior in the layout so I will leave them as-is for now and only change the `ConfirmationDialog`.

PR #1013, opened 2020-02-11
Make channelDeposit & channelWithdrawn confirmable actions

andrevmatos:
On top of #1009 
Part of #613


raiden-ts/src/actions.ts
```diff
@@ -57,3 +57,36 @@ export const RaidenEvents = [
...
+ *     txBlock: number;
+ *     confirmed: undefined | boolean;
+ *   };
+ *   meta: any;
+ * }
+ */
+export const ConfirmableActions = [
+  ChannelsActions.channelOpen.success,
+  ChannelsActions.channelDeposit.success,
+  ChannelsActions.channelWithdrawn,
```
nephix: So close/settle aren't confirmable or just not yet implemented as comfirmable actions?

kelsos: Well the PR description says ChannelDeposit and channelWithdraw and channelOpen was done in the previous PR. So the close, settle would probably go in a new PR.

nephix: So that means all our actions are confirmable? (That was really my question, I just danced around it)

kelsos: Most of the onchain actions that deal with balances I think should be confirmable. I remember discussing that, for example, the on-chain secret registration would not be confirmable as it would be better to know as fast as possible. But with the balances, we need to be sure that the balance changes are finalized.

nephix: Makes sense, thank you 🙏

andrevmatos: All on-chain events should be confirmable. Then each epic or reducer can chose to react on pending or confirmed event only. I'm going to implement `close` & `settle` on the next PR, just splitting as we've talked about =)

nephix: Yes I know, it was more of a general question and not specifically related to close/settle. Thank you 🙏

Issue #1000, opened 2020-02-06
Implement errors in SDK


kelsos:
## Description
The SDK should handle errors in api/blockchain and provide a unified way to expose them to the user of the SDK. 

## Acceptance criteria
- `Error()` is handled in a unified way

## Tasks
- [ ]

PR #1014, opened 2020-02-12
Implement SDK error handling

nephix:
Implements #1000 (🎉)

Introduces a new `RaidenError` class which has an additional `code` property.

Example usage:
```typescript
throw new RaidenError(ErrorCodes.PFS_DISALBED, [{ someDetail: 'important infos' }])
```

In `RaidenError` the `message` would be the error message targeted towards developers, while the `code` would be used display error messages to the user by e.g. hooking it up to i18n.

The new error messages in t he `ErrorCodes` enum should have a meaningful prefix, e.g. `PFS_...` for path related things or `XFER_...` for transfer related things.

The `details` prop is meant to be used for additional infos related to the error, e.g. transaction IDs or PFS URLs which couldn't be parsed.


raiden-ts/src/utils/error.ts
```diff
@@ -0,0 +1,26 @@
+export type ErrorDetail = { [key: string]: string };
+
+export default abstract class RaidenError extends Error {
+  code: string;
+  details?: ErrorDetail[];
+
+  constructor(message: string, detail?: ErrorDetail[] | ErrorDetail) {
```
andrevmatos: Suggestion: this `ErrorDetail` indexed mapping isn't type-safe, `details` don't have type hints for the types. Also, I'm not sure what's the purpose of making it an indexed mapping and an array simultaneously. I'd say only one of them would make sense (preferably something closer to the first).

Instead, I'd suggest something like this:
```
class RaidenError<D> extends Error {
  details?: D;

  constructor(message?: string, details?: D) {
    super(message);
    this.details = details;
  }
}

class PfsError extends RaidenError<{ pfs: string }> {}
// now, PfsError's details is typesafe
throw new PfsError('Err1', { pfs: 'bar' }); // correct
throw new PfsError('Err2', { pfs: 123 }); // error
```
Also, I'm not sure about the purpose of the `addDetail` method, as usually error objects are mostly read-only and defined at instantiation time with all details included.

The above approach is serializable as per `ErrorCodec` (which could/should also be moved to here), although the details wouldn't be serializable. For that, you would need to make `RaidenError` and/or derivatives codec-based, or at least the details and have some kind of type mapping somewhere to find the right class and use it to decode/deserialize/instantiate the error object.

nephix: > Suggestion: this ErrorDetail indexed mapping isn't type-safe

I don't see any reason for a need to be type safe, can you come up with scenarios?

> I'm not sure what's the purpose of making it an indexed mapping and an array simultaneously

I guess that's more a taste thing, thanks for the feedback though

> Also, I'm not sure about the purpose of the addDetail method, as usually error objects are mostly read-only and defined at instantiation time with all details included.

It's there so we can prevent having multi line throw statements where too many details are getting arranged in the constructor, as some details might need some conversions before it gets thrown.

> The above approach is serializable as per ErrorCodec (which could/should also be moved to here), although the details wouldn't be serializable. For that, you would need to make RaidenError and/or derivatives codec-based, or at least the details and have some kind of type mapping somewhere to find the right class and use it to decode/deserialize/instantiate the error object.

You shouldn't put codec types into details, that's why `ErrorDetail` was `{ [key: string]: string }`. Details should always be a simple JSON object

nephix: > I don't see any reason for a need to be type safe, can you come up with scenarios?

I saw that we already push codecs into errors/payload/actions so I guess no way around it

andrevmatos: This was due to #759 , but more for consistency, we don't serialize error actions yet (only Confirmable actions are serialized, which are success actions), and I'm not sure we'll need to. Anyway, we don't push codecs to payloads, the schema declaration contain codecs, so when needing to [de]serialize/validate it, we know where to look into. Right now, only for generic Error, and as I said, due to our usage, should be enough.
But these are two different things though: typed detail (what I suggested) and serializable/codec-based errors.
Thea bove isn't the only reason to have details typed in the errors though. Internally, we make a very limited use of the errors, but if the idea is to communicate it to the user, a valid usecase may be:
```ts
try {
  await raiden.findPFS(...);
} catch (e) {
  if (isError(e, PfsError)) { // or some proper typeguard based on `name` or `message`/`ErrorType`
    showDialog(`Could not contact PFS at ${e.detail.pfs}`); // <- type-checked
  } else {
    showDialog('A generic error ocurred');
  }
}
```

nephix: Not a fan. There're too many cases where errors need different types of infos. For example, if:
* user disables PFS, and
* there is no direct route

We throw an error at some point. Having a pfs prop there makes no sense. And there many more cases where enforcing a prop makes no sense

andrevmatos: These cases should be different Pfs error classes, with different detail types. My example was just a toy example to show how it'd help.

raiden-ts/src/utils/error.ts
```diff
@@ -0,0 +1,88 @@
+import * as t from 'io-ts';
+import { map } from 'fp-ts/lib/Either';
+import { pipe } from 'fp-ts/lib/pipeable';
+
+import { PfsErrorCodes } from '../path/errors';
+
+export const ErrorDetails = t.array(
```
andrevmatos: This codec declaration is equivalent to `{ key: string; value: string | number }[]`. If you wanna a mapping, you could use `t.record(t.union([t.string, t.number, ...]))`, which is equivalent to a proper indexed type, like `{ [k: string]: string | number | ... }`

raiden-ts/src/utils/error.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+
+export class PfsError extends RaidenError {
+  constructor(message: PfsErrorCodes, details?: ErrorDetails) {
+    super(message, details);
+    this.name = 'PfsError';
+  }
+
+  getCode(message: string): string {
```
andrevmatos: I don't understand the reason for a code. If the `code <=> message` mapping is already 1x1, they're redundant, and you can go with message only

raiden-ts/src/utils/error.ts
nephix: Ah was looking for that but took an easier right for now. Will change, thank you 🙏

raiden-ts/src/utils/error.ts
nephix: * `message` is a non-localized message for the developer for debugging/logging
* `code` is used to implement localized end-user messages

nephix: So ideally, we are going hook up each error `code` with a locale string in our i18n setup

andrevmatos: Got it. I'm not a big fan of including yet another non-standard property in these error classes, but this is just personal taste, I got your intention and it's correct from this PoV. Thanks

nephix: I also wish that `code` would be part of `Error`, never understood why it isn't 🤷‍♂

raiden-ts/src/utils/error.ts
```diff
@@ -0,0 +1,107 @@
+import * as t from 'io-ts';
+import { map } from 'fp-ts/lib/Either';
+import { pipe } from 'fp-ts/lib/pipeable';
+import { CustomError } from 'ts-custom-error';
+
+export enum ErrorCodes {
+  // Path errors
+  PFS_EMPTY_URL = 'A registered Pathfinding Service returned an empty service URL.',
+  PFS_INVALID_URL = 'A registered Pathfinding Service returned an invalid service URL.',
+  PFS_INVALID_INFO = 'Could not any valid Pathfinding services. Client and PFS versions are possibly out-of-sync.',
```
andrevmatos:
```suggestion
  PFS_INVALID_INFO = 'Could not find any valid Pathfinding services. Client and PFS versions are possibly out-of-sync.',
```

PR #1045, opened 2020-02-13
Introduce loglevel logging framework

### LINKED ISSUE #926: Introduce or implement logging framework for the SDK
## Description
This is a proposal about improving and making the logging in the SDK more manageable.
It could be done by either using some popular framework (well maintained and tested) or by writing our own utility to properly handle logging in the SDK. 

The major issue with our current approach comes from the fact that there are two different classes of logging:
- Configurable: Redux-logger, matrix-logger
- Non-configurable: console.log, console.error

Everything that is logged by the first class of loggers is easy to manage and turn off when logging is not needed.

For the second category, there is no easy way to currently turn them off and they always appear in the console. While this can be sometimes useful for debugging purposes the verbosity of the messages can be sometimes a bit overwhelming.

After deciding what direction we will follow in regards to logging it is important to decide on log levels for each `console.log` call in the SDK codebase. So messages like the signature messages `Signing x Message` can be at `info` log level for example.

## Acceptance criteria
- The number of messages varies depending on the log level
- The user of the SDK can completely turn off logging if needed.
- Implement our own logger class or use a dependency (look into logger dependencies)?
- Nice and informative logging both on node.js and on the browser.

## Tasks
- [ ] Evaluate thirdparty dependencis vs first party implementation
- [ ] Evaluate log level of various `console.log` entries

andrevmatos:
Fix #926 
Notice this comes with a slight behavior change: `config.logger`, instead of specifying the level of the emits by `redux-logger` only, now comply with the common case of specifying the minimum logging level of the whole SDK. So:
- before: `config.logger = 'info'` would make redux-logger logging be printed with `info` level on console
- now: `config.logger = 'info'` will omit anything logged with level < info, i.e. debug

Default levels for redux-logger are: `info` for action entry, `debug` for prev & next state. Default config is `info`. Therefore, on default config, anything `info` & above will be printed, and redux-logger will print only `action`, which is cleaner. To enable full debug logging, one must set with: `raiden.updateConfig({ config: { logger: 'debug' }});`.
This level also applies to matrix logger.

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -5,6 +5,7 @@
 - [#613] Implement waiting for confirmation blocks on on-chain transactions (configurable)
 
 ### Changed
+- [#926] Introduce loglevel logging framework (config.logger now specifies logging level)
 - [#1042] Support decoding addresses on messages on lowercased format
 
 ## [0.3.0] - 2020-02-07
--- a/raiden-ts/package.json
+++ b/raiden-ts/package.json
@@ -82,6 +82,7 @@
     "io-ts": "^2.1.0",
     "isomorphic-fetch": "^2.2.1",
     "lodash": "^4.17.15",
+    "loglevel": "^1.6.7",
     "lossless-json": "^1.0.3",
     "matrix-js-sdk": "^3.0.0",
     "redux": "^4.0.5",
--- a/raiden-ts/src/messages/utils.ts
+++ b/raiden-ts/src/messages/utils.ts
@@ -3,6 +3,7 @@ import { Signer } from 'ethers';
 import { keccak256, RLP, verifyMessage } from 'ethers/utils';
 import { arrayify, concat, hexlify } from 'ethers/utils/bytes';
 import { HashZero } from 'ethers/constants';
+import logging from 'loglevel';
 
 import { Address, Hash, HexString, Signature, UInt, Signed, decode, assert } from '../utils/types';
 import { encode, losslessParse, losslessStringify } from '../utils/data';
@@ -322,9 +323,10 @@ export function decodeJsonMessage(text: string): Signed<Message> {
 export async function signMessage<M extends Message>(
   signer: Signer,
   message: M,
+  { log }: { log: logging.Logger } = { log: logging },
 ): Promise<Signed<M>> {
   if (isSigned(message)) return message;
-  console.log(`Signing message "${message.type}"`, message);
+  log.debug(`Signing message "${message.type}"`, message);
   const signature = (await signer.signMessage(arrayify(packMessage(message)))) as Signature;
   return { ...message, signature };
 }
--- a/raiden-ts/package-lock.json
+++ b/raiden-ts/package-lock.json
@@ -13570,9 +13570,9 @@
       "dev": true
     },
     "loglevel": {
-      "version": "1.6.6",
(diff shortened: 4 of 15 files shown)


raiden-ts/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - [#613] Implement waiting for confirmation blocks on on-chain transactions (configurable)
 
 ### Changed
+- [#926] Introduce loglevel logging framework (config.logger now specifies logging level)
```
nephix: I think we need to add a legend for the issues as well, similarly to https://raw.githubusercontent.com/raiden-network/light-client/master/raiden-dapp/CHANGELOG.md

raiden-ts/package.json
```diff
@@ -82,6 +82,7 @@
     "io-ts": "^2.1.0",
     "isomorphic-fetch": "^2.2.1",
     "lodash": "^4.17.15",
+    "loglevel": "^1.6.7",
```
nephix: I tried to find out whether loglevel supports to log to other streams as well but seems like it's only wrapping around `console.log()` right?

Absolutely not in scope of ticket, but maybe a nice option for the future: 
In order to get higher quality bug reports it would be nice to have a way to log to memory per session as well. If then an error occurs, we could show in the frontend the debug output to the user. The user can then go ahead and copy/paste it into github, without us having to ask for the logs all the time

raiden-ts/src/messages/utils.ts
```diff
@@ -3,6 +3,7 @@ import { Signer } from 'ethers';
 import { keccak256, RLP, verifyMessage } from 'ethers/utils';
 import { arrayify, concat, hexlify } from 'ethers/utils/bytes';
 import { HashZero } from 'ethers/constants';
+import logging from 'loglevel';
```
nephix:
```suggestion
import { Logger } from 'loglevel';
```

raiden-ts/src/messages/utils.ts
```diff
@@ -322,9 +323,10 @@ export function decodeJsonMessage(text: string): Signed<Message> {
 export async function signMessage<M extends Message>(
   signer: Signer,
   message: M,
+  { log }: { log: logging.Logger } = { log: logging },
```
nephix:
```suggestion
  { log }: { log: Logger } = { log: logging },
```

raiden-ts/package.json
andrevmatos: Yes, I agree, that would be nice. `loglevel` seems to support plugins and to be very extensive, they even have [a plugin to sending logs to log server](https://github.com/artemyarulin/loglevel-serverSend), but this per-test logging or so would need to be implemented. We may want to do that eventually, thank you

raiden-ts/src/messages/utils.ts
andrevmatos: This can't be. Notice we use `logging` (the root logger) in the `signMessage` utils (as default parameter), so we need the whole default export of `loglevel`, which is the root logger.

raiden-ts/src/messages/utils.ts
andrevmatos: If you make the change you suggested, there's no such `logging` var

nephix: Whoops

kelsos:
Awesome, thanks for taking care of it :D

PR #1051, opened 2020-02-18
Restructures Channel list component and sub-components

### LINKED ISSUE #1012: Restructure Channel list component and sub-components
## Description
The plan is for the `confirmation` dialogs and `deposit` dialog to be extracted to the parent level instead of `per channel`. 

The lifecycle and buttons should be always visible now since the modal content does not appear in place in the expanded area. One other reason is that hopefully by moving everything to the view we will be able to have a blur overlay that only affects the card area and not the full screen. 

Evaluate what else does or doesn't make sense in the specific dialog.

## Acceptance criteria
- 

## Tasks
- [ ]

kelsos:
Closes #1012 

This should also take care of the overlay over the card for the dialogs there finalizing the last issue we have with the uniform experience in modals.

Also makes the channels to work as a single expand again regardless of the sub-list based on the channel state.


raiden-dapp/src/components/RaidenDialog.vue
```diff
@@ -29,9 +29,9 @@ import { Component, Emit, Vue, Prop } from 'vue-property-decorator';
 
 @Component({ components: { BlurredOverlay } })
 export default class RaidenDialog extends Vue {
-  @Prop({ required: true, default: false })
+  @Prop({ required: true, default: false, type: Boolean })
   visible!: boolean;
-  @Prop({ required: false })
+  @Prop({ required: false, default: false, type: Boolean })
```
nephix: Why were these type defs needed again? For eslint to find missing required props?

nephix:
Looks slick, thank you

raiden-dapp/src/components/RaidenDialog.vue
kelsos: So the default is there because otherwise, the property will be undefined which makes vue complain and might or might not create issues with reactivity. Adding type especially for booleans allows you to use the existence/non-existence of the attribute as boolean. 

So instead of having to do `:hide-close="true"` you can just use `hide-close` and this will be interpreted the same as the first one. If you don't have the type there the second option won't work.

PR #1055, opened 2020-02-18
Reports transfer as failed as soon as invalid SecretRequest is detected

### LINKED ISSUE #1054: Report transfer as failed upon invalid SecretRequest
On the unhappy case that the mediators take out more fees than we expected (may happen e.g. in case PFS report a wrong fee estimation, even though we add by default safety margin of 10%), target is going to request the secret with a smaller-than-expected amount, which we don't recognize and don't reply with the usual `SecretReveal`.

In this case, the transfer didn't fail yet (it does so when it expires), and theoretically the target could realize it and send a higher secret request which we could then accept, but we know by implementation this doesn't happen, and we should report as early as possible the transfer as failed when receiving an invalid amount for `SecretRequest`, and let the user know this transfer probably isn't going to succeed, even though it'll only finally fail upon `LockExpiration` (as usual for failed transfers).

### Tasks:
- [x] Persist invalid `SecretRequest` in `SentTransfer`/`state.sent` state
- [x] Detect it and report in `transfers$` with `success: false`
- [x] Identify it in dApp and informs to user on pending transfer dialog as failed (even though still pending, to be completed only after respective `LockExpired`

andrevmatos:
Fix #1054

### DIFF
--- a/raiden-ts/src/transfers/utils.ts
+++ b/raiden-ts/src/transfers/utils.ts
@@ -67,26 +67,30 @@ export function makeMessageId(): UInt<8> {
  */
 export function raidenSentTransfer(sent: SentTransfer): RaidenSentTransfer {
   const [status, changedAt]: [RaidenSentTransferStatus, number] = sent.lockExpiredProcessed
-      ? [RaidenSentTransferStatus.expired, sent.lockExpiredProcessed[0]]
-      : sent.unlockProcessed
-      ? [RaidenSentTransferStatus.unlocked, sent.unlockProcessed[0]]
-      : sent.lockExpired
-      ? [RaidenSentTransferStatus.expiring, sent.lockExpired[0]]
-      : sent.unlock
-      ? [RaidenSentTransferStatus.unlocking, sent.unlock[0]]
-      : sent.secretReveal
-      ? [RaidenSentTransferStatus.revealed, sent.secretReveal[0]]
-      : sent.channelClosed // channelClosed before revealing
-      ? [RaidenSentTransferStatus.closed, sent.channelClosed[0]]
-      : sent.refund
-      ? [RaidenSentTransferStatus.refunded, sent.refund[0]]
-      : sent.transferProcessed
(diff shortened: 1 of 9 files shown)


raiden-ts/src/transfers/utils.ts
```diff
@@ -67,26 +67,30 @@ export function makeMessageId(): UInt<8> {
...
-      ? [RaidenSentTransferStatus.revealed, sent.secretReveal[0]]
-      : sent.channelClosed // channelClosed before revealing
-      ? [RaidenSentTransferStatus.closed, sent.channelClosed[0]]
-      : sent.refund
-      ? [RaidenSentTransferStatus.refunded, sent.refund[0]]
-      : sent.transferProcessed
-      ? [RaidenSentTransferStatus.received, sent.transferProcessed[0]]
-      : [RaidenSentTransferStatus.pending, sent.transfer[0]],
-    success: boolean | undefined =
+    ? [RaidenSentTransferStatus.expired, sent.lockExpiredProcessed[0]]
```
kelsos: I get the feeling that we have a few ternaries too much here. Isn't any alternative approach that is elegant and easier to read?

raiden-ts/src/transfers/utils.ts
```diff
@@ -102,7 +106,7 @@ export function raidenSentTransfer(sent: SentTransfer): RaidenSentTransfer {
     token: sent.transfer[1].token,
```
kelsos: nitpick: If we are not limited by the number of lines in the function, would it make sense `sent.transfer[1]` into a variable instead of repeating?

raiden-ts/src/transfers/utils.ts
andrevmatos: Yes, I agree, but I couldn't think on a better alternative yet, because it's not just conditions, but actually a cascading priority order for them. Maybe a separate function can help, but I'm open to suggestions 😄

andrevmatos: Done =)

PR #1058, opened 2020-02-20
Updates dependencies

kelsos:
Everything should be here, to avoid dependabot spam, and spending a couple of hours approving and rebasing.


raiden-dapp/.eslintrc.js
```diff
@@ -49,7 +49,7 @@ module.exports = {
       }
     ],
     'vue-i18n/no-unused-keys': ['error', {
-      'src': ['./src'],
+      'src': './src',
```
nephix: Nice, did this fix the unused key rule?

PR #1070, opened 2020-02-24
SDK-Development doc

taleldayekh:
Moving the SDK-Development guide to the raiden-ts directory. Almost everything is the same, I started going through the content to phrase it in a more "accessible" way. Note that this was not with the intention to remove anything since the purpose of the document is to provide in-depth technical descriptions of the SDK.

This is an example of the kind of language changes I was striving for:
**Before**
>Raiden client class is the entrypoint for the public API. This class's focus must be Developer Experience (DX), so we should avoid exposing too much of the internals on it, as well as providing anything needed for every reasonable usecase. This balance is important to be kept in mind, and is team's responsibility to decide on whatever non-trivial decision. It's desirable for the interface's changes to be as backwards-compatible as possible, although this isn't guaranteed before MVP release.

**After**
>The `Raiden` client class is the entrypoint for the public API. The focus of this class is Developer Experience and we should avoid exposing to much of its internals while still providing what is needed for every reasonable use case.

>This balance is important to keep in mind. It is desirable that changes in the interface are as backwards compatible as possible, although not necessarily before a MVP release.

As you can see it provides the very same information, but the second version is more digestible in my opinion.

This would be another example:
**Before**
> One of the main development targets is correctness and TypeScript compile-time checks to aid implementation, validation and integration of the diffent parts of the codebase. io-ts is a library which all allows to bridge the gap between TypeScript's compile-time safety and runtime validations. Basically, the problem it solves is: you can tell TypeScript that a function argument, return or a variable is of a given type, and it'll help you not do wrong things with this information (e.g. passing a number to a function which expects a string) while writting your code; but when you're dealing with unknown data (e.g. a handled state file, a incoming message, etc), you can't be sure it matches your expectations, and if not, very bad things can happen.

**After**
> TypeScript helps us check for correctness and aid implementation, validation and integration of various parts of the codebase. We can tell TypeScript what type of argument, return value or variable is expected in a function which helps us avoid passing wrong types when writing our code, like passing a number to a function that expects a string.

>However, when we are dealing with unknown data we cannot always be sure it matches our expectations. To bridge this gap we use the `io-ts` library.

To go through such a dense document as the DSK Development docs is quiet time consuming as I noticed. So for now, this PR moves the document from the Wiki to the repo and through future iterations, we could rework it completely if there is an interest in doing so.

### DIFF
--- a/raiden-ts/README.md
+++ b/raiden-ts/README.md
@@ -58,7 +58,7 @@ External off-chain communication with the Raiden Network is provided by a dedica
                      +-------------+
 ```
 
-A technical deep dive into the SDK architecture, technologies, tips and details on the design goals and decisions can be found in the [project's Wiki page](https://github.com/raiden-network/light-client/wiki/SDK-Development). Reading it is highly recommended to anyone wishing to better understand how the Raiden Light Client works under the hood or to contribute to it, though not required to use this library as a dApp developer.
+A technical deep dive into the SDK architecture, technologies, tips and details on the design goals and decisions can be found in the [SDK Development](https://github.com/raiden-network/light-client/blob/sdk-dev-doc/raiden-ts/SDK-Development.md) document. Reading it is highly recommended to anyone wishing to better understand how the Raiden Light Client works under the hood or to contribute to it, though not required to use this library as a dApp developer.
 
 ## Getting Started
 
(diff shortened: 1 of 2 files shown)


taleldayekh:
> @taleldayekh Simplifying language for the reader always makes sense. Thanks for pointing this out :)
> 
> Could you create a separate issue to improve readability of the docs?

Sure thing, I'll do that.

Issue #693, opened 2019-12-06
I want to understand the different PFS error messages


christianbrb:
## User Story
As Chris I want to understand the different PFS error messages so that I can derive my next actions and feel confident using the Raiden Network.

## Acceptance criteria
Error messages make more sense for the pfs that we choose.

## Additional information
https://github.com/raiden-network/raiden-services/blob/master/src/pathfinding_service/exceptions.py

This issue depends on #1000

christianbrb:
As a first simple step:

| Problem  | Solution | Error Codes|
|:-|:-|:-|
| No route found  | Please try the request again or select a different PFS  | HTTP 404 (Error Code 22xx) |
| PFS Request failed | Please select a different PFS | Error Codes 20xx |
| PFS Payment failed  | Deposit more tokens to the UDC or select a different PFS  | Error Codes 21xx  |

As a hover over or small/hidden we could add the full PFS error message.

christianbrb:
@nephix If you haven't started this yet, we would like to move it to the backlog in favor of other improvements needed for EthCC

nephix:
@christianbrb I'm almost done with it

PR #1077, opened 2020-02-25
PFS error message handling

nephix:
#693 

This PR adds localization for `Error` and `RaidenError` classes in the dApp.
There is a new component `ErrorMessage` that hooks up `raidenError.code` to our locales in 

> src/locales/en.json 

and renders for every known error code:
* A title
* A problem description
* A possible solution

The solution in the locale file can either be a single message or an Array of messages. If it's an array of messages, the `ErrorMessage` component will render the solution as an ordered list. That way, are able to guide users in case of errors with precise steps.


raiden-dapp/src/locales/en.json
```diff
@@ -331,5 +331,73 @@
...
+    "PFS_LAST_IOU_REQUEST_FAILED": {
+      "title": "Error while depositing tokens for Pathfinding Service",
+      "problem": "While trying to deposit tokens for route requests, the Pathfinding Service returned an error.",
+      "solution": "This should happen very rarely, please retry to deposit. If the problem persists, please file an issue on https://github.com/raiden-network/light-client."
+    },
+    "PFS_IOU_SIGNATURE_MISMATCH": {
+      "title": "Error while depositing tokens for Pathfinding Service",
+      "problem": "While trying to deposit tokens for route requests, the Pathfinding Service encountered mismatching signatures.",
+      "solution": "This should happen very rarely, please retry to deposit. If the problem persists, please file an issue on https://github.com/raiden-network/light-client."
+    }
```
nephix: Feel free to provide committable suggestions if some text aren't hitting the nail on the head

raiden-dapp/src/locales/en.json
```diff
@@ -331,5 +331,73 @@
       "rtt": "Ping",
       "price": "Price"
     }
+  },
+  "error-message": {
+    "problem": "Problem:",
+    "solution": "Solution:"
+  },
+  "errors": {
+    "RDN_GENERAL_ERROR": {
```
nephix: That's the error code for "unknown" errors, e.g. if an `Error` object gets thrown instead of a `RaidenError`

Issue #1057, opened 2020-02-20
Blurred dApp parts when opening a modal


nephix:
Currently, the blurred modal backdrop is inconsistently implemented across the dApp, mainly due absolute positioning it in the first parent with `position: relative;`

It should be discussed how to iron this out

## Steps to Reproduce
1. Go to https://lightclient.raiden.network
2. Connect
3. Click on "Deposit"
4. Check other modals as well, e.g. Minting dialog

## Expected Result
- Backdrop looks consistent

## Actual Result
![Screenshot 2020-02-20 at 08 02 28](https://user-images.githubusercontent.com/3169205/74909664-15eb0380-53b9-11ea-872d-a72e6ae9ddea.png)
<img width="1758" alt="Screenshot 2020-02-19 at 17 43 05" src="https://user-images.githubusercontent.com/3169205/74909682-1d121180-53b9-11ea-9755-44b7cc5901a6.png">

kelsos:
@nephix what I got from our discussion with @taleldayekh is that we want the blur to cover the card area.

nephix:
Decision from planning meeting: Cover the whole screen, as it's the simplest solution

taleldayekh:
> Decision from planning meeting: Cover the whole screen, as it's the simplest solution

Genau

PR #1079, opened 2020-02-25
Made backdrop of RaidenDialog fullscreen

nephix:
#1057


nephix:
Didn't do changelog entry because not worth mentioning IMO

raiden-dapp/src/components/RaidenDialog.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <blurred-overlay :show="visible">
+  <blurred-overlay :show="visible" :fullscreen="true">
```
kelsos: Since this is not attached to a variable you can do `fullscreen` instead of `:fullscreen="true"` 

That is the reason why the property has `type: Boolean` in the blurred-overlay component

PR #1086, opened 2020-02-26
Adds derived subkey support to dApp

taleldayekh:
Screenshot 2020-02-26 at 16 44 57
Screenshot 2020-02-26 at 16 45 07


raiden-dapp/src/App.vue
```diff
@@ -57,10 +57,10 @@ export default class App extends Vue {
     );
   }
 
-  async connect() {
-    this.connecting = true;
+  async connect(subkey?: true) {
+    this.connecting = true && !subkey;
```
andrevmatos: Nitpick: It still needs to Connect to Metamask (if it's not) to prompt signing the subkey message, so I'd say you could still have the loading indicator even with subkey on.

taleldayekh: Ok, but I have to add a new loading indicator since that one displays a spinner inside the button even when the subkey text link is pressed. But no problem, I'll look into adding a second spinner.

taleldayekh:
Now there is a spinner also when connecting with subkey and I made the disclaimer text always visible even when the "No Web3 provider..." error appears.

Screenshot 2020-02-27 at 10 29 11
Screenshot 2020-02-27 at 10 25 31

andrevmatos:
Seems very nice now! But, sorry if I'm nitpicking, could you disable the other button once one of them is loading/connecting/clicked? Not sure what would happen if a <smart> user clicked the other while one was loading, but I think we shouldn't discover in a bug report saying dApp ate the user's cat.

taleldayekh:
> Seems very nice now! But, sorry if I'm nitpicking, could you disable the other button once one of them is loading/connecting/clicked? Not sure what would happen if a user clicked the other while one was loading, but I think we shouldn't discover in a bug report saying dApp ate the user's cat.

Sure thing, I also noticed that the CI build failed, I have some type issues that I also need to look over.

Issue #1067, opened 2020-02-21
Summary screen displays direct transfer & service fee


nephix:
## Steps to Reproduce
1. Go to https://lightclient.raiden.network
2. Connect
3. Enter address and amount for mediated transfer
4. Select PFS
5. Pay for request
6. Select a route, if appropriate
7. Land on summary screen
8. Hit back button (top left)
9. Enter an address that you can do a direct transfer to
10. Enter an amount
11. Continue
12. land on summary screen

## Expected Result
- Summary displays direct transfer
- No properties of the mediated transfer before are shown

## Actual Result
- Previously transferred service fee is shown:

![Screenshot 2020-02-21 at 12 37 42](https://user-images.githubusercontent.com/3169205/75031807-617bdb00-54a7-11ea-9647-196d849d4ad4.png)

PR #1095, opened 2020-02-27
Fix for displayed service fee in summary

nephix:
#1067


raiden-dapp/src/views/TransferSteps.vue
```diff
@@ -525,37 +525,6 @@ export default class TransferSteps extends Mixins(
 <style lang="scss" scoped>
 @import '../scss/colors';
 
-.confirmation-overlay {
```
nephix: Not related, just removed it because it's not used

PR #1113, opened 2020-02-28
Implement log storage across sessions and logs download

### LINKED ISSUE #1112: Implement log storage across sessions for dApp
## Description

We often get bug reports which lack the logs needed to understand what happened, specially actions/events. Sometimes, the relevant logs to understand an edge condition happened in a past session and were lost upon reload.
We must store the logs in the browser storage across sesions in an efficient way, and enable users to download and submit them on bug reports.

## Acceptance criteria
- logs are stored in browser
- logs persist through reloads
- logs can be downloaded
- it must be processed & stored in an efficient, sync way, to minimize performance overhead

## Tasks
- [ ]

andrevmatos:
Fix #1112 
It was implemented using `loglevel` and `idb`, a feature library which wraps IndexedDB, an efficient, modern browser storage API which allows us to store the (log) items independently and asynchronously, without requiring serializing the items ourselves.

### DIFF
--- a/raiden-dapp/src/components/SplashScreen.vue
+++ b/raiden-dapp/src/components/SplashScreen.vue
@@ -123,7 +123,6 @@ export default class Loading extends Vue {
 
   async downloadLogs() {
     const [lastTime, content] = await getLogsFromStore();
-    if (!content) return;
     const filename = `raiden_${new Date(lastTime).toISOString()}.log`;
     const file = new File([content], filename, { type: 'text/plain' });
     const url = URL.createObjectURL(file);
--- a/raiden-dapp/src/components/SplashScreen.vue
+++ b/raiden-dapp/src/components/SplashScreen.vue
@@ -77,6 +77,9 @@
             {{ $t('splash-screen.getting-started.link-name') }}
           </a>
         </i18n>
+        <div class="splash-screen__no-provider text-center font-weight-light">
+          <a @click="downloadLogs">{{ $t('splash-screen.download-logs') }}</a>
+        </div>
       </v-col>
     </v-row>
   </v-container>
@@ -86,6 +89,7 @@
 import { Component, Emit, Prop, Vue } from 'vue-property-decorator';
 import { Web3Provider } from '@/services/web3-provider';
 import { DeniedReason } from '@/model/types';
+import { getLogsFromStore } from '@/utils/logstore';
 import { mapState } from 'vuex';
 import NoAccessMessage from '@/components/NoAccessMessage.vue';
 import ActionButton from '@/components/ActionButton.vue';
@@ -115,6 +119,24 @@ export default class Loading extends Vue {
   @Emit()
   connect(subkey?: true) {
     return subkey;
+  }
+
+  async downloadLogs() {
+    const [lastTime, content] = await getLogsFromStore();
+    if (!content) return;
+    const filename = `raiden_${new Date(lastTime).toISOString()}.log`;
+    const file = new File([content], filename, { type: 'text/plain' });
+    const url = URL.createObjectURL(file);
+    const el = document.createElement('a');
+    el.href = url;
+    el.download = filename;
+    el.style.display = 'none';
+    document.body.appendChild(el);
+    el.click();
+    setTimeout(() => {
+      URL.revokeObjectURL(url);
+      document.body.removeChild(el);
+    }, 0);
   }
 }
 </script>
--- a/raiden-dapp/CHANGELOG.md
+++ b/raiden-dapp/CHANGELOG.md
@@ -6,6 +6,7 @@
 - [#1015] More prominent warning for low UDC token balance
 - [#695] Skipping of transfer steps.
 - [#1071] Mint token feature.
+- [#1112] Implement log storage across sessions
 
 ### Changed
 - [#693] Localized error messages for Pathfinding errors.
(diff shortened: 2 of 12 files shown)


raiden-dapp/src/components/SplashScreen.vue
```diff
@@ -77,6 +77,9 @@
             {{ $t('splash-screen.getting-started.link-name') }}
           </a>
         </i18n>
+        <div class="splash-screen__no-provider text-center font-weight-light">
+          <a @click="downloadLogs">{{ $t('splash-screen.download-logs') }}</a>
+        </div>
```
nephix: Can we also offer a way to download the logs outside inside the dapp, not only from the splash screen? Some options:
* `ErrorDialog`
* Link in footer

raiden-dapp/src/components/SplashScreen.vue
```diff
@@ -116,6 +120,24 @@ export default class Loading extends Vue {
   connect(subkey?: true) {
     return subkey;
   }
+
+  async downloadLogs() {
+    const [lastTime, content] = await getLogsFromStore();
+    if (!content) return;
```
nephix: Probably good idea to notify the user too

andrevmatos: This is just a debugging option, and doesn't happen if the user has logged at least once in the dApp, which I think is reasonable to assume before the user tries to download the logs, I don't think it's worth it to put a dialog only to tell them that. I can instead allow them to download the file with only a single line of text: `The logs were empty`, so they'll get informed.

raiden-dapp/src/components/SplashScreen.vue
andrevmatos: No, I'd prefer not, because then the logs would still be flowing. It's safer to do it when the SDK is stopped, i.e. splash screen, so no concurrent access while the file is being generated.

raiden-dapp/src/components/SplashScreen.vue
nephix: Or only show the link if there are logs

andrevmatos: This was my first option, but would require another function just to access and check if there're items in the IndexedDB. If the logs are empty because they never logged in, I'll allow them to download the empty file, what else would they expect?

raiden-dapp/src/components/SplashScreen.vue
nephix: AFAIK you can safely read stuff from IndexedDB while still being able to write new stuff.

Telling the user in case of errors to go to the connect screen to get logs seems very odd

raiden-dapp/src/components/SplashScreen.vue
nephix: Was just another (and I think cleaner) option but up to you

raiden-dapp/src/components/SplashScreen.vue
andrevmatos: This feature isn't "to get logs in case of errors", if one wants to see it in real time, they can simply open DevTools, the place designed exactly for this kind of thing. It's an info gathering tool for bug reports, which is reasonable to ask do to with the SDK stopped, IMO.
Besides, I'm against adding it in multiple places, it'd just confuse the user when trying to tell them where to go fetch it, and clutch UI, and if it's to be in a single place, it needs to be on splash, so we are not locked out if some state bug prevents the user from connecting on the

nephix: > This feature isn't "to get logs in case of errors"

I guess that is the exact use case for:

> It's an info gathering tool for bug reports

It's pretty common to give users option to collect the info when an error occurs

> it'd just confuse the user
> and clutch UI

I doubt that, in the end it's just a link

andrevmatos: I politely disagree with you. Please, if there's nothing technically wrong with the PR, approve it and open a feature **discussion** issue requesting and justifying buttons in multiple places to access this debugging tool. The PR fulfills the acceptance criteria, and I was not informed when first requested about this additional requirement.

andrevmatos:
@nephix There's no follow up issues from my stand point or from the initial requirements.
Please, feel free to open them if you have a different opinion on that and would like to call it for discussion.

nephix:
These issues result from your PR, not from the requirements of the initial ticket. Therefore, please create them

andrevmatos:
What issues are you referring to? I can't see an issue resulting from my PR beyond your *wish* to have done a part of it differently, which couldn't be implied from the original issue nor requirements, nor is a problem with the PR itself but a matter of taste, which I, as the implementer, disagree, and therefore needs discussion with the whole team to decide if or when we're going to change what was done. Therefore, I won't create issues I don't see as existent. Feel free to create them if *you* see.

nephix:
The original ticket was created by you, as well as the acceptance criteria and we already discussed the logging feature and what should be done. Looks like you only picked acceptance criteria that suit you?

But your choice 🤷‍♂

andrevmatos:
I was asked to write the issue, not offered me, and put as acceptance criteria everything I remembered from the discussion, and I definitely don't remember the points raised by you after you saw the PR ready. I'm not saying it won't be done, but instead that, as our perspective about what were the original requirements diverge, and I'm on the _status quo_ side about it, possible changes can be discussed and decided on appropriate issues, but I'm not the one proposing them and therefore am not the responsible for creating them, but completely support you thinking differently, creating the issues, calling for discussion and, if the majority agrees with the proposal, it'll get implemented, maybe even by me, or whoever gets assigned to the newly decided issue.

Issue #296, opened 2019-08-30
Display currently used raiden versions in light client


nephix:
## User Story

As a light client user I want to know the raiden and raiden contract version the light client is currently compatible with.

Related to #226 

## Acceptance criteria

- [ ] Raiden version should be displayed in the UI
- [ ] Raiden contract version should be displayed in the UI

nephix:
I'm hard coding the contract version for now, as it's currently not possible to determine them dynamically (https://github.com/raiden-network/raiden-contracts/issues/1287)

PR #1114, opened 2020-02-28
Display raiden versions in footer

nephix:
#296 

Screenshot 2020-02-28 at 11 16 37


taleldayekh:
That's was a smart solution!

PR #1124, opened 2020-03-03
Fixed missing error message when connecting to main net

### LINKED ISSUE #1115: No error message when pressing connect on an unsupported network
## Steps to Reproduce
1. Switch to mainnet
2. Press connect

## Expected Result
- Warning about network not being supported

## Actual Result
- Button/label change colors and then nothing happens

nephix:
Fix #1115

### DIFF
--- a/raiden-dapp/src/components/SplashScreen.vue
+++ b/raiden-dapp/src/components/SplashScreen.vue
@@ -56,6 +56,13 @@
           </div>
         </v-col>
       </template>
+      <div class="splash-screen__message">
+        <no-access-message
+          v-if="accessDenied"
+          :reason="accessDenied"
+        ></no-access-message>
+      </div>
+
       <v-col v-if="!injectedProvider" cols="8">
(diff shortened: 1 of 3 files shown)


raiden-dapp/src/components/SplashScreen.vue
```diff
@@ -188,12 +195,6 @@ export default class Loading extends Vue {
     height: 48px;
   }
 
-  &__disclaimer {
-    margin: 0 auto;
-    margin-top: 180px;
-    max-width: 560px;
```
taleldayekh: Thank you for taking care of this. I'm just wondering why the max-width is removed? I added it to have the text break in those two lines if the window is large.

nephix: The text already broke, but the max width made the text appear uncentered

taleldayekh: Ah, ok, did not notice that. Thank you.

PR #1164, opened 2020-03-06
Upgrade matrix to 5.1.0

andrevmatos:
And get rid of custom type definitions for `matrix-js-sdk`, favoring the `@types/` one.
Some adaptation and better error handling were needed too, for matrix rooms operations and maintainance.
Includes rewrite of `package-lock.ts`s.


raiden-ts/tests/e2e/provider.ts
```diff
@@ -30,7 +30,7 @@ export class TestProvider extends Web3Provider {
           seed: 'testrpc_provider',
           network_id: 1338,
           db: memdown(),
-          logger: console,
+          // logger: console,
```
kelsos: Why comment instead of removing?

andrevmatos: I uncomment eventually to debug e2e tests, and some time seems like I forgot to comment out again 😅

kelsos: Can we make it activatable by some environment variable or something? :P

andrevmatos: Haha, you guys hate logs, this is more for local testing & test debugging in some very specific cases, but if someone feels the need, we can improve it at some point ;)

kelsos: Well not hate. I like them when there is an actual issue. But they come with a performance overhead and they can be overwhelming when they are too verbose. 

In the context ofthis comment though, I would say not as much as I hate commented code :P 

The environment variable here was an alternative to commenting and uncommenting the logger code for the tests.

Issue #744, opened 2019-12-13
Investigate and reduce bundle size


kelsos:
## Description

We have a problem regarding the 3rd party (vendor) dependency chuck size.

```
  File                                      Size             Gzipped

  dist/js/chunk-vendors.36b61744.js         14906.35 KiB     3860.37 KiB
```

The generated file is around 15MiB unziped and around ~4MiB gzipped.

![image](https://user-images.githubusercontent.com/2269732/70795911-41da7f00-1da1-11ea-8b6b-d9675ee7c7b5.png)

After generating a report `npm run build -- --report` you can see that more than half of this is the `node_modules` of `raiden-ts` which means that instead of using the build version of the linked library.

We need to find out a way to reduce the bundle's size.

## Acceptance criteria
- chunk-vendors size is more reasonable.

## Tasks
- [ ] Start with upgrading and continue investigating later

nephix:
The local dependency in `raiden-dapp/package.json` seems not to be the problem. I changed
```
-    "raiden-ts": "file:../raiden-ts",
+    "raiden-ts": "^0.2.2",
```
and ran `npm run build` and the bundle size pretty much stayed the same.

I've investigated a bit about the individual packages coming from `raiden-ts`:
```
$ npx howfat -r table raiden-ts

npx: installed 18 in 2.161s
raiden-ts@0.2.2 (102 deps, 31.36mb, 10097 files)
╭────────────────────────┬──────────────┬──────────┬───────╮
│ Name                   │ Dependencies │     Size │ Files │
├────────────────────────┼──────────────┼──────────┼───────┤
│ matrix-js-sdk@2.4.6    │           61 │  12.81mb │  2740 │
├────────────────────────┼──────────────┼──────────┼───────┤
│ ethers@4.0.40          │           16 │   5.59mb │   282 │
├────────────────────────┼──────────────┼──────────┼───────┤
│ rxjs@6.5.3             │            1 │   4.88mb │  3616 │
├────────────────────────┼──────────────┼──────────┼───────┤
│ @types/lodash@4.14.149 │            0 │   1.77mb │  1410 │
├────────────────────────┼──────────────┼──────────┼───────┤
│ lodash@4.17.15         │            0 │   1.34mb │  1049 │
├────────────────────────┼──────────────┼──────────┼───────┤
│ fp-ts@2.3.0            │            0 │   1.28mb │   372 │
├────────────────────────┼──────────────┼──────────┼───────┤
[...]

PR #1165, opened 2020-03-07
Some improvements to bundle size

andrevmatos:
- tree-shaking lodash & ethers import syntax in SDK
- common modules aliasing

Part of #744 , reduces ~1.5MB on my bundle, not much, but on the right direction.


andrevmatos:
There's some in the PR's description, but more detailed: from previous report from thomas of ~13.1MB, after this PR, if felt to ~11.5MB, 3.484MB -> 3.247MB Gzipped.

nephix:
Good job on the reduction @andrevmatos 👍

Issue #348, opened 2019-09-26
Integration tests with the Raiden client and services


christianbrb:
## Description
We have got a very good test coverage at the moment, but the Raiden client could break compatibility at any time. As a retro action item we wanted to discuss a way how to do integration tests.

**Requires a lot of setup. Perhaps save it for a later time since we are splitting from the Raiden develop branch at the moment.**

## Acceptance criteria
- Initially for the SDK
- That we are compatible with the Python client

## Tasks
- [ ] Lool into Python client test code
- [ ] Get cli merged to master (error handling needs some work, features need to be exposed)
- [ ] Check, if the core / python test set-up could be used (Private chain, matrix, raiden node)

christianbrb:
Discussion in the Refinement
- Basic set-up for an automated integration test
- More like the scenario player
- With a UI framework

christianbrb:
We could use the script, that @andrevmatos created for the #551 (not merged to master yet).
We could use derived subkey or another ETH provider (injected key) -> see raiden-installer POC (maybe using it).

Could we use ganache? -> Would be easier?

Not using the scenario player right now.

andrevmatos:
@christianbrb The script at `cli` branch, being a node.js implementation, already uses ethers' `Wallet`, the same as a derived subkey, so it isn't needed, and can be used directly.
I don't think ganache is needed, as integration tests already setup a private geth network, which can be used directly by both light and full client, as well as the test `synapse` instance.

kelsos:
Closing since the PR has been merged, any work related to the docker image will be done in the followup PRs

PR #1167, opened 2020-03-09
Preparing private chain Dockerfile

kelsos:
Infrustructure setup for #348


kelsos:
Integration tests seem to work but jest fails to complete.

```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```
Running with  `--detectOpenHandles` does nothing, the tests seem to pass properly when the flag is active and nothing gets reported.

Maybe for CI we could use the `--forceExit` flag to terminate jest after running the tests.

kelsos:
@nephix Can you have a look tomorrow morning to verify that this works for you on OSX. And if not we can coordinate to ensure that you can also run the tests locally.

nephix:
@kelsos sure, anything I should know/do to get it running?

kelsos:
@nephix have a look at the readme and let me know if anything is missing

nephix:
Works perfectly fine on macOS, great job! 👍 
I also ran into the 
> Jest did not exit one second after the test run has completed.
> This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.

kelsos:
@nephix that is awesome. I was not sure if it would run out of the box.

nephix:
I'm unassigning myself because I don't feel qualified to give feedback on the python implementation or Dockerfile setup

kelsos:
@nephix The important thing is that you confirmed that it works on OSX without issues :)

nephix:
I saw in the test logs that we don't use confirmation blocks, do we? Because I received `channel/open/success` twice with `confirmed: undefined` and `confirmed: true` within the same transaction block

kelsos:
To be honest I don't really remember how confirmation blocks work in this case. Do the promise calls `openChannel` return with the unconfirmed. In that case, do we need to wait for confirmation blocks in the test itself? Or do I just specify the confirmation blocks on the constructor and then the promise waits for them?

integration/Dockerfile
```diff
@@ -0,0 +1,133 @@
...
+COPY geth/genesis.py /usr/local/bin
+
+ARG LOCAL_BASE=/usr/local
+RUN download_geth.sh
+
+COPY geth/deploy_contracts.py /usr/local/bin
+COPY geth/deploy.sh /usr/local/bin
+
+ARG DATA_DIR=/opt/chain
+RUN deploy.sh
```
andrevmatos: Consecutive RUNs can/should be concatenated together with `&& \` in order to reduce the number of intermediary images: docker keeps each command on a Dockerfile as a separate image on disk (although deduplicated with the others), and this can save up several GBs on large Dockerfiles.
Also, any in-disk files downloaded/created on intermediary images are kept because of this, that's why you usually see `apt-get update && ... && rm -rf /var/...` RUN commands, so these apt-get files/cache/dbs aren't kept on these intermediary images

kelsos: I can try :) the whole final image at this point is around 1.6GiB though (after removing ethash)

kelsos: Any tips you have on how to improve though are really welcome.

andrevmatos: Yes, this shouldn't help much with the final image size, but can at least save a lot in the intermediary ones. My main tip is, try to group all similar commands (as much as possible, of course, as some may depend on others) and check if there're caches left behind that could/should be removed (e.g. pip caches, npm caches, apt caches, etc)

andrevmatos:
> To be honest I don't really remember how confirmation blocks work in this case. Do the promise calls `openChannel` return with the unconfirmed. In that case, do we need to wait for confirmation blocks in the test itself? Or do I just specify the confirmation blocks on the constructor and then the promise waits for them?

The `txBlock` is the effective block of the mined transaction, and shouldn't change between unconfirmed & confirmed actions/events (unless a reorg occurs). Confirmation is checked on `newBlock`s, so I think it's a requirement that a new block goes through for the confirmation to happen, but `config.confirmationBlocks` could be 1 or 0 and it'd still confirm on the next block flowing through.

Anyway, if using the public methods in `Raiden` class, the relevant promises wait for confirmation automatically, so you don't need to do anything, just `await` it. This is done by the 4th param of `asyncActionToPromise` utility on methods which depends on the state machine, or the `waitConfirmation` helper on methods which doesn't need the state machine.

integration/README.md
```diff
@@ -0,0 +1,48 @@
...
+```bash
+docker build -t lightclient-integration .
+```  
+  
+### Running
+
+To start the container you can run:
+
+```bash
+docker run --name lc-integration -p 80:80 -p 6000:6000 -p 5001:5001 -p 5002:5002 -p 8545:8545 lightclient-integration
```
andrevmatos: May I recommend prepending `localhost:` to the ports mapping, so they aren't exposed outside if the user doesn't have a proper firewall configured?

raiden-ts/tests/integration/integration.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+  const deploymentServiceInfoFile = process.env.DEPLOYMENT_SERVICES_INFO;
+
+  assert(deploymentInfoFile !== undefined);
+  assert(deploymentServiceInfoFile !== undefined);
+  assert(fs.existsSync(deploymentInfoFile));
+  assert(fs.existsSync(deploymentServiceInfoFile));
+
+  const options = { encoding: 'utf8', flag: 'r' };
+
+  const deployFile = await fs.readFileSync(deploymentInfoFile, options);
```
andrevmatos: `*Sync` methods don't need `await`, they're synchronous, but are also blocking. Prefer [fs.promises](https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_promises_api) APIs (async)

raiden-ts/tests/integration/integration.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+  const contractsInfo = ({
+    ...deploy.contracts,
+    ...servicesDeploy.contracts,
+  } as unknown) as ContractsInfo;
+
+  return await Raiden.create('http://localhost:8545', account, stateOrStorage, contractsInfo, {
+    pfs: 'http://localhost:6000',
+    matrixServer: 'http://localhost',
+    revealTimeout: 10,
+    settleTimeout: 50,
```
andrevmatos: `confirmationBlocks: 1` is advised, default of 5 can conflict with small revealTimeout of 10

raiden-ts/tests/integration/integration.spec.ts
kelsos: I was trying to find the promises api but I could only find the callback one and I probably missed the await there, thanks for noticing :)

integration/Dockerfile
kelsos: We should not have any npm caches in the image but I am looking into the other :) thanks

kelsos: But I can see that the install I run is actually responsible for a great part of the size. That steps add around 500MB. I am trying to figure out if somehow unneeded dependencies are installed.

PR #1186, opened 2020-03-17
Added transfer identifier

### LINKED ISSUE #921: I want to add a payment identifier to my transfer
## User Story
As Chris, I want to add a payment identifier to my transfer so that the target can identify the payment.

## Acceptance criteria
- Included identifier with each transfer if it is in the parameters

### LINKED ISSUE #1187: Use current timestamp as identifier
## Description

Currently, the SDK uses `randomBytes` from ethers/crypto when the `paymentId` option of a transfer is `undefined`. As a follow up to https://github.com/raiden-network/light-client/pull/1186#discussion_r393660879 this should be changed to the current timestamp in ms

## Acceptance criteria
- 

## Tasks
- [ ]

nephix:
Resolves #921 
Resolves #1187 

Untitled

### DIFF
--- a/raiden-dapp/src/views/TransferSteps.vue
+++ b/raiden-dapp/src/views/TransferSteps.vue
@@ -201,7 +201,7 @@
 <script lang="ts">
 import { Component, Mixins } from 'vue-property-decorator';
 import { RaidenPFS, RaidenError } from 'raiden-ts';
-import { BigNumber, bigNumberify } from 'ethers/utils';
+import { BigNumber, bigNumberify, randomBytes } from 'ethers/utils';
 
 import { BalanceUtils } from '@/utils/balance-utils';
 import { Token, Route, Transfer } from '@/model/types';
@@ -250,6 +250,7 @@ export default class TransferSteps extends Mixins(
   pfsFeesPaid: boolean = false;
   pfsSelectionSkipped: boolean = false;
   routeSelectionSkipped: boolean = false;
+  paymentId: BigNumber = bigNumberify(randomBytes(8));
   freePfs: boolean = false;
   showMintDeposit: boolean = false;
   mediationFeesConfirmed: boolean = false;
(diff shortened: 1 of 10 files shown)


raiden-dapp/src/views/TransferSteps.vue
```diff
@@ -250,6 +250,7 @@ export default class TransferSteps extends Mixins(
   pfsFeesPaid: boolean = false;
   pfsSelectionSkipped: boolean = false;
   routeSelectionSkipped: boolean = false;
+  paymentId: BigNumber = bigNumberify(randomBytes(8));
```
kelsos: nitpick: On Raiden-py if this is not defined the epoch in ms during the send is used instead. It is not required but maybe we could follow the same scheme. Is that ok?

nephix: Shouldn't this be done in the SDK? I just chose the same method that was used there in case the `paymentId` is `undefined`

andrevmatos: I'd suggest leaving it as `undefined`, if it isn't provided. The SDK will create it then.

nephix: I think @kelsos meant that the SDK should send the current timestamp if it's `undefined` (instead of random bytes currently)

andrevmatos: Yes, and that's what the SDK does, but only if it isn't handled an explicit `paymentId` (i.e. leaving it as undefined if isn't explicitly set)

nephix: Have created #1187 for this

andrevmatos: Oh, right, the paymentId is random, the messageId is timestamp-based. I'd say this isn't critical, but there's a utility on the SDK which can be used on these cases, `makePaymentId`

nephix:
I have included #1187 here with the hope that circleci updates the build checks

nephix:
I'm merging this even though the build checks haven't completed on GitHub yet, but they did on CircleCI: https://circleci.com/workflow-run/90755261-fdf4-45eb-8bb0-c01d4bc7c499

Issue #1209, opened 2020-03-20
Receiving transfers in the SDK


christianbrb:
## Description
Follow up to #1132 

Finalize receiving transfers in the SDK including
- [x] Single PR for receiving
- [x] Include CLI optimizations
- [ ] Tests
- [ ] Knowledge sharing
    - [ ] Code Review (explanation by André)
    - [x] Testing and reviewing the code (Thomas/Kelsos)
    - [ ] Going through the [Raiden Docs / protocol](https://raiden-network-specification.readthedocs.io/en/latest/mediated_transfer.html#mediated-transfer-happy-path-scenario
) together looking at the pictures (maybe enhanced by André)
- [x] Discussion of the approach with the Raiden Python Team
- [x] Pass CodeClimate
- [ ] Documentation
- [x] Manual testing

## Acceptance criteria
- We have got a feature flag (Disabled = default)
- In the dApp we have got an alert (Callback)
- We can use the CLI to test the receiving transfers
- The whole team understands the code
- The Raiden core team understands what we are doing

## Critical Issues
- https://github.com/raiden-network/raiden-contracts/issues/1370

## Tasks
- [ ]

christianbrb:
Documentation: 
- https://raiden-network-specification.readthedocs.io/en/latest/mediated_transfer.html#mediated-transfer-happy-path-scenario

christianbrb:
Review by the Raiden Python Team (@hackaugusto @Dominik1999)

Notes
- Code Walkthrough by @andrevmatos 
- Q & A:
    - Q: Processed still in the code A: Working with Processed and without
    - Q: Do refunds work? A: Not yet
- Remarks by @hackaugusto 
    -  Considering processing messages to the MS and PFS at once
         - HTTP is expensive
         - Collapsing balance proofs -> only sending the last one
    - Looking at the python code how sync issues have been solved (Race conditions MS/PFS <->client)

Supports:
- Restarts
- Capabilities
- Works without delivered

Not yet implemented:
- Refund (maybe it's going to be removed as it needs a lot of collateral)
- Monitoring Service
- The economic viability of on-chain secret registration
- Reroute
- Signature for RevealSecret
- Remove Delivered
- Mediation (one additional epic)

@Dominik1999 @hackaugusto @andrevmatos Let me know, if this is documented correctly. Thanks :)

CC @GataKamsky

PR #1210, opened 2020-03-21
Transport capabilities

andrevmatos:
Implement transport capabilities, preparation for #1209

### DIFF
--- a/raiden-ts/src/messages/utils.ts
+++ b/raiden-ts/src/messages/utils.ts
@@ -284,7 +284,7 @@ export function getBalanceProofFromEnvelopeMessage(
  */
 export function encodeJsonMessage(message: Message | Signed<Message>): string {
   if ('signature' in message) return losslessStringify(Signed(Message).encode(message));
-  else return losslessStringify(Message.encode(message));
+  return losslessStringify(Message.encode(message));
 }
 
 /**
@@ -304,7 +304,7 @@ export function decodeJsonMessage(text: string): Message | Signed<Message> {
     `Invalid message type: ${parsed?.['type']}`,
   );
   if ('signature' in parsed) return decode(Signed(Message), parsed);
-  else return decode(Message, parsed);
+  return decode(Message, parsed);
 }
 
 /**
--- a/raiden-ts/src/messages/utils.ts
+++ b/raiden-ts/src/messages/utils.ts
@@ -20,7 +20,6 @@ const CMDIDs: { readonly [T in MessageType]: number } = {
   [MessageType.REFUND_TRANSFER]: 8,
   [MessageType.UNLOCK]: 4,
   [MessageType.LOCK_EXPIRED]: 13,
-  [MessageType.TO_DEVICE]: 14,
   [MessageType.WITHDRAW_REQUEST]: 15,
   [MessageType.WITHDRAW_CONFIRMATION]: 16,
   [MessageType.WITHDRAW_EXPIRED]: 17,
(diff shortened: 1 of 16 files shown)


raiden-ts/src/transport/epics.ts
```diff
@@ -420,18 +451,14 @@ function fetchSortedMatrixServers$(matrixServerLookup: string, httpTimeout: numb
...
   server: string,
   setup: RaidenMatrixSetup | undefined,
-  {
-    address,
-    signer,
-  }: {
-    address: RaidenEpicDeps['address'];
-    signer: RaidenEpicDeps['signer'];
-  },
+  { address, signer }: Pick<RaidenEpicDeps, 'address' | 'signer'>,
```
kelsos: What is the `Pick` type?

andrevmatos: It's a typescript type utility which selects the given keys from the interface passed as parameter; in this case, it'll be like `{ address: RaidenEpicDeps['address']; signer: RaidenEpicDeps['signer'] }`. I use it to avoid requiring the whole `RaidenEpicDeps` object on local functions/utils, so I can pass this destructured subset

raiden-ts/src/transport/epics.ts
```diff
@@ -650,88 +696,98 @@ export const matrixPresenceUpdateEpic = (
...
+              { address: recovered },
+            );
+          }),
+          catchError(err => (log.debug('Error validating presence event, ignoring', err), EMPTY)),
+        );
+      }),
+    )
+    .pipe(
+      withLatestFrom(latest$),
+      // filter out if presence update is to offline, and address became online in another user
```
nephix: Nice edge case, I would have likely forgotten about it =)

raiden-ts/src/messages/utils.ts
```diff
@@ -311,7 +303,8 @@ export function decodeJsonMessage(text: string): Signed<Message> {
       Object.values(MessageType).some(t => t === parsed['type']),
     `Invalid message type: ${parsed?.['type']}`,
   );
-  return decode(Signed(Message), parsed);
+  if ('signature' in parsed) return decode(Signed(Message), parsed);
+  else return decode(Message, parsed);
```
nephix: Same

raiden-ts/src/messages/utils.ts
```diff
@@ -286,13 +277,14 @@ export function getBalanceProofFromEnvelopeMessage(
...
+ * Uses lossless-json to encode BigNumbers as JSON 'string' type, as Raiden
  *
  * @param message - Message object to be serialized
  * @returns JSON string
  */
-export function encodeJsonMessage(message: Signed<Message>): string {
-  return losslessStringify(Signed(Message).encode(message));
+export function encodeJsonMessage(message: Message | Signed<Message>): string {
+  if ('signature' in message) return losslessStringify(Signed(Message).encode(message));
+  else return losslessStringify(Message.encode(message));
```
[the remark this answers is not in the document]
kelsos: I certainly like the idea of no-else-return :smile:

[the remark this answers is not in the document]
andrevmatos: I fixed it here, but the eslint rule could go in a separate PR, as it will probably touch code outside of the scope of this PR. Thanks

raiden-ts/src/transport/epics.ts
andrevmatos: Thanks. Yes, shouldn't be too serious on our test environment, as we have a single server, but on  federated environments, quickly switching servers could give users unexpected bugs of appearing offline.

PR #1221, opened 2020-03-24
Receive payments

### LINKED ISSUE #1209: Receiving transfers in the SDK

andrevmatos:
Fix #1209

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+### Added
+- [#1209] Support for receiving payments
+
+[#1209]: https://github.com/raiden-network/light-client/issues/1209
+
 ## [0.5.0] - 2020-03-27
 
 ### Added
@@ -128,6 +134,8 @@
 - Add protocol message implementation.
 
 
+[Unreleased]: https://github.com/raiden-network/light-client/compare/v0.5.0...HEAD
+[0.5.0]: https://github.com/raiden-network/light-client/compare/v0.4.2...v0.5.0
 [0.4.2]: https://github.com/raiden-network/light-client/compare/v0.4.1...v0.4.2
 [0.4.1]: https://github.com/raiden-network/light-client/compare/v0.4.0...v0.4.1
 [0.4.0]: https://github.com/raiden-network/light-client/compare/v0.3.0...v0.4.0
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -62,6 +62,26 @@ import {
 } from './actions';
 
 /**
+ * Receives an async function and returns an observable which will retry it every interval until it
+ * resolves, or throw if it can't succeed after 10 retries.
+ * It is needed e.g. on provider methods which perform RPC requests directly, as they can fail
+ * temporarily due to network errors, so they need to be retried for a while.
+ * JsonRpcProvider._doPoll also catches, suppresses & retry
+ *
+ * @param func - An async function (e.g. a Promise factory, like a defer callback)
+ * @param interval - Interval to retry in case of rejection
+ * @param retries - Max number of times to retry
+ * @returns Observable version of async function, with retries
+ */
+function retryAsync$<T>(func: () => Promise<T>, interval = 1e3, retries = 10): Observable<T> {
+  return defer(func).pipe(
+    retryWhen(err$ =>
+      err$.pipe(mergeMap((err, i) => (i < retries ? timer(interval) : throwError(err)))),
+    ),
+  );
+}
+
+/**
  * Fetch current blockNumber, register for new block events and emit newBlock actions
  *
  * @param action$ - Observable of RaidenActions
@@ -74,7 +94,7 @@ export const initNewBlockEpic = (
   {}: Observable<RaidenState>,
   { provider }: RaidenEpicDeps,
 ): Observable<newBlock> =>
-  from(provider.getBlockNumber()).pipe(
+  retryAsync$(() => provider.getBlockNumber(), provider.pollingInterval).pipe(
     mergeMap(blockNumber => merge(of(blockNumber), fromEthersEvent<number>(provider, 'block'))),
     map(blockNumber => newBlock({ blockNumber })),
   );
@@ -107,14 +127,16 @@ export const initTokensRegistryEpic = (
         );
       // else, do an initial registry scan, from deploy to now
       else
-        return defer(() =>
-          provider.getLogs({
-            ...registryContract.filters.TokenNetworkCreated(null, null),
-            fromBlock: contractsInfo.TokenNetworkRegistry.block_number,
-            toBlock: 'latest',
-          }),
+        return retryAsync$(
+          () =>
+            provider.getLogs({
+              ...registryContract.filters.TokenNetworkCreated(null, null),
+              fromBlock: contractsInfo.TokenNetworkRegistry.block_number,
+              toBlock: 'latest',
+            }),
+          provider.pollingInterval,
         ).pipe(
-          mergeMap(from),
+          mergeMap(logs => from(logs)),
           map(log => ({ log, parsed: registryContract.interface.parseLog(log) })),
           filter(({ parsed }) => !!parsed.values?.token_network_address),
           // for each TokenNetwork found, scan for channels with us
@@ -123,21 +145,26 @@ export const initTokensRegistryEpic = (
               concat$(
                 // concat channels opened by us and to us separately
                 // take(1) won't subscribe the later if something is found on former
-                defer(() =>
-                  provider.getLogs({
-                    address: parsed.values.token_network_address,
-                    topics: [null, null, encodedAddress] as string[], // channels from us
-                    fromBlock: log.blockNumber ?? contractsInfo.TokenNetworkRegistry.block_number,
-                    toBlock: 'latest',
-                  }),
+                retryAsync$(
+                  () =>
+                    provider.getLogs({
+                      address: parsed.values.token_network_address,
+                      topics: [null, null, encodedAddress] as string[], // channels from us
+                      fromBlock:
+                        log.blockNumber ?? contractsInfo.TokenNetworkRegistry.block_number,
+                      toBlock: 'latest',
+                    }),
+                  provider.pollingInterval,
                 ).pipe(mergeMap(from)),
-                defer(() =>
-                  provider.getLogs({
-                    address: parsed.values.token_network_address,
-                    topics: [null, null, null, encodedAddress] as string[], // channels to us
-                    fromBlock: log.blockNumber ?? contractsInfo.TokenNetworkRegistry.block_number,
-                    toBlock: 'latest',
-                  }),
+                retryAsync$(
+                  () =>
+                    provider.getLogs({
+                      address: parsed.values.token_network_address,
+                      topics: [null, null, null, encodedAddress] as string[], // channels to us
+                      fromBlock: log.blockNumber!,
+                      toBlock: 'latest',
+                    }),
+                  provider.pollingInterval,
                 ).pipe(mergeMap(from)),
               ).pipe(
                 // if found at least one, register this TokenNetwork as of interest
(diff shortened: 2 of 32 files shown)


raiden-ts/src/transfers/epics/locked.ts
```diff
@@ -477,6 +492,304 @@ function makeAndSignWithdrawConfirmation(
...
+        });
+
+      const processed$ = defer(() => {
+        const processed: Processed = {
+          type: MessageType.PROCESSED,
+          // eslint-disable-next-line @typescript-eslint/camelcase
+          message_identifier: transfer.message_identifier,
+        };
+        return signMessage(signer, processed, { log });
+      });
```
nephix: Can you please explain what this processed message does? Wasn't there a recent discussion to get rid of them or was this related to something else?

raiden-ts/src/transfers/epics/locked.ts
```diff
@@ -477,6 +492,304 @@ function makeAndSignWithdrawConfirmation(
...
+            expiration: transfer.lock.expiration,
+            message_identifier: makeMessageId(),
+          };
+          return signMessage(signer, request, { log });
+        });
+
+      const processed$ = defer(() => {
+        const processed: Processed = {
+          type: MessageType.PROCESSED,
+          // eslint-disable-next-line @typescript-eslint/camelcase
```
nephix: Comment is probably unnecessary, as linter does not complain about it in https://github.com/raiden-network/light-client/pull/1221/files#diff-da17a2bff4a32e6c562b937e7d182ebfR583 too?

raiden-ts/src/transfers/epics/locked.ts
```diff
@@ -477,6 +492,304 @@ function makeAndSignWithdrawConfirmation(
...
+        };
+        return signMessage(signer, processed, { log });
+      });
+
+      // if any of these signature prompts fail, none of these actions will be emitted
+      return combineLatest([processed$, request$]).pipe(
+        mergeMap(function*([processed, request]) {
+          yield transferSigned({ message: transfer, fee: Zero as Int<32> }, meta);
+          // request initiator's presence, to be able to request secret
+          yield matrixPresence.request(undefined, { address: transfer.initiator });
```
nephix: Do we have to request the presence, even though we may not request the secret?

raiden-ts/src/transfers/epics/secret.ts
```diff
@@ -218,30 +294,154 @@ export const monitorSecretRegistryEpic = (
...
+
+/**
+ * Process newBlocks and pending received transfers. If we know the secret, and transfer doesn't
+ * get unlocked before revealTimeout blocks are left to lock expiration, request to register secret
+ *
+ * @param action$ - Observable of newBlock actions
+ * @param state$ - Observable of RaidenStates
+ * @returns Observable of transferSecretRegister.request actions
+ */
+export const transferAutoRegisterEpic = (
```
nephix: Should we automatically register the secret of a transfer on chain, even if the amount is really low (even lower than gas to be paid)?

raiden-ts/src/transfers/epics/locked.ts
andrevmatos: The discussions in https://github.com/raiden-network/raiden/issues/3254 mostly relate to removing the need of `Delivered` messages.
Although there are ways to minimize the need for `Processed` for `EnvelopeMessage`s (messages carrying a balance proof/channel state change), it was decided they're not reliable enough and still need to be confirmed by an explicit `Processed` for now.

raiden-ts/src/transfers/epics/locked.ts
andrevmatos: Nice catch! You're right, if `NO_RECEIVE`  is truty, we don't need to whitelist/request presence from initiator.

raiden-ts/src/transfers/epics/secret.ts
andrevmatos: There're some TODOs about economic viability of on-chain operations, we definitely must check on this later, but AFAIK, Raiden doesn't check either yet. I'll add a todo here for us to check it as well. Thanks for spotting this.

raiden-ts/src/transfers/epics/locked.ts
```diff
@@ -492,38 +805,49 @@ export const transferGenerateAndSignEnvelopeMessageEpic = (
...
+    ),
+    // merge separatedly, to filter per message type before concat
+    action$.pipe(
+      filter(
+        isMessageReceivedOfType([Signed(LockedTransfer), Signed(Unlock), Signed(LockExpired)]),
+      ),
+    ),
+  ).pipe(
+    concatMap(action =>
+      transfer.request.is(action)
```
kelsos: Just out of curiosity, was there a problem with the switch?

andrevmatos: Yes, unfortunately: as we need to `switch` with `messageReceivedTyped`, and for the same action type (`messageReceived`) switch different functions (receive LockedTransfer, Unlock, LockExpire) under the same lock, and switch can only compare the value of the type, we couldn't use it for a shared action narrowing on the types of a deep property (i.e. `payload.message.type`). An intermediary option, but which would make for heterogenous handling, is to `switch` on the action type `messageReceived`, and nested, `if` or `switch` on the message type; I prefered the former, but let me know if you guys would prefer the later, I don't have strong feelings about this.

raiden-ts/src/transfers/epics/secret.ts
kelsos: In the best case this should be configurable per token. Because it depends on the market values of each token and they might be completely different

raiden-ts/src/channels/epics.ts
```diff
@@ -677,67 +701,64 @@ export const channelDepositEpic = (
...
+                if (!receipt.status)
+                  throw new RaidenError(ErrorCodes.CNL_APPROVE_TRANSACTION_FAILED, {
+                    token,
+                    transactionHash: tx.hash!,
+                  });
+                log.debug(`approve tx "${tx.hash}" successfuly mined!`);
+                return tx.hash;
+              }),
+            ),
+          ),
```
nephix: We do this quite often:
* Wait for transaction
* Throw an error if status is invalid
* Otherwise return transaction hash

Might be good refactor candidate

kelsos: That is a really nice idea.

raiden-ts/src/transfers/epics/locked.ts
kelsos: Ok so we don't only narrow in a single condition but in some case we do on multiple now.

raiden-ts/src/transfers/epics/locked.ts
```diff
@@ -477,6 +492,304 @@ function makeAndSignWithdrawConfirmation(
...
+          state.received[secrethash].transfer[1].message_identifier.eq(msgId)
+        ) {
+          // transferProcessed again will trigger messageSend.request
+          return of(
+            transferProcessed({ message: state.received[secrethash].transferProcessed![1] }, meta),
+          );
+        } else return EMPTY;
+      }
+
+      // full balance proof validation
```
nephix: Neglectable: Maybe sounds like a good candidate for a separate function, since this one here is already quite long? Did a quick search but wasn't able to find other places where we validate the balance proof

andrevmatos: We do in all places handling incoming EnvelopeMessages: receiving LockedTransfer, Unlock, LockExpired & RefundTransfer. The last one we don't handle properly yet, the others have specific rules to validate (specially related to amounts), but maybe some could be extracted, I'll check, thanks.

raiden-ts/src/channels/epics.ts
andrevmatos: Well spotted. I'll try. Thanks.

raiden-ts/src/transfers/epics/locked.ts
andrevmatos: Yes. We could have simple different epics for handling  `messageReceived`, filtering for each message and outputting specific actions for each, but as there's no logic in such epics, I'd say it'd not be worth the hassle. We can go this way in the future if needed, but now, should be enough.

christianbrb: > We could have simple different epics for handling `messageReceived`, filtering for each message and outputting specific actions for each,

Would be good for readability

raiden-ts/src/channels/epics.ts
```diff
@@ -61,6 +66,54 @@ import {
...
+
+/**
+ * Custom operator to wait & assert transaction success
+ *
+ * @param method - method name to use in logs
+ * @param error - ErrorCode to throw if transaction fails
+ * @param deps - object containing logger
+ * @returns operator function to wait for transaction and output hash
+ */
+function assertTx(
```
kelsos: Nice addition :D

andrevmatos: Thanks! It's staged to be moved to channels utils, so I can use it on the SecretRegistry epics on transfers as well! but no logic change

raiden-ts/src/channels/epics.ts
andrevmatos: Done, `assertTx`

raiden-ts/src/channels/utils.ts
```diff
@@ -59,3 +65,31 @@ export function channelAmounts(channel: Channel) {
...
+ * @param error - ErrorCode to throw if transaction fails
+ * @param deps - object containing logger
+ * @returns operator function to wait for transaction and output hash
+ */
+export function assertTx(
+  method: string,
+  error: ErrorCodes,
+  { log }: Pick<RaidenEpicDeps, 'log'>,
+): OperatorFunction<ContractTransaction, Hash> {
+  return tx =>
```
nephix, line 82: I don't get it, where does `tx` come from, some generic voodoo from `OperatorFunction`?

kelsos: @nephix It justs returns a function that takes a transaction as a parameter. Then when you put it in the `pipe()` pipe passes the value to the function. All these operator functions are for use in pipes. Does this help?

raiden-ts/src/transfers/epics/secret.ts
nephix: Might be better to create follow up issues instead of comment todos

raiden-ts/src/channels/utils.ts
nephix: Yes thanks a lot 🙏

raiden-ts/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+### Added
+- [#1209] Support for receiving payments
```
christianbrb, line 5:
```suggestion
- [#1209] Support for receiving payments (experimental) - not supported in the dApp
```

christianbrb:
@andrevmatos Nice PR and thanks for all your time explaining this :)
There is documentation missing. As this is an experimental feature it might be good to not document it right now. If you agree, please open an issue to document the receiving functionality. Thanks :)

Issue #1231, opened 2020-03-26
Informative text describing the state in the *Backup state* - Download not yet available


christianbrb:
## Description
Split from #1222 in order to make a release
@taleldayekh: My idea is to have an informative text describing the state in the *Backup state* menu. In this way, a user will get informed about the state and its purpose when exploring the dApp and encountering this menu.
<img width="669" alt="state-description" src="https://user-images.githubusercontent.com/43838780/77545998-89e26500-6eab-11ea-9ed7-fd868f04ed6f.png">

## Acceptance criteria
- Upload button is removed
- Text to inform the user, that upload is coming soon has been added

## Tasks
- [ ]

PR #1244, opened 2020-03-26
Temporarily comments out upload state button

taleldayekh:
Takes care of issue #1231


taleldayekh:
Of course, always manage to forget the changelog, I'll add it.

Issue #1246, opened 2020-03-27
Integration tests don't exit immediately


nephix:
## Steps to Reproduce
1. Start docker
2.
```
$ cd light-client/integration
$ ./run-integration.sh
```

## Expected Result
Jest exits immediately after the tests are done

## Actual Result
Jest keeps running, sometimes for almost a minute

> Test Suites: 1 passed, 1 total
> Tests:       3 passed, 3 total
> Snapshots:   0 total
> Time:        92.321s
> Ran all test suites.
> Jest did not exit one second after the test run has completed.
> 
> This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.

nephix:
Maybe related to https://github.com/matrix-org/matrix-js-sdk/issues/1101

kelsos:
So up to this point, we know that if the `--detectOpenHandles` flag is used, there is no error, no open handles are getting logged and the tests terminate properly. 

I did some investigation but I could not figure out the actual cause of this issue. One of the things tried was monkey patching the request library inside Matrix to call the `abort()` method on the requests by hand. Unfortunately that did not work at all.

After a lot of unsuccessful attempts I found https://github.com/Raynos/leaked-handles which I used to detect any open handles. I could find only two reported file descriptors that kept pointing to `/dev/pts/0` which doesn't make any sense to me. 

I did a search but I could not find more information.

nephix:
I also tried a bunch of stuff, like 
* Fiddling with `abort-controller` because we use `abort-controller/polyfill` which only seems to polyfill IE when bundling (which we don't do in tests)
* Trying to find open handles

I also skipped a couple of integration tests (minting and channel opening) and the waiting time at the end was still there, although it was a lot shorter.

Since the issue seems to be difficult to resolve and the tests work, I'd say lets put this in the back log and revisit it once the waiting time becomes an actual issue

andrevmatos:
So I found the culprit:
As I had suggested, being mostly sure matrix was the issue, I used matrix-js-sdk's `request` function to replace/wrap matrix internal `request` with a function which would log on call (start), callback (result) and abort (shutdown);
I used mostly the `cli`, which upon first `Ctrl+C` calls `stop`, and upon 2nd exits immediately, and did suffer from this issue as well (as it's also a Node app as much as jest), in the form of the first Ctrl+C hanging and not exiting for up to almost 2min.
[...]

PR #1249, opened 2020-03-27
Adds integration workflow to CicleCI config

### LINKED ISSUE #1216: Introduce a CI configuration to run the integration tests
## Description
As part of the #348 we need to add a CI configuration to run our e2e tests.
Since the integration tests are more complex and potentially slower compared the unit tests an existing suite I would suggest to make then run in a cron/nightly basis as an indicator on the status of our compatibility.

## Acceptance criteria
- The integration tests are running automatically
- The developers are getting notified by email

## Tasks
- [ ]

kelsos:
We should first ensure that #1246 is resolved before this is merged otherwise things might not work properly.

Closes #1216

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,11 @@ executors:
     docker:
       - image: *docker_image
 
+  integration-executor:
+    working_directory: ~/src
+    docker:
+      - image: raidennetwork/lightclient-integration:latest
+
 jobs:
   checkout:
     executor: base-executor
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -186,6 +191,27 @@ jobs:
...
+    executor: integration-executor
+    working_directory: *sdk_working_dir
+    steps:
+      - attach_workspace: *attach_options
+      - run:
+          name: Restore dependencies
+          command: npm ci
+      - run:
+          name: Run integration tests
+          command: source /etc/profile.d/smartcontracts.sh && npm run test:integration -- --ci --runInBand --reporters=default --reporters=jest-junit
```
kelsos: sourcing the script here is required because otherwise the `TTT_TOKEN_ADDRESS` will be undefined and the tests will fail

integration/Dockerfile
```diff
@@ -109,4 +109,7 @@ EXPOSE 8545 8546 8547 30303 30303/udp 5001 5002 9001 6000 80 9101 9092 9093
 
 COPY supervisord.conf /etc/supervisor/conf.d/supervisord.conf
 
-CMD ["entrypoint.sh"]
+# In order to preserve the entrypoint for CicleCI https://circleci.com/docs/2.0/custom-images/#adding-an-entrypoint
+LABEL com.circleci.preserve-entrypoint=true
```
kelsos: Without this option the services won't start

kelsos:
@andrevmatos since this doesn't directly block our PRs being a nightly, I would argue that it is better to enabled it to see how it goes.

PR #1251, opened 2020-03-27
Add node 12.6.1 to integration image

### LINKED ISSUE #1250: Add node 12 to integration image
## Description

It seems that I missed adding node in the image. This means that when trying to run the tests inside the docker image this fails because npm is nowhere to be found.

This is only when running the tests inside the image. e.g. Using the image on CircleCI

Using

```bash
#!/bin/bash -eo pipefail
npm ci

/bin/bash: npm: command not found

Exited with code exit status 127
```

## Acceptance criteria
- 

## Tasks
- [ ]

kelsos:
Closes #1250 

Copied the configuration from -> https://github.com/nodejs/docker-node/blob/5351774b970cd32b80fc0c47a8abff5ba155fccc/12/stretch/Dockerfile#L8

### DIFF
--- a/integration/Dockerfile
+++ b/integration/Dockerfile
@@ -41,6 +41,49 @@ RUN apt-get update \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
+
+ENV NODE_VERSION 12.16.1
+# https://github.com/nodejs/docker-node/blob/5351774b970cd32b80fc0c47a8abff5ba155fccc/12/stretch/Dockerfile#L8
+RUN ARCH= && dpkgArch="$(dpkg --print-architecture)" \
+  && case "${dpkgArch##*-}" in \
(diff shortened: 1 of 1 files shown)


nephix:
Has it advantages for CI to have python and node in the same image? Intuitively I thought having two images with one of them having
```
FROM node:12
```
and then using docker compose in CircleCI but no expert as all 😅

kelsos:
@nephix 

> Has it advantages for CI to have python and node in the same image? Intuitively I thought having two images with one of them having
> 
> ```
> FROM node:12
> ```
> 
> and then using docker compose in CircleCI but no expert as all sweat_smile

I am not a Docker expert either. I was looking for a way to expose environment variables (generated during the build of the image) to the top-level image in this case by using docker-compose, however I couldn't find an easy way. That is the main reason I went with a monolithic configuration in a single Dockerfile instead.

PR #1264, opened 2020-03-31
Fixes width of transfer inputs on mobile devices

### LINKED ISSUE #1243: Transfer screen inputs are full width on mobile
## Steps to Reproduce
1. Serve dapp
2. Connect

## Expected Result
- On mobile, the address and amount inputs are `col-10-sm` wide

## Actual Result
- The inputs are full width

kelsos:
Closes #1243

### DIFF
--- a/raiden-dapp/src/views/Transfer.vue
+++ b/raiden-dapp/src/views/Transfer.vue
@@ -63,7 +63,7 @@
       @submit.prevent="navigateToTransferSteps(target, amount)"
     >
       <v-row justify="center" align="center" class="transfer__recipient">
-        <v-col cols="12" sm="10">
+        <v-col cols="10">
           <address-input
             v-model="target"
             :exclude="[token.address, defaultAccount]"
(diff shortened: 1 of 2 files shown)


raiden-dapp/src/views/Transfer.vue
```diff
@@ -63,7 +63,7 @@
       @submit.prevent="navigateToTransferSteps(target, amount)"
     >
       <v-row justify="center" align="center" class="transfer__recipient">
-        <v-col cols="12" sm="10">
```
nephix: I think it worked before with `cols="12" sm="10"` as well, but I accidentally moved the `v-form` next to a row which seems to cause this issue. Looks like vuetify only wants v-rows to have v-row siblings

kelsos: Yes, it could possibly be it. I didn't spend more time looking into it as the change seems to have the required result :)

nephix: Yes no big deal =)

Issue #1261, opened 2020-03-31
Use WebRTC for faster transfers


christianbrb:
## Description
Use WebRTC for faster transfers

- Capabilities: advertised using avatar_url
    - Backwards compatible, uses and falls back to current Matrix behavior/protocol
    - No Receive: Fail early sending payments to clients advertising it
    - No Mediate: Fail early mediating through partners advertising it
    - No Delivery: skip signing/sending unneeded Delivered messages
- P2P transport with WebRTC!

- WebRTC channels/protocol, through Matrix, falling back to Room messages
- Works in browser & Node.js, Python lib available
- Actual Matrix VoIP protocol for signaling
    - m.call.invite, m.call.answer, m.call.candidates, m.call.hangup
    - Alleviates burden on Matrix servers by using it mostly for discovery/validation/negotiation/signaling
- P2P, UDP, DTLS (mandatory e2e encryption, native fast)
- Can work with just STUN (including public servers like Google’s) and TURN
    - STUN: P2P, bypass NAT, should work on ~90% of networks, reveals IP/endpoint to peer, fastest
    - Still very limited attack surface, browser-hardened API, strict messages/validation
    - Can use TURN server deployed with Synapse to relay messages if it can’t P2P connect with STUN
    - Can force TURN for privacy (relayed, no p2p), slower than STUN, faster than plain Matrix

![grafik](https://user-images.githubusercontent.com/35398162/78012415-35c10000-7345-11ea-874a-76b7d9018084.png)

![grafik](https://user-images.githubusercontent.com/35398162/78012463-47a2a300-7345-11ea-8bd9-a4ccf7a6f258.png)

## Acceptance criteria
- 

## Tasks
- [ ]

nephix:
## Matrix event schemas

* [m.call.answer](https://github.com/matrix-org/matrix-doc/blob/master/event-schemas/examples/m.call.answer)
* [m.call.candidates](https://github.com/matrix-org/matrix-doc/blob/master/event-schemas/examples/m.call.candidates)

nephix:
I have bumped the coverage to 50%

christianbrb:
We should make a release once https://github.com/raiden-network/light-client/pull/1290 is merged. 

@eorituz could test it afterwards with his demo.

christianbrb:
@andrevmatos Is there documentation needed for a developer using this feature?

andrevmatos:
[...]

PR #1290, opened 2020-04-02
WebRTC

andrevmatos:
Implements #1261

- [x] Implement usage of TURN servers provided by matrix, if available
- [x] Tests


raiden-ts/src/transport/epics.ts
```diff
@@ -1193,11 +1203,18 @@ function waitMemberAndSend$(
...
-      defer(() => matrix.sendEvent(roomId, type, content, '')).pipe(
-        mapTo(roomId),
+    mergeMap((via) =>
+      defer(
+        () =>
+          typeof via === 'string'
+            ? matrix.sendEvent(via, type, content, '') // via room
+            : via.send(content.body), // via RTC channel
+      ).pipe(
+        mapTo(typeof via === 'string' ? via : via.label),
```
nephix, line 1217: Maybe better to have something more obvious/self explanatory here, like `Mode.MATRIX ? ... : ...` instead of `typeof via === 'string'`

raiden-ts/src/transport/epics.ts
```diff
@@ -1503,3 +1520,352 @@ export const deliveredEpic = (
...
+  start$: Subject<null>,
+  info: CallInfo,
+  { httpTimeout }: RaidenConfig,
+  deps: Pick<RaidenEpicDeps, 'log' | 'latest$' | 'config$'>,
+) {
+  const { callId, peerId, peerAddress } = info;
+  const { log, latest$, config$ } = deps;
+
+  const connection = new RTCPeerConnection({
+    iceServers: [{ urls: 'stun:stun.l.google.com:19302' }],
```
nephix, line 1587: Maybe move to `RaidenConfig`

raiden-ts/src/transport/epics.ts
```diff
@@ -1503,3 +1520,352 @@ export const deliveredEpic = (
...
+          }).pipe(
+            tap((e) => log.debug('RTC: sent invite', callId, e)),
+            ignoreElements(),
+          ),
+        );
+      }),
+      take(1),
+      tap(() => log.info('RTC: got answer', callId)),
+      map((event) => {
+        connection.setRemoteDescription(new RTCSessionDescription(event.event.content.answer));
```
nephix, line 1631: What is the description used for?

raiden-ts/src/transport/epics.ts
```diff
@@ -1503,3 +1520,352 @@ export const deliveredEpic = (
...
+        ),
+        fromEvent<RTCDataChannelEvent>(connection, 'datachannel').pipe(pluck('channel')),
+      );
+    }),
+    take(1),
+  );
+}
+
+// setup listeners & events for a channel, when it gets opened, and teardown when closed
+function listenChannel$(
```
nephix, line 1703: While reviewing the naming can get confusing: Do we mean payment channels or webrtc channels here? Might be good to be explicit about it

raiden-ts/src/transport/epics.ts
andrevmatos: This is an implementation detail on an local/internal helper function, I don't think it's worth the hassle. Notice the result of this observable isn't used anywhere, so all it does is performed as side-effects and the returned type doesn't matter.

raiden-ts/src/transport/epics.ts
andrevmatos: Yes, it's partly in that firs TODO in this issue, the default/fallback server needs to be configured/defined somewhere else, and we should be able to use the turn servers provided by matrix, if any; thanks for the headsup.

raiden-ts/src/transport/epics.ts
andrevmatos: These description are part of WebRTC protocol, this is how you set the exchanged RTC `offer`/`answer`s, by setting your side with `setLocalDescription` and the other side with `setRemoteDescription`; the exchange itself is done through signaling, in our case, dedicated matrix events.

raiden-ts/src/transport/epics.ts
andrevmatos: I know this can get confusing, but this is how RTC channels are called. Just notice transport knows nothing about Raiden/on-chain channels, so this must be something else. Could you suggest better names than knowing these scopes?

nephix: Not sure, in the spec of WebRTC they are called dataChannels I think, but that doesn't improve it much 😅 maybe simply `rtcChannel`?

raiden-ts/src/transport/epics.ts
nephix: Thank you 🙏

raiden-ts/src/transport/epics.ts
nephix: How would Matrix provide these? I wasn't able to find it in their docs. They already do this for VoIP right?

andrevmatos: To setup: https://github.com/matrix-org/synapse/blob/master/docs/turn-howto.md
To fetch on clients:
https://matrix.org/docs/spec/client_server/latest#get-matrix-client-r0-voip-turnserver
https://github.com/matrix-org/matrix-js-sdk/blob/1834e6688fc819e0d6d45306faa710b2db374997/src/client.js#L4589-L4591

raiden-ts/src/transport/epics.ts
andrevmatos: That would conflict with the action, which is called the same, but something in that direction could be useful.

raiden-ts/src/transport/epics.ts
```diff
@@ -1503,3 +1520,352 @@ export const deliveredEpic = (
...
+  return fromEvent<MatrixEvent>(matrix, 'event').pipe(
+    filter(
+      // FIXME: remove any when MatrixEvent type exposes getAge & getContent methods
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (event: any) =>
+        event.getType() === ('m.call.invite' as EventType) &&
+        event.getSender() === peerId &&
+        event.getContent()?.call_id === callId &&
+        event.getAge() <= (event.getContent()?.lifetime ?? httpTimeout) &&
+        !!event.getContent()?.offer,
```
kelsos, line 1660: nitpick: this filter seems really similar to the other own for the answer, any chance we could some how extract the common part?

raiden-ts/src/transport/epics.ts
andrevmatos: Done, gone with the name `dataChannel`. Thanks for the suggestion

raiden-ts/src/transport/epics.ts
andrevmatos: Done, nice catch.. the function got a little bit big, because of the stricter typing inference on the returned event, but worked for all 4 cases (events), thanks for the suggestion

raiden-ts/src/transport/epics.ts
andrevmatos: Done, set in `fallbackIceServers` in config, and also fetching (with cache) from matrix, if available. Thanks

raiden-ts/src/config.ts
```diff
@@ -99,7 +102,9 @@ export function makeDefaultConfig(
       [Capabilities.NO_DELIVERY]: true,
       [Capabilities.NO_RECEIVE]: true,
       [Capabilities.NO_MEDIATE]: true,
+      [Capabilities.WEBRTC]: true,
     },
+    fallbackIceServers: [{ urls: 'stun:stun.l.google.com:19302' }],
```
nephix, line 107: Do we need to mention this in the privacy policy or get some kind of allowance?

christianbrb: Thanks :) I am picking this up

andrevmatos: I don't think this is needed, these are public servers, and no information pass through them, it's used only as "connection faciliators", with a role not that different from e.g. DNS, and I don't remember we getting an allowance for that.
Nonetheless, there're tons of those public servers available, we can switch to some more respectable provider, and also this will be mostly unused once our matrix servers provide TURN, which I already implemented here, and our own TURN servers are preferred over these fallback STUN servers.

nephix: Yes I just asked because @kelsos mentioned getting approval for hosting on GH pages was quite the hassle and that we should not deploy the dApp somewhere else (e.g. Netlify)
If this is a problem (not sure if it is) then switching to another server wouldn't solve the issue as we would still need users to inform that this could happen in the policy

raiden-ts/src/transport/epics.ts
```diff
@@ -1503,3 +1521,398 @@ export const deliveredEpic = (
...
 };
+
+type CallInfo = { callId: string; peerId: string; peerAddress: Address };
+
+// fetches and caches matrix set turnServer
+const _matrixIceServersCache = new WeakMap<MatrixClient, [number, RTCIceServer[]]>();
+async function getMatrixIceServers(matrix: MatrixClient): Promise<RTCIceServer[]> {
+  const cached = _matrixIceServersCache.get(matrix);
+  if (cached && Date.now() < cached[0]) return cached[1];
+  const fetched = ((await matrix.turnServer()) as unknown) as
```
nephix, line 1532: Can't we use `matrix.getTurnServers()` from [here](https://matrix-org.github.io/matrix-js-sdk/5.2.0/module-client.MatrixClient.html#getTurnServers)? At least the signature looks much easier and I don't understand how this works here because we don't use the `matrix.turnServer(callback)`

andrevmatos: No, unfortunately, because I had to disable matrix-js-sdk's handling of the `m.call.*` events, or it interferes too much in our approach, and disabling it also disables the repeating function  which fetches and populates `getTurnServers` (it's cached internally).
My version here works by using a cache to store previous `turnServer()` response (which is the actual request made) up to the `ttl` of it, or 1h for empty/failing responses, parses it in the format of `RTCIceServer` interface, and returns the cached version of it, if so.

nephix:
We also need a change log entry for the release logs

raiden-ts/src/transport/epics.ts
kelsos: Any change to add more details to where this detail comes from though? Some kind of comment with any information regarding the context?

raiden-ts/src/polyfills.ts
```diff
@@ -25,3 +25,13 @@ request((opts: object, cb: ReqCb) => {
     },
   });
 });
+
+import wrtc from 'wrtc';
+if (!('RTCPeerConnection' in globalThis)) {
```
nephix, line 30: `globalThis` was introduced after node10 IIRC, but we may need to support node10 because it's the last version with official armv6 releases which could be useful for IoT things. 

I remember running into this issue when working with the demo train

nephix: @eorituz said that armv7 support is probably sufficient because armv6 is too old

raiden-ts/tests/unit/epics/transport.spec.ts
```diff
@@ -1477,4 +1478,259 @@ describe('transport epic', () => {
...
+      });
+    });
+
+    test('success callee, receive message & channel error', async () => {
+      expect.assertions(2);
+
+      // change partner to one with smaller address, to be the caller
+      partner = '0x0000000000000000000000000000000000000088' as Address;
+      partnerUserId = `@${partner.toLowerCase()}:${matrixServer}`;
+      matrix.getRoom.mockReturnValue({
```
kelsos, line 1564: This looks the same as the one used above, is there any specific reason for the duplication? Could this be extracted to some method?

raiden-ts/src/polyfills.ts
eorituz: IIRC I was able to run node 12 on an armv6 but anyway. We can still figure out how to hack a version for armv6 if we really need to.

raiden-ts/tests/unit/epics/transport.spec.ts
andrevmatos: It's because I changed the `partner` & `partnerUserId`, I needed to update the object to be returned by `getRoom` and `getMember`. I think it can be replaced by a `mockImplementation*` of some sort, will do it to simplify this part

raiden-ts/src/transport/epics.ts
andrevmatos: I added a comment for it. Otherwise, you can just see, all values from the observables returned by this function is rejected, it's like a wrapper for `sendEvent`, but which also may wait for several reactive conditions, and will send the message when possible, and just return anything to notify it succeeded (or errors otherwise)

PR #1308, opened 2020-04-08
Upload state feature

taleldayekh:


raiden-dapp/src/components/UploadStateDialog.vue
```diff
@@ -0,0 +1,183 @@
+<template>
+  <raiden-dialog class="upload-state" :visible="visible" @close="cancel">
+    <v-card-title>
+      {{ $t('backup-state.upload') }}
+    </v-card-title>
+
+    <v-card-text v-if="dropzoneError">
+      {{ 'There is an dropzone error' }}
```
nephix, line 8: Can you please make this a localized string? Otherwise linting should fail

raiden-dapp/src/locales/en.json
```diff
@@ -180,6 +183,7 @@
   "no-access": {
     "unsupported-network": "The current network is unsupported. Please choose a different network.",
     "sdk-initialization-failure": "SDK initialization failed. Please check the console for more information.",
+    "rdn-state-migration": "Cannot replace state with older uploaded state.",
```
nephix, line 186: In the future we should get rid of these and use the `ErrorMessage` component to provide some more info on these errors. As a user it's really difficult to understand how they can get out of their misery

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -173,6 +179,11 @@ export default class RaidenService {
       let deniedReason: DeniedReason;
       if (e.message && e.message.indexOf('No deploy info provided') > -1) {
         deniedReason = DeniedReason.UNSUPPORTED_NETWORK;
+      } else if (
+        e.message &&
```
nephix, line 183: You should be able to do `e.code === 'RDN_STATE_MIGRATION'`

taleldayekh: Nein, that did not work for me. :)

raiden-dapp/src/components/UploadStateDialog.vue
```diff
@@ -0,0 +1,212 @@
...
+      this.activeDropzone = false;
+    }
+  }
+
+  onDropzoneDrop(e: DragEvent) {
+    e.preventDefault();
+    this.activeDropzone = false;
+
+    if (!e.dataTransfer?.files) {
+      this.dropzoneError();
```
nephix, line 113: If there's an error we can probably stop executing here and `return;`

raiden-dapp/src/components/UploadStateDialog.vue
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    const uploadedFile = e.dataTransfer?.files;
+    this.uploadState(uploadedFile!);
+  }
+
+  onFileSelect(e: Event) {
+    if (!(e.target as HTMLInputElement).files) {
+      this.dropzoneError();
```
nephix, line 122: Same here?

raiden-dapp/src/components/UploadStateDialog.vue
```diff
@@ -0,0 +1,212 @@
...
+  uploadState(uploadedFile: FileList) {
+    if (uploadedFile.length > 1) {
+      this.dropzoneError();
+    }
+
+    let reader = new FileReader();
+    /* istanbul ignore next */
+    reader.onload = e => {
+      if (!e.target?.result) {
+        this.dropzoneError();
```
nephix, line 146: And here

raiden-dapp/src/views/BackupState.vue
```diff
@@ -28,27 +31,32 @@
...
-      </v-list-item> -->
+      </v-list-item>
     </v-list>
     <download-state-dialog
       :visible="downloadState"
       @cancel="downloadState = false"
     >
     </download-state-dialog>
+    <upload-state-dialog :visible="uploadState" @cancel="uploadState = false">
+    </upload-state-dialog>
```
nephix, line 42:
```suggestion
    <upload-state-dialog :visible="uploadState" @cancel="uploadState = false" />
```

raiden-dapp/tests/unit/components/upload-state-dialog.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  test('calls methods for uploading state on file select', async () => {
+    // @ts-ignore
+    wrapper.vm.uploadState = jest.fn();
+    // @ts-ignore
+    expect(wrapper.vm.uploadState).not.toBeCalled();
+
+    const stateInput = wrapper.find('input');
+    // @ts-ignore
+    stateInput.element.value = '';
+    stateInput.trigger('change');
```
nephix, line 89: I think we have a utility function for this [here](https://github.com/raiden-network/light-client/blob/master/raiden-dapp/tests/unit/utils/interaction-utils.ts#L4) so something like
```
mockInput(wrapper, '');
```
would be sufficient

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -83,7 +87,7 @@ export default class RaidenService {
     }
   }
 
-  async connect(subkey?: true) {
+  async connect(stateBackup: string | undefined, subkey?: true) {
```
nephix, line 90: If `stateBackup` is optional one could also do:
```suggestion
  async connect(stateBackup?: string, subkey?: true) {
```

raiden-dapp/tests/unit/raiden-service.spec.ts
```diff
@@ -67,7 +67,7 @@ describe('RaidenService', () => {
     expect.assertions(1);
     providerMock.mockResolvedValue(mockProvider);
     factory.mockResolvedValue(mockRaiden());
-    await raidenService.connect();
+    await raidenService.connect('');
```
nephix, line 70: From looking at `raidenService.connect` it looks like the `stateBackup` is optional
I think it would be nicer to not have to pass an empty string to it. If really necessary one could use [default parameters](https://www.typescriptlang.org/docs/handbook/functions.html#optional-and-default-parameters) as an alternative

raiden-dapp/src/components/UploadStateDialog.vue
```diff
@@ -0,0 +1,212 @@
...
+      }
+
+      try {
+        this.uploadingStateProgress = true;
+        const retrievedState = e.target?.result;
```
kelsos, line 151: I would probably extract the `e.target` to a const at the beginning of the `onload`, check if it is set there and return if it is not. This would stop the execution from proceeding if the target is falsy and will remove the repetition of checking for nullability

raiden-dapp/src/views/BackupState.vue
kelsos: @nephix I think we probably should start enforcing that with `eslint`.

raiden-dapp/src/views/Home.vue
```diff
@@ -86,18 +87,23 @@ export default class Home extends Vue {
     );
   }
 
-  async connect(subkey?: true) {
+  async connect(connectOptions: ConnectOptions) {
+    const stateBackup = connectOptions['uploadedState'] || undefined;
```
kelsos, line 91: is there any specific reason for `connectOptions['uploadedState']` instead of `connectOptions.uploadedState`

raiden-dapp/src/types.d.ts
```diff
@@ -7,6 +7,11 @@ export type Tokens = { [token: string]: Token };
 export type Transfers = { [secretHash: string]: RaidenTransfer };
 export type ChannelAction = 'close' | 'deposit' | 'settle';
 
+export interface ConnectOptions {
+  uploadedState: string;
+  subkey?: true | undefined;
```
kelsos, line 12: `| undefined` here is redundant the `?` implies that it is optional and might have an undefined value

raiden-dapp/src/components/ConnectDialog.vue
```diff
@@ -78,9 +79,8 @@ export default class ConnectDialog extends Vue {
   close() {}
 
   @Emit()
-  connect(subkey?: true) {
-    this.hideClose = true;
-    return subkey;
+  connect(uploadedState: string, subkey?: true) {
+    return { uploadedState, subkey };
```
kelsos, line 83: Is there any specific reason why we don't return the `undefined` here if empty instead of https://github.com/raiden-network/light-client/pull/1308/files#diff-8f4f0c339d26d1e6c6c13add936452d9R91

raiden-dapp/src/views/BackupState.vue
nephix: I have created #1332 for this

nephix:
We also need a change log entry for the release logs

raiden-dapp/src/views/Home.vue
taleldayekh: Nope, the only reason was just me not knowing better. :)

raiden-dapp/CHANGELOG.md
```diff
@@ -19,6 +22,7 @@
 - [#1243] Fix width of transfer input on mobile screens.
 - [#1242] Fix transfer modal not closing instantly.
 
+[#694] https://github.com/raiden-network/light-client/issues/694
```
nephix, line 25: This entry is already there

Issue #1252, opened 2020-03-27
Mediated transfers


nephix:
## User Story

As Chris I want to be able to mediate transfers using the SDK (after #1209 is merged)

## Acceptance criteria

* Not mediating through the dApp
* Everyone understands the approach
* Use via CLI

Issue #1268, opened 2020-04-01
Test walkthrough


christianbrb:
## Description
Test walkthrough and creating issues for technical debt in tests

## Acceptance criteria
- Test walkthrough has taken place by all devs (individually)
- Create issues for the most obvious test debt (where we have got already knowledge)
- At the end all devs are aligned (dedicated call)
- Research how to improve tests

## Tasks
- [ ]

christianbrb:
Feedback from @taleldayekh @nephix needed for **One fixture to rule 'em all** #1377

andrevmatos:
The was done.
Feedback on #1377 , and possibly new suggestion on new issue

PR #1315, opened 2020-04-09
Mediate transfers

andrevmatos:
Implement #1252

### DIFF
--- a/raiden-ts/src/messages/types.ts
+++ b/raiden-ts/src/messages/types.ts
@@ -7,7 +7,7 @@
  */
 import * as t from 'io-ts';
 
-import { Address, Hash, Secret, UInt } from '../utils/types';
+import { Address, Hash, Secret, UInt, Int } from '../utils/types';
 import { Lock } from '../channels/types';
 
 // types
@@ -24,6 +24,7 @@ export enum MessageType {
   WITHDRAW_CONFIRMATION = 'WithdrawConfirmation',
   WITHDRAW_EXPIRED = 'WithdrawExpired',
   PFS_CAPACITY_UPDATE = 'PFSCapacityUpdate',
+  PFS_FEE_UPDATE = 'PFSFeeUpdate',
 }
 
 // Mixin of a message that contains an identifier and should be ack'ed with a respective Delivered
@@ -238,6 +239,29 @@ export const PFSCapacityUpdate = t.readonly(
 );
 export interface PFSCapacityUpdate extends t.TypeOf<typeof PFSCapacityUpdate> {}
 
+export const PFSFeeUpdate = t.readonly(
+  t.type({
+    type: t.literal(MessageType.PFS_FEE_UPDATE),
+    canonical_identifier: t.readonly(
+      t.type({
+        chain_identifier: UInt(32),
+        token_network_address: Address,
+        channel_identifier: UInt(32),
+      }),
+    ),
+    updating_participant: Address,
+    timestamp: t.string,
+    fee_schedule: t.type({
+      cap_fees: t.boolean,
+      // if not null, it should be an array of [tokenAmount, fee] tuples
+      imbalance_penalty: t.union([t.null, t.array(t.tuple([UInt(32), Int(32)]))]),
+      proportional: Int(32),
+      flat: Int(32),
+    }),
+  }),
+);
+export interface PFSFeeUpdate extends t.TypeOf<typeof PFSFeeUpdate> {}
+
 export const Message = t.union([
   Delivered,
   Processed,
@@ -251,6 +275,7 @@ export const Message = t.union([
   WithdrawConfirmation,
   WithdrawExpired,
   PFSCapacityUpdate,
+  PFSFeeUpdate,
 ]);
 // prefer an explicit union to have the union of the interfaces, instead of the union of t.TypeOf's
 export type Message =
@@ -265,5 +290,6 @@ export type Message =
   | WithdrawRequest
   | WithdrawConfirmation
   | WithdrawExpired
-  | PFSCapacityUpdate;
+  | PFSCapacityUpdate
+  | PFSFeeUpdate;
 export type EnvelopeMessage = LockedTransfer | RefundTransfer | Unlock | LockExpired;
--- a/raiden-ts/src/path/epics.ts
+++ b/raiden-ts/src/path/epics.ts
@@ -65,6 +65,11 @@ const PathError = t.readonly(
   }),
 );
 
+// returns a ISO string truncated at the integer second resolution
+function makeTimestamp(time?: Date): string {
+  return (time ?? new Date()).toISOString().substr(0, 19);
+}
+
 const makeIOU = (
   sender: Address,
   receiver: Address,
@@ -103,7 +108,7 @@ const signIOU$ = (iou: IOU, signer: Signer): Observable<Signed<IOU>> =>
 
 const makeAndSignLastIOURequest$ = (sender: Address, receiver: Address, signer: Signer) =>
   defer(() => {
-    const timestamp = new Date().toISOString().split('.')[0],
+    const timestamp = makeTimestamp(),
       message = concat([sender, receiver, toUtf8Bytes(timestamp)]);
     return from(signer.signMessage(message) as Promise<Signature>).pipe(
       map((signature) => ({ sender, receiver, timestamp, signature })),
@@ -495,7 +500,7 @@ export const pfsFeeUpdateEpic = (
           channel_identifier: bigNumberify(channel.id) as UInt<32>,
         },
         updating_participant: address,
-        timestamp: new Date().toISOString().substr(0, 23) + '000',
+        timestamp: makeTimestamp(),
         fee_schedule: {
           cap_fees: true,
           imbalance_penalty: null,
--- a/raiden-ts/src/path/epics.ts
+++ b/raiden-ts/src/path/epics.ts
@@ -31,8 +31,9 @@ import { RaidenAction } from '../actions';
 import { RaidenState } from '../state';
 import { RaidenEpicDeps } from '../types';
 import { messageGlobalSend } from '../messages/actions';
-import { MessageType, PFSCapacityUpdate } from '../messages/types';
+import { MessageType, PFSCapacityUpdate, PFSFeeUpdate } from '../messages/types';
 import { MessageTypeId, signMessage } from '../messages/utils';
+import { channelMonitor } from '../channels/actions';
 import { ChannelState, Channel } from '../channels/state';
 import { channelAmounts } from '../channels/utils';
 import { Address, decode, Int, Signature, Signed, UInt, isntNil } from '../utils/types';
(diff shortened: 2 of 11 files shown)


raiden-ts/src/messages/types.ts
```diff
@@ -238,6 +239,29 @@ export const PFSCapacityUpdate = t.readonly(
 );
 export interface PFSCapacityUpdate extends t.TypeOf<typeof PFSCapacityUpdate> {}
 
+export const PFSFeeUpdate = t.readonly(
```
kelsos, line 242: This is the message we send to the PFS for the fees we are taking when mediating right?

andrevmatos: Yes, exactly, but even though we don't take fees yet, we need to send it at least once for each channel to tell PFS they can mediate through us and to calculate the fees we'll be taking as Zero.

raiden-ts/src/path/epics.ts
```diff
@@ -466,6 +463,57 @@ export const pfsCapacityUpdateEpic = (
...
+ * @param action$ - Observable of channelMonitor actions
+ * @param state$ - Observable of RaidenStates
+ * @param deps - Raiden epic dependencies
+ * @returns Observable of messageGlobalSend actions
+ */
+export const pfsFeeUpdateEpic = (
+  action$: Observable<RaidenAction>,
+  state$: Observable<RaidenState>,
+  { log, address, network, signer, config$ }: RaidenEpicDeps,
+): Observable<messageGlobalSend> =>
```
kelsos, line 480: any specific reason for using GlobalSend instead of broadcast in the naming here?

raiden-ts/tests/unit/epics/mediate.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+} from 'raiden-ts/transfers/utils';
+import { Direction } from 'raiden-ts/transfers/state';
+
+import { epicFixtures } from '../fixtures';
+import { raidenEpicDeps } from '../mocks';
+
+describe('mediate transfers', () => {
+  let depsMock: ReturnType<typeof raidenEpicDeps>;
+  let token: ReturnType<typeof epicFixtures>['token'],
+    tokenNetwork: ReturnType<typeof epicFixtures>['tokenNetwork'],
```
kelsos, line 34: Didn't we start splitting this into different declarations at some point?

raiden-ts/tests/unit/epics/path.spec.ts
```diff
@@ -11,24 +11,28 @@ import {
...
   pfsCapacityUpdateEpic,
   pfsServiceRegistryMonitorEpic,
+  pfsFeeUpdateEpic,
 } from 'raiden-ts/path/epics';
 import { pathFind, pfsListUpdated, iouPersist, iouClear } from 'raiden-ts/path/actions';
 import { messageGlobalSend } from 'raiden-ts/messages/actions';
 import { MessageType } from 'raiden-ts/messages/types';
 import { losslessStringify } from 'raiden-ts/utils/data';
+import { pluckDistinct } from 'raiden-ts/utils/rx';
+import { ErrorCodes } from 'raiden-ts/utils/error';
```
kelsos, line 30: Didn't we have eslint doing import sorting automatically? this change feels like something that would be handled by eslint

raiden-ts/tests/unit/utils.spec.ts
```diff
@@ -328,7 +329,8 @@ describe('data', () => {
     expect(() => encode(-1, 2)).toThrowError('negative');
     expect(() => encode(bigNumberify(65537), 2)).toThrowError('too large');
     expect(() => encode('0x01', 2)).toThrowError(ErrorCodes.DTA_ARRAY_LENGTH_DIFFRENCE);
-    expect(() => encode((true as unknown) as number, 2)).toThrowError('data is not');
+    // @ts-ignore // TODO: replace with @ts-expect-error ts@^3.9
+    expect(() => encode(null, 2)).toThrowError('data is not');
```
kelsos, line 333: Could you explain the commend above, also I am trying to get the `data is not` will that reach the SDK user at some point?  Looking this my question would be "data is not what?"

raiden-ts/src/path/epics.ts
andrevmatos: Sorry, I don't understand what you mean, can you rephrase?
There're 2 actions for sending messages:
- `messageSend.request` is based on `meta.address` to send to a specific Address (transport will take care of inviting the user, waiting for it to be online and joined before sending, etc)
- `messageGlobalSend` instead takes the name of one of the joined global rooms, and send it to it, without waiting for replies (as we don't even listen on these rooms).

The 2nd one is the one used here, it sends the output message once to a global room (in this case, `raiden_<network>_path_finding`); this is the only way to broadcast a message

raiden-ts/tests/unit/epics/mediate.spec.ts
andrevmatos: Not for fixtures, only for actions. This is just a way to refer to the declaration of the given function, instead of having to retype the specific types everytime; this is only needed to avoid defining these variables at declaration time, since they'll be re-set on `beforeEach`

raiden-ts/tests/unit/epics/path.spec.ts
andrevmatos: Not sure if that includes tests, it didn't get sorted automatically by my prettier plugin, so I suppose it's not into effect there. No big deal anyway for tests.

raiden-ts/src/path/epics.ts
kelsos: `global send` means `broadcast`, right? What I am asking, since there is a well-known term aka `broadcast`, is there is any specific reason that the action is not named `messageBroadcast` instead?

kelsos: I am just curious :)

raiden-ts/tests/unit/utils.spec.ts
andrevmatos: This function accepts some types and encode them to serialized forms of given size. This test ensures that, when passed a type that isn't handled, it throws. Before, it was boolean; now I handle boolean and therefore had to change the test to something else, i.e. `null`.

To test that, I need to pass a type which isn't accepted by the function declaration. As TS would complain, I need to ts-ignore (which is bad). On TypeScript 3.9 beta, there's a new declaration specifically for this kind of usecase: `// @ts-expect-error`, which makes TS accept something wrong from its perspective (useful for unit tests), but differently from ts-ignore, it'll error if there's no compile-time error on that line, so you get a more strict validation of errors. E.g. if in the future we start accepting `null`, `@ts-expect-error` would complain, while `@ts-ignore` would just ignore; of course in this case, the test would fail as well, but it's useful to have proper typing at development time as well.

The message thrown is variable and dependent of the input type, that's why I just include a substring of it here, just to be notified if the message changes for some reason (e.g. it becomes a `TypeError` due to some change in the encoding algorithm while I expected this specific string error). I could put some more specific part of it there, but I didn't think it was worth it, but I'll change that.

raiden-ts/src/path/epics.ts
andrevmatos: Yes, `broadcast` also means `global send` (i.e. send to a global/broadcast room); I just wanted to make a parallel with `messageSend`, since they work almost the same from transport's perspective, simply one sends to an address and the other sends to a global/broadcast room. This was just naming, since the rooms were already named `global` on transport and config.

raiden-ts/tests/unit/utils.spec.ts
andrevmatos: Anyway, that error should not reach users at any point, it's something for internal helper/utility function, more of an assert.

raiden-ts/tests/unit/epics/path.spec.ts
kelsos: It should be a big deal, we should treat our tests the same as our production code. :D 
not for this issue but I will create a note in the iteration to have a look.

raiden-ts/tests/unit/epics/mediate.spec.ts
kelsos: I would suggest starting to follow the same rules regardless of tests and production code. I don't think it makes sense to have a different approach between production and test code. This can only lead to confusion.

raiden-ts/tests/unit/utils.spec.ts
kelsos: Ok thanks for clarifying, if it is only internal then no need to change :)

raiden-ts/tests/unit/epics/mediate.spec.ts
andrevmatos: I agree, test & production should follow the same rules, but specifically on this case, it's not comparable. The actions looks like simple functions, but they're redefined on each action declaration time (so much there're helpers for creating them), and there, it makes sense to have specific interfaces for the returned type (in that case, the actual action object, used in multiple places). Here, it's two one-shot functions (`raidenEpicDeps` & `epicFixtures`), which are, well, fixtures, and the second one returns a bunch of values which we destructure on the body of the suites, so these `ReturnType` are just a way to avoid re-declaring the type of these values and getting them out of sync if we change something.
It's the same (and I'm considering refactoring to it) as doing something like `let fix; fix = epicFixtures(...)`, and then access each value as `fix.value`, as it's no big deal, would just add a couple of characters before fixtures/constants usage. If you think the `ReturnType` types are bad, we could move to that, and that probably would/will be part of the #1268 issue and overall code refactoring.

raiden-ts/src/path/epics.ts
```diff
@@ -466,6 +463,57 @@ export const pfsCapacityUpdateEpic = (
...
+
+      const message: PFSFeeUpdate = {
+        type: MessageType.PFS_FEE_UPDATE,
+        canonical_identifier: {
+          chain_identifier: bigNumberify(network.chainId) as UInt<32>,
+          token_network_address: action.meta.tokenNetwork,
+          channel_identifier: bigNumberify(channel.id) as UInt<32>,
+        },
+        updating_participant: address,
+        timestamp: new Date().toISOString().substr(0, 23) + '000',
```
nephix, line 498: This looks rather hacky plus we already use a timestamp that PFS understands when dealing with IOUs. Probably makes sense to unify these two into one approach

andrevmatos: Unfortunately, this is not "what PFS understands", but instead, a very strict format which goes in the signature, so it needs to be exact. The signature expects a ISO date string with 26 characters, with microseconds resolution, and it does not accept the `Z` at the end of milliseconds which usually goes with JS Date to ISO string. So the `substr` is to ensure we "stop" at the milliseconds, and then we add zeroes to the microseconds. In the end, it's a string, and gets validated as string (see `packMessage`), but needs to be very specific, and I didn't find a standard way to do this on JS.

nephix: I know, but probably the same as in https://github.com/raiden-network/light-client/blob/master/raiden-ts/src/path/epics.ts#L105 or do we need to have two separate ways of handling it?

andrevmatos: Hmm. As you can see below, python code does send with microseconds:
image
But I tested it by hand, and python's `PFSFeeUpdate` marshmallow can indeed deserialize with only the second resolution there, which is the same format as used in the signature and on the IOU, so I'm changing to standardize it there.
Thanks for the headsup

PR #1333, opened 2020-04-15
Disable download state button if disconnected

### LINKED ISSUE #1310: Downloading app state should not be possible while disconnected
## Steps to Reproduce
1. Go to https://lightclient.raiden.network/
2. Click on identicon
3. Click on Back up state
4. Click on Download State

## Expected Result
Nothing happens, it should be disabled because there is no web3 provider
Maybe a hint/tooltip why downloading state is disabled

## Actual Result
Warning modal, which after confirm downloads a state file that contains `undefined`

nephix:
Fixes #1310

### DIFF
--- a/raiden-dapp/tests/unit/views/backup-state.spec.ts
+++ b/raiden-dapp/tests/unit/views/backup-state.spec.ts
@@ -2,6 +2,8 @@ jest.useFakeTimers();
 import { mount, Wrapper } from '@vue/test-utils';
 import Vue from 'vue';
 import Vuetify from 'vuetify';
+
+import store from '@/store/index';
 import BackupState from '@/views/BackupState.vue';
 
 Vue.use(Vuetify);
@@ -18,7 +20,8 @@ describe('BackupState.vue', () => {
       stubs: ['v-dialog'],
       mocks: {
         $t: (msg: string) => msg
-      }
+      },
+      store
     });
   });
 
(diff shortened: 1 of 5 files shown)


raiden-dapp/tests/unit/views/backup-state.spec.ts
```diff
@@ -44,6 +47,14 @@ describe('BackupState.vue', () => {
...
 
+  test('download state button disabled if disconnected', () => {
+    const downloadStateButton = wrapper.find(
+      '.backup-state__buttons__download-state'
+    );
+
+    expect(downloadStateButton.classes()).toContain('v-list-item--disabled');
+  });
+
   // test('upload state', () => {
```
kelsos, line 58: I know it's unrelated to the PR but any idea why we have a commented test here?

nephix: Probably the upload/download state ticket was too big and got split up
They will be implemented in #1308

nephix:
Codecov's reporting doesn't make any sense here. I'm force-merging this

raiden-dapp/tests/unit/views/backup-state.spec.ts
kelsos: Ok but it would be nice to avoid commented code in the repo in the future :)

PR #1344, opened 2020-04-16
Allow non-https PFS addresses if not running on production

### LINKED ISSUE #1338: Allow PFS addresses without https
## User Story

Necessary for #1211

As a developer I may want to use a PFS that may not use HTTPS while developing. We could  enforce HTTPS only if `NODE_ENV` isn't `production`

## Acceptance criteria

* HTTPS only necessary for production
* Ports should be allowed as well

nephix:
Resolves #1338

### DIFF
--- a/raiden-ts/src/helpers.ts
+++ b/raiden-ts/src/helpers.ts
@@ -369,6 +369,6 @@ export const isValidUrl = (url: string): boolean => {
   const regex =
     process.env.NODE_ENV === 'production'
       ? /^(?:https:\/\/)?[^\s\/$.?#&"']+\.[^\s\/$?#&"']+$/
-      : /^(?:(https|http):\/\/)?[^\s\/$.?#&"']+\.[^\s\/$?#&"']+$/;
+      : /^(?:(http|https):\/\/)?([^\s\/$.?#&"']+\.)*[^\s\/$?#&"']+(?:(\d+))*$/;
   return regex.test(url);
 };
--- a/raiden-ts/src/helpers.ts
+++ b/raiden-ts/src/helpers.ts
@@ -357,3 +357,18 @@ export async function waitConfirmation(
     )
     .toPromise();
 }
+
+/*
+ * Returns true if `url` is a valid URL or domain.
+ * On production `https://` is required for URLs, otherwise `http://` matches as well.
+ *
+ * @param url - A URL or hostname
+ * @returns true if valid URL or domain
+ */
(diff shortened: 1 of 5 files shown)


raiden-ts/src/helpers.ts
```diff
@@ -357,3 +357,18 @@ export async function waitConfirmation(
...
+ * On production `https://` is required for URLs, otherwise `http://` matches as well.
+ *
+ * @param url - A URL or hostname
+ * @returns true if valid URL or domain
+ */
+export const isValidUrl = (url: string): boolean => {
+  const regex =
+    process.env.NODE_ENV === 'production'
+      ? /^(?:https:\/\/)?[^\s\/$.?#&"']+\.[^\s\/$?#&"']+$/
+      : /^(?:(https|http):\/\/)?[^\s\/$.?#&"']+\.[^\s\/$?#&"']+$/;
```
kelsos, line 372: You have to allow the port number too in the `development` environment the container running the tests has pfs at `http://localhost:6000` this should also be considered a valid URL in develop

nephix: I guess makes sense for production as well?

kelsos: @nephix no not really. In production this would be https where the port `443` is omitted

raiden-ts/src/path/utils.ts
```diff
@@ -93,9 +91,9 @@ export function pfsInfo(
     withLatestFrom(config$),
     mergeMap(([url, { httpTimeout }]) => {
       if (!url) throw new RaidenError(ErrorCodes.PFS_EMPTY_URL);
-      else if (!urlRegex.test(url)) throw new RaidenError(ErrorCodes.PFS_INVALID_URL, { url });
+      else if (!isValidUrl(url)) throw new RaidenError(ErrorCodes.PFS_INVALID_URL, { url });
       // default to https for domain-only urls
-      else if (!url.startsWith('https://')) url = `https://${url}`;
+      else if (!url.startsWith('https://') && !url.startsWith('http://')) url = `https://${url}`;
```
kelsos, line 96: Any change we could also add a test for this case too?

nephix: Done by https://github.com/raiden-network/light-client/pull/1344/files#diff-4d311b8a210fcb7f2c2090f0aec4ca2dR228

kelsos: I was referring to the different with `http`

Issue #1305, opened 2020-04-07
dApp doesn't go back to "connect" screen upon shutdown


andrevmatos:
## Steps to Reproduce
1. Trigger some unexpected error (e.g. by changing slightly the transport's `acess_token` on state when app is not connected)

## Expected Result
- `raidenShutdown` action goes through, and dApp shows "Connect" screen

## Actual Result
- dApp stays at "Transfer" screen, with all controls enabled, even though it can't be used anymore because SDK did shutdown by unexpected error

## Additional Infos (Optional) 
```
0.5.2
```
Probably related to #1168 or some other change related to making the general screen the landing page/connect screen of dApp

christianbrb:
In the done column within the project, but not closed. Closing

PR #1366, opened 2020-04-17
Redirect on shutdown fix

taleldayekh:
This seems to fix it when testing manually after going through the steps to reproduce #1305 .

### DIFF
--- a/raiden-dapp/CHANGELOG.md
+++ b/raiden-dapp/CHANGELOG.md
@@ -12,9 +12,9 @@
 - [#1306] Redirects to home view upon page refresh
 - [#1310] Disable state download if not connected to web3 provider.
 
-[#1305] https://github.com/raiden-network/light-client/issues/1305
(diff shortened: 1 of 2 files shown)


raiden-dapp/CHANGELOG.md
```diff
@@ -1,24 +1,35 @@
...
 
 ### Fixed
+
+- [#1305] Redirects to home view upon shutdown
+- [#1306] Redirects to home view upon page refresh
 - [#1310] Disable state download if not connected to web3 provider.
 
+[#1305] https://github.com/raiden-network/light-client/issues/1305
+[#1306] https://github.com/raiden-network/light-client/issues/1306
 [#1222] https://github.com/raiden-network/light-client/issues/1222
```
nephix, line 17: Thee `:` is needed otherwise it will mess with the markdown rendering like here: https://github.com/raiden-network/light-client/blob/master/raiden-dapp/CHANGELOG.md

PR #1391, opened 2020-04-22
Better error handling if no route can be found by PFS

### LINKED ISSUE #1382: No route between nodes handled as common error
## Steps to Reproduce
Same as on https://github.com/raiden-network/raiden-hub/issues/11

## Expected Result
If there is no route, we explain that to the user and what he can do about it
SDK should also have a separate RaidenError for this, not just the generic `PFS_ERROR_RESPONSE` 

## Actual Result
A RaidenError with:
> errorCode: 2201
> errors: “No route between nodes found.”

Screenshot 2020-04-21 at 14 20 38

nephix:
Fixes #1382

### DIFF
--- a/raiden-dapp/src/locales/en.json
+++ b/raiden-dapp/src/locales/en.json
@@ -471,7 +471,7 @@
     },
     "PFS_NO_ROUTES_BETWEEN_NODES": {
       "title": "No route between you and your target",
-      "problem": "The Pathfinding Service was not able to find a route between you and the target you want to transfer tokens to.",
+      "problem": "The Pathfinding Service was not able to find a usable route between you and the target you want to transfer tokens to.",
       "solution": [
         "Hit the ← button in the top left",
         "Click on the current token network",
--- a/raiden-dapp/src/locales/en.json
+++ b/raiden-dapp/src/locales/en.json
@@ -468,6 +468,18 @@
         "File a new issue on https://github.com/raiden-network/light-client",
         "Upload logs to the issue"
       ]
+    },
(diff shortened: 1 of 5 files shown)


raiden-dapp/src/locales/en.json
```diff
@@ -468,6 +468,18 @@
         "File a new issue on https://github.com/raiden-network/light-client",
         "Upload logs to the issue"
       ]
+    },
+    "PFS_NO_ROUTES_BETWEEN_NODES": {
+      "title": "No route between you and your target",
+      "problem": "The Pathfinding Service was not able to find a route between you and the target you want to transfer tokens to.",
```
andrevmatos, line 474: `find a usable route`; there may be a route but it is unusable for more than one reason: not enough capacity on our side or one of the mediating channels; some peer not online; etc

raiden-dapp/src/locales/en.json
```diff
@@ -468,6 +468,18 @@
...
+      "solution": [
+        "Hit the ← button in the top left",
```
andrevmatos, line 476: Maybe it'd be worth it to, instead of describing each click the user needs to do, just to tell them they may solve it by connecting to a better connected hub? Not sure, just a thought

nephix: That'd be ok if the average user would know what a hub is. 

Unfortunately we don't do a good job of explaining what it is at the moment and have also identified this in an on-site meeting last year, so I think being explicit here is the right thing to do

PR #1398, opened 2020-04-23
Summary screen amounts expands to display exact amounts

taleldayekh:


raiden-dapp/src/components/AmountDisplay.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <div
+    @mouseover="exactAmount ? (displayExactAmount = true) : null"
+    @mouseleave="exactAmount ? (displayExactAmount = false) : null"
+  >
+    <span v-if="displayExactAmount">{{ amount }} {{ symbol }}</span>
+    <span v-else>
+      {{ amount | displayFormat(amount.decimals) }} {{ symbol }}
```
nephix, line 8: if `amount` is a string this would probably fail because it doesn't have `decimals`

taleldayekh: Yes :) I will pass a token instead.

kelsos:
@taleldayekh I am getting the following warning in the `TransferSteps` is that the same for you?
```
[Vue warn]: Property or method "on" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.

found in

---> <TransferSteps> at src/views/TransferSteps.vue
       <VContent>
         <VApp>
           <App> at src/App.vue
             <Root>
```

taleldayekh:
> I am getting the following warning in the `TransferSteps` is that the same for you?

Give me a minute and I'll take a look.

PR #1419, opened 2020-04-24
Initial version of standalone cli app

### LINKED ISSUE #1401: CLI bundling does not work
## Description

Currently in the `cli` branch, running `npm run build:bundle` does not seem to work and makes testing for the demo train difficult. 

CC @eorituz 

## Acceptance criteria
- `npm run build:bundle` creates a single `bundle.js` that includes all dependencies

andrevmatos:
Fix #1401 
Although still limited exposed functionality, I think we can already merge the `cli` branch to `master`.
It isn't part of the official functionality yet, it's like an internal testing tool, but I think it's better to have it kept in sync with `master` already.

### DIFF
--- a/raiden-cli/src/index.ts
+++ b/raiden-cli/src/index.ts
@@ -1,0 +1,320 @@
+/* eslint-disable @typescript-eslint/camelcase */
+// imports
+import { promises as fs } from 'fs';
+import * as path from 'path';
+import { Server } from 'http';
+
+import inquirer from 'inquirer';
+import yargs from 'yargs';
+import { first } from 'rxjs/operators';
+import { LocalStorage } from 'node-localstorage';
+import { Wallet } from 'ethers';
+import { BigNumberish, parseEther } from 'ethers/utils';
+import express, { Express } from 'express';
+import logging from 'loglevel';
+
+import { Raiden, RaidenChannel, RaidenTransfer } from 'raiden-ts';
+
+// global declarations
+let raiden: Raiden;
+let log: logging.Logger = logging;
+let app: Express;
+let server: Server;
+
+async function raidenGetChannel(token: string, partner: string) {
+  const channels = await raiden.channels$.pipe(first()).toPromise();
+  return channels?.[token]?.[partner];
+}
+
+async function raidenEnsureChannelIsFunded(
+  channel: RaidenChannel,
+  deposit: BigNumberish,
+  mint?: boolean,
+) {
+  const start = Date.now();
+  if (channel.ownDeposit.lt(deposit)) {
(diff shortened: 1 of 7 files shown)


raiden-cli/src/index.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+  }
+  return channel;
+}
+
+async function raidenTransfer(token: string, target: string, value: BigNumberish) {
+  let revealAt: number;
+  let unlockAt: number;
+  await raiden.getAvailability(target);
+  const tId = await raiden.transfer(token, target, value);
```
kelsos, line 79: nitpick: this is transferId right? any change we can avoid small variable names

raiden-cli/src/index.ts
```diff
@@ -0,0 +1,311 @@
...
+
+  const t = await new Promise<RaidenTransfer>((resolve, reject) => {
```
kelsos, line 81: nitpick: also here I would rather avoid single letter variable names

raiden-cli/src/index.ts
```diff
@@ -0,0 +1,311 @@
...
+        type: 'string',
+        desc: 'Open a channel with given partner',
+        implies: 'token',
+      },
+      deposit: {
+        type: 'string',
+        desc: 'Deposit this value to channel with partner',
+        implies: 'partner',
+      },
+      mint: {
```
kelsos, line 238: I would argue that in the future minting/opening channels would make more sense as something exposed through the Rest API instead of the CLI

raiden-cli/src/index.ts
```diff
@@ -0,0 +1,311 @@
...
+        desc: 'Serve HTTP API on given port',
+      },
+    })
+    .env('RAIDEN')
+    .help().argv;
+
+  if (!argv.password)
+    argv.password = readline.question('Private Key Password: ', { hideEchoBack: true });
+
+  const pk = await Wallet.fromEncryptedJson(
```
kelsos, line 265: nitpick: again variable name

raiden-cli/src/index.ts
andrevmatos: Yes, there's a paragraph about this on the README, for now, we expose only a limited set of features either on cli parameters and on REST API.
This one is definitely something that I want exposed in the API too, but having open + mint + deposit + 1-transfer exposed through arguments allow a specific usage of the CLI: integration testing, where we want it to sync on the chain, perform some operations adn exit.
But as soon as we have more time to dedicate to this API, of course I want all the featureset exposed through REST. If you have time to work on this, it'd be great ;)

Issue #1395, opened 2020-04-23
Links to the dApp don't work because of early redirect


kelsos:
## Description

Routes in the dApp are not supposed to be evaluated before pressing connect. This happens because before the connection we don't know about the validity of a token, and we can only make sure that a route is valid or not after connecting and getting access to the valid tokens.

This enables us to use links to the transfer and connect screens with pre-filled values, to make it easier for the user to connect. Such links can be used for example in the Hub to make it easier for the user to open a channel with the hub. 

Unfortunately at the moment, this doesn't work and the redirect happens almost instantly instead.

## Acceptance criteria
- Links to the various screens should not redirect you to the home page unless you pressed connect

## Tasks
- [ ]

PR #1420, opened 2020-04-24
Fix undefined config

andrevmatos:
Needed for #1395

### DIFF
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -71,6 +71,7 @@ import { RaidenError, ErrorCodes } from './utils/error';
 export class Raiden {
   private readonly store: Store<RaidenState, RaidenAction>;
   private readonly deps: RaidenEpicDeps;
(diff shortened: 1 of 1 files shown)


raiden-ts/src/raiden.ts
```diff
@@ -71,6 +71,7 @@ import { RaidenError, ErrorCodes } from './utils/error';
 export class Raiden {
   private readonly store: Store<RaidenState, RaidenAction>;
   private readonly deps: RaidenEpicDeps;
+  public config: RaidenConfig;
```
kelsos, line 74: Can it also been modified by hand by `raiden.config = {}` and do we want to do that?

andrevmatos: It could, but it'd be replaced on next `config$` emit; I don't think we need to bother, SDK tries best practices, but if a developer wants to shoot their own foot, I'm not gonna stop them

Issue #1368, opened 2020-04-20
I want my balance proof to be monitored when I receive a transfer


christianbrb:
## User Story
As Chris I want my balance proof to be monitored when I receive a transfer so that I can go offline afterward.

Number 1 split from https://github.com/raiden-network/light-client/issues/1352#issuecomment-615241579

## Acceptance criteria
- I already got service tokens in the UDC
- [A fixed service token amount](https://github.com/raiden-network/light-client/issues/1369#issuecomment-616406672) is submitted as monitoring reward
- It is triggered on each receiving transfer

christianbrb:
See https://github.com/raiden-network/light-client/issues/1369#issuecomment-616406672 for the calculation of the monitoring reward.

christianbrb:
@andrevmatos to take a deeper look into it and we are going to discuss it on Monday

PR #1437, opened 2020-04-29
Rename path(finder) submodule to services

andrevmatos:
Needed for #1368 
Renames the `src/path` submodule to `src/services`, so it can be used for monitoring as well.
Just moving code around, no logic change (yet). The change is purely internal.


raiden-ts/src/channels/reducer.ts
```diff
@@ -216,7 +216,8 @@ const pendingTxs: Reducer<RaidenState['pendingTxs'], RaidenAction> = (
  * name of the reducer. channels root reducer instead must be handled the complete state instead,
  * so it compose the output with each key/nested/combined state.
  */
-export const channelsReducer = partialCombineReducers(
+const channelsReducer = partialCombineReducers(
   { blockNumber, tokens, channels, pendingTxs },
   initialState,
 );
+export default channelsReducer;
```
kelsos, line 223: Any specific reason you decided to go with the default export now?

andrevmatos: Mostly for consistency with other modules which export a single value. Also, on renaming of the module, I had to rename the property on the imports as well (`pathReducer` to `servicesReducer`), while as it was the only function exported, it didn't matter, only its location on the import, so changing it also makes one less place to rename on future refactors (which I hope won't help, but just to be consistent).

PR #1438, opened 2020-04-29
Monitoring

### LINKED ISSUE #1368: I want my balance proof to be monitored when I receive a transfer

andrevmatos:
Closes #1368
When receiving a transfer, if there's enough UDC deposit and `config.monitoringReward` is set, sign and send a `RequestMonitoring` message to the monitoring global room, so we can safely go offline.

### DIFF
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -958,13 +958,21 @@ export const channelSettleEpic = (
       };
       let part1 = {
           address: partner,
-          ...(channel.partner.balanceProof || zeroBalanceProof),
+          ...zeroBalanceProof,
+          ...channel.partner.balanceProof,
         },
         part2 = {
           address,
-          ...(channel.own.balanceProof || zeroBalanceProof),
+          ...zeroBalanceProof,
+          ...channel.own.balanceProof,
         };
-      if (channel.isFirstParticipant) [part1, part2] = [part2, part1];
+
+      if (
+        part2.transferredAmount
+          .add(part2.lockedAmount)
+          .lt(part1.transferredAmount.add(part1.lockedAmount))
+      )
+        [part1, part2] = [part2, part1];
 
       // send settleChannel transaction
       return from(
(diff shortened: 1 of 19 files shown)


raiden-ts/src/channels/epics.ts
```diff
@@ -788,7 +788,7 @@ export const channelCloseEpic = (
       const closingMessage = concat([
         encode(tokenNetwork, 20),
         encode(network.chainId, 32),
-        encode(1, 32), // raiden_contracts.constants.MessageTypeId.BALANCE_PROOF
+        encode(MessageTypeId.BALANCE_PROOF, 32),
```
kelsos, line 791: Nice small addition here :)

raiden-ts/src/messages/utils.ts
```diff
@@ -26,13 +26,17 @@ const CMDIDs: { readonly [T in MessageType]: number } = {
...
   [MessageType.PFS_FEE_UPDATE]: -1,
+  [MessageType.MONITOR_REQUEST]: -1,
 };
 
 // raiden_contracts.constants.MessageTypeId
 export enum MessageTypeId {
   BALANCE_PROOF = 1,
+  BALANCE_PROOF_UPDATE = 2,
   WITHDRAW = 3,
+  COOP_SETTLE = 4,
```
kelsos, line 37: This was added for the sake of completion right? it is not used anywhere even in the python client.

andrevmatos: Yes, exactly, but we may want to add this feature at some point in the future, seems supported by the SC

PR #1455, opened 2020-05-02
Refactor state types

### LINKED ISSUE #1462: Simplify state schemas
## Description

State types and schemas need some refactoring. Most of it was created on the fly during early development and features implementation, and can use a 2nd look.
Mostly:
- `ChannelEnd` has too much optional parameters which need constant checks on presence: `balanceProof?.`, `locks ?? []`, etc.. These should be replaced by actually _empty/zeroed_ values, always defined.
- `Channel.state` is a too complex, non-shared tagged union of objects. Most of its data is available only of `state !== 'opening'`, and this state was only needed as a _holder_ for the channel, but which constantly need complex unneeded branch checking. We should remove `ChannelState.opening` (`channelOpen.request` is going to fail if duplicated anyway), and share as much data as possible across different state/tagged unions.
- `Channel` doesn't contain complete info, and requires external `tokenNetwork`, `partner` address, `token`, etc, which depends on external state, and `tokenNetwork` + `partner` is used as unique key for channel state access. This makes it hard to access some fo these info, or if at some point the unique key needs to be changed (e.g. if channel stops being unique per tokenNetwork per partner), besides the deep structure.
  - We should have an explicit unique key for channel, which is used to index a shallow mapping to the channel states
  - Channel state contains all relevant info: tokenNetwork, partner address, token address
- We lose channel data when it gets settled. As Channel now stores the complete info, it can be useful to store channels in an `oldChannels` mapping after settled, also indexed by a unique key (now also considering uniqueness on time as well, e.g. by including channel Id), so it can be used for reference in the future. `Channel.state` for finally uses `settled`, and `settleBlock` is also persisted for it.
- Transport data is always about matrix. Currently, `matrix` is an optional object inside transport, which needs constant optional checking. As with webrtc, we've been settling on matrix as transport base layer, we should move matrix transport auth data to toplevel `transport` to avoid uneeded deep access
- Same for `path.iou`

All these changes are purely internal, and doesn't touch public API. Channel references on channel methods is still `token/partner`, and this pair is still used for 2-level-deep nesting on `channels$` observable.

## Acceptance criteria
- Changes proposed above are implemented

## Tasks
- [ ]

andrevmatos:
Fixes #1462 

**Short description**
Refactoring of channel state schema, to make it simpler and reduce nesting and improve type safety.
See issue for complete description of changes.


raiden-ts/src/channels/epics.ts
```diff
@@ -769,21 +767,14 @@ export const channelCloseEpic = (
...
-          channel.partner.balanceProof.transferredAmount,
-          channel.partner.balanceProof.lockedAmount,
-          channel.partner.balanceProof.locksroot,
-        );
-        nonce = channel.partner.balanceProof.nonce;
-        additionalHash = channel.partner.balanceProof.messageHash;
-        nonClosingSignature = channel.partner.balanceProof.signature;
-      }
+      const balanceHash = createBalanceHash(
+        channel.partner.balanceProof.transferredAmount,
```
kelsos, line 771: nitpick: Would it be easy to extract `balanceProof` to a variable to a variable instead of repeating `channel.partner.balanceProof` everywhere?

raiden-ts/src/channels/reducer.ts
```diff
@@ -33,96 +31,103 @@ const tokens = createReducer(initialState.tokens).handle(
...
-    own: { deposit: Zero as UInt<32> },
-    partner: { deposit: Zero as UInt<32> },
     id: action.payload.id,
+    token: action.payload.token,
+    tokenNetwork: action.meta.tokenNetwork,
     settleTimeout: action.payload.settleTimeout,
     isFirstParticipant: action.payload.isFirstParticipant,
     openBlock: action.payload.txBlock,
+    own: {
+      ...emptyChannelEnd,
```
kelsos, line 84: This channel end is the same structure as the one in the python client right?

raiden-ts/src/channels/reducer.ts
```diff
@@ -192,32 +204,14 @@ const channels: Reducer<RaidenState['channels'], RaidenAction> = createReducer(
...
+ * blockNumber, tokens & pendingTxs reducers get its own slice of the state, corresponding to the
  * name of the reducer. channels root reducer instead must be handled the complete state instead,
  * so it compose the output with each key/nested/combined state.
  */
-const channelsReducer = partialCombineReducers(
-  { blockNumber, tokens, channels, pendingTxs },
-  initialState,
-);
+const partialReducer = partialCombineReducers({ blockNumber, tokens, pendingTxs }, initialState);
+// channels reducer is a reduce-reducers like reducer
```
kelsos, line 214: This sounds like a tongue twister :smile:, can we clarify what a `reduce-reducers like reducer` is in the comment?

raiden-ts/src/channels/types.ts
```diff
@@ -19,7 +23,7 @@ export type Lock = t.TypeOf<typeof Lock>;
  * because BP signature requires the hash of the message, for authentication of data not included
  * nor relevant for the smartcontract/BP itself, but so for the peers (e.g. payment_id)
  */
-export const SignedBalanceProof = t.type({
+export const BalanceProof = t.type({
```
kelsos, line 26: I am asking here, I might find it later but how are we handling the signed balance proofs now?

raiden-ts/src/messages/utils.ts
```diff
@@ -285,14 +285,14 @@ export function getMessageSigner(message: Signed<Message>): Address {
...
+ * Get the signed BalanceProof associated with an EnvelopeMessage
  *
  * @param message - Signed EnvelopeMessage
- * @returns SignedBalanceProof object for message
+ * @returns Signed BalanceProof object for message
  */
 export function getBalanceProofFromEnvelopeMessage(
   message: Signed<EnvelopeMessage>,
-): SignedBalanceProof {
+): Signed<BalanceProof> {
```
kelsos, line 295: Ok I got my answer here :+1: The review interface is really inconvinient for such big PRs you cannot easily get the context.

raiden-ts/src/migration/3.ts
```diff
@@ -0,0 +1,59 @@
...
+  channelId: Zero,
+  nonce: Zero,
+  transferredAmount: Zero,
+  lockedAmount: Zero,
+  locksroot: HashZero,
+  additionalHash: HashZero,
+  signature: SignatureZero,
+};
+
+function migrateChannelEnd(channel: any, endsAddr: any) {
```
kelsos, line 19: I don't seem to remember but do we backup the old state before the migration? what happens if for some unforseen reason the migration fails?

raiden-ts/src/migration/3.ts
```diff
@@ -0,0 +1,59 @@
...
+  for (const [tokenNetwork, partnerChannels] of Object.entries<any>(state.channels)) {
+    for (const [partner, channel] of Object.entries<any>(partnerChannels)) {
+      migrateChannelEnd(channel, { own: state.address, partner });
+      // rename messageHash to additionalHash, remove unused sender from BP
+      const token = findKey(state.tokens, (tn) => tn === tokenNetwork);
+      Object.assign(channel, { token, tokenNetwork });
+      Object.assign(newChannels, { [`${partner}@${tokenNetwork}`]: channel });
+    }
+  }
+  Object.assign(state, { channels: newChannels, oldChannels: {} });
```
kelsos, line 54: Before this PR we would not keep the old channels anywhere right?

raiden-ts/src/raiden.ts
```diff
@@ -990,7 +991,7 @@ export class Raiden {
   public async getUDCCapacity(): Promise<BigNumber> {
     const balance = await this.deps.latest$.pipe(pluck('udcBalance'), first(isntNil)).toPromise();
     const blockNumber = this.state.blockNumber;
-    const owedAmount = Object.values(this.state.path.iou)
+    const owedAmount = Object.values(this.state.iou)
```
kelsos, line 994: Was `iou` the only thing we stored in the path state? that is why it was moved to the root state?

raiden-ts/src/services/epics.ts
```diff
@@ -389,27 +389,10 @@ export const pathFindServiceEpic = (
   );
 };
 
-function channelEntries(channels: RaidenState['channels']): ReadonlyArray<[string, Channel]> {
```
kelsos, line 392: So with the new channel structure this is not needed anymore, because channels contain all the needed information?

raiden-ts/src/transfers/epics/locked.ts
```diff
@@ -69,12 +68,12 @@ import { Direction } from '../state';
...
  */
-function nextNonce(balanceProof?: SignedBalanceProof): UInt<8> {
-  return (balanceProof?.nonce ?? Zero).add(1) as UInt<8>;
+function nextNonce(balanceProof: BalanceProof): UInt<8> {
+  return balanceProof.nonce.add(1) as UInt<8>;
 }
 
-function getChannelLocksroot(channel: Channel, secrethash: Hash): Hash {
-  const locks: Lock[] = (channel.own.locks ?? []).filter((l) => l.secrethash !== secrethash);
+function withoutLock(end: ChannelEnd, secrethash: Hash): Hash {
```
kelsos, line 75: Any reason for the change of the name? At first glance it's not obvious to me what the function does if I don't see the implementation.

raiden-ts/src/transfers/reducer.ts
```diff
@@ -73,22 +68,18 @@ function transferSignedReducer(state: RaidenState, action: transferSigned): Raid
...
     transfer.locksroot !== locksroot ||
     // nonce must be next
-    !transfer.nonce.eq(
-      (channel[end].balanceProof ? channel[end].balanceProof!.nonce : Zero).add(1),
-    ) ||
-    !transfer.transferred_amount.eq(
-      channel[end].balanceProof ? channel[end].balanceProof!.transferredAmount : Zero,
-    ) ||
-    !transfer.locked_amount.eq((channel[end].balanceProof?.lockedAmount ?? Zero).add(lock.amount))
+    !transfer.nonce.eq(channel[end].balanceProof.nonce.add(1)) ||
```
kelsos, line 80: nitpick: again it would be nice if we could extract the `channel[end].balanceProof` to a variable

kelsos:
Also, I think it would make sense to add a changelog entry for the schema change/version change. It shouldn't affect users, but it might so it would be nice to mention it.

raiden-ts/src/channels/epics.ts
andrevmatos: I'll do it here, but they're `const`/`readonly` properties which [now] are always defined, so I don't see much advantage in extracting them. It's not like accessing a deep, unrelated object. These members could as well all be on the root of the object (which would save the deep access), but are split/nested just for deduplication/code organization

raiden-ts/src/channels/reducer.ts
andrevmatos: Similar, but differ on some things. I wanted to deduplicate things, and be more certain about some things, but they're related, yes.

raiden-ts/src/channels/reducer.ts
andrevmatos: It's already in the first place this kind of reducer is mentioned: root `raidenReducer`. `reduce-reducers` is a known utility on the redux ecosystem, I just borrowed the idea, but will write a comment to make the information more easily accessible

raiden-ts/src/migration/3.ts
andrevmatos: We don't backup because it's an in-memory object. If it fails, Raiden initialization fails and throws without touching storage. After this batch of migrations are performed successfully, we yet proceed to do a full in-depth validation with `RaidenState.decode`, and only then this migrated state is used and overwrites the original data on local storage.

raiden-ts/src/migration/3.ts
andrevmatos: Right, settled channels were `unset` and forgotten. It's not used nor I expect to, but it may be convenient to not lose anything and save for later if needed. Channel state is not big

raiden-ts/src/raiden.ts
andrevmatos: Yes, exactly. This PR is about simplifying state schema, which were a little cluttered without need. See the linked issue description for arguments.

raiden-ts/src/services/epics.ts
andrevmatos: Exactly.

raiden-ts/src/transfers/epics/locked.ts
andrevmatos: It was used because of the optional `locks`, but also was only useful for our `own` end (because previously we didn't receive, so there was no need to calculate partner's locks without a specific lock). I expanded it to be used for both ends. The name was not clear before, `getChannelLocksroot` looks like `getLocksroot`, which just calculates the hash of the locks. Now, `withoutLock(channel.partner, secrethash)` will return the locksroot of `channel.partner` *without lock* `secrethash`.

andrevmatos:
Sure, I'll do, this PR just grew a little because it's just refactoring, it was mostly automated tasks, I didn't expect an in depth review because it changes very little of the previous behavior, just move things around, and for this reason it came in a single PR.

raiden-ts/src/transfers/epics/locked.ts
andrevmatos: It could be `getEndLocksrootWithoutLock`, but I figured out we didn't need this much for an internal helper function :smile:

raiden-ts/src/channels/epics.ts
```diff
@@ -767,14 +767,15 @@ export const channelCloseEpic = (
         return of(channelClose.failure(error, action.meta));
       }
 
+      const balanceProof = channel.partner.balanceProof;
```
kelsos, line 770: My reason for this comment mostly is that we have unnecessary repetition, plus it's easier for the eye to focus without the repetition. thank you.

PR #1463, opened 2020-05-05
Retry without stored setup if auth fails with config.matrixServer

### LINKED ISSUE #1456: Can't connect at https://lightclient.raiden.network/
## Steps to Reproduce
1. click connect
2. it fails

Mozilla/Ubuntu ... I have been using https://lightclient.raiden.network/ previously. Might have something to do with cache although I tried killing the cache.

I think this is what kills it:
image

Requests go to https://raidentransport.test001:
image

Responses:
image

I can ping:
image

If you need more info I will try and provide. This is all I thought you might need :)

andrevmatos:
Fixes #1456

**Short description**
Even if `config.matrixServer` is passed (currently done in dApp in testenv001), and stored `server` setup (auth) matches passed config, append a 2nd server option with same server but without setup to retry auth if stored one fails.

1. Try connecting on testenv setup in a dApp with an auth error

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -10,6 +10,11 @@
 [#1368]: https://github.com/raiden-network/light-client/issues/1368
 [#1252]: https://github.com/raiden-network/light-client/issues/1252
 
+### Fixed
+- [#1456] Retry without stored setup if auth fails
+
+[#1456]: https://github.com/raiden-network/light-client/issues/1456
+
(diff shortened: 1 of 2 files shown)


andrevmatos:
The linked issue contains a very specific state, which I don't understand how it happened, where stored auth info would cause http errors. This PR solves it by ensuring even with specified `config.matrixServer`, the auth is retried once without stored setup if it fails, which will re-login and fix that issue.
The test steps are as descriptive as I could: if you see that issue, test it there.

Issue #1369, opened 2020-04-20
I want to set a monitoring reward calculated based on my channel capacity


christianbrb:
## User Story
As Chris I want to set a monitoring reward calculated based on my channel capacity so that paying the monitoring service will be economically viable.

Split from #1352 

## Acceptance criteria
Decide on either (or both):
- Have a fixed reward, send only monitoring requests when the received amount of tokens are valued more than the amount being paid.
- Set a max threshold, set monitoring requests with smaller values, and hope a monitoring service will pick it up (you will never pay to much).
- Value proportion between a token and a monitoring token.

christianbrb:
@karlb How do you calculate the `reward_amount` the Raiden Node is submitting via the [Monitor Request](https://raiden-network-specification.readthedocs.io/en/latest/monitoring_service.html#monitor-request)?

karlb:
> @karlb How do you calculate the `reward_amount` the Raiden Node is submitting via the [Monitor Request](https://raiden-network-specification.readthedocs.io/en/latest/monitoring_service.html#monitor-request)?

Currently, the python client uses
```
MONITORING_REWARD = TokenAmount(5 * 10 ** 18)  # about 1$
```

There is no way to negotiate sensible rewards at the moment. So the current approach is "set it to a reasonably high value and trust that at least one MS will act". This is definitively something to be improved in the future. Limiting the reward to the user's channel balance makes sense, but will require conversion rates between RDN and the token used in the channel.

christianbrb:
@karlb Thanks for the information :)

christianbrb:
@andrevmatos @kelsos @nephix 
Whoever starts to implement this. Let me know, when you are starting to have a discussion about the solution. I would like to keep the complexity and effort of this issue as low as possible.

christianbrb:
Configuration parameter
- Multiplier = Token / SVT ~ Exchange rate

PR #1480, opened 2020-05-08
Update profile's caps on config.caps change and react on peers updates

andrevmatos:
Part of #1369

**Short description**
So far, caps were reflected on `avatar_url` on Matrix only on initial (startup's) config, and also peers would not have our presence's caps reflected upon avatar's change.
This PR:
- Updates our profile's avatar_url when `config.caps` is updated
- avatar_url's updates are monitored by peers and a new matrix presence update is emitted reflecting these new capabilities

This is a non-breaking change, which just improves on a feature only useful between light-clients, and fully backwards compatible

**Definition of Done**

1. Have two light clients monitoring each other's presence (e.g. by calling `raiden.getAvailability` with other's address)
2. On one of them, update caps with e.g `raiden.updateConfig({ caps: { noDelivery: true, test: true } })
3. See the other detect it and emit a new `matrix/presence/success` action with updated `action.payload.caps` object

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -1,7 +1,12 @@
 # Changelog
 
-## [0.7.0] - 2020-05-08
+## [Unreleased]
+### Changed
+- [#1480] Update profile's caps on config.caps change and react on peers updates
 
+[#1480]: https://github.com/raiden-network/light-client/pull/1480
+
+## [0.7.0] - 2020-05-08
 ### Added
 - [#1392] Raiden on-chain methods provide easy ways to transfer entire token & ETH balances
 - [#1368] Monitoring transfers (experimental)
@@ -174,7 +179,7 @@
 - Add protocol message implementation.
 
 
-[unreleased]: https://github.com/raiden-network/light-client/compare/v0.7.0...HEAD
+[Unreleased]: https://github.com/raiden-network/light-client/compare/v0.7.0...HEAD
 [0.7.0]: https://github.com/raiden-network/light-client/compare/v0.6.0...v0.7.0
 [0.6.0]: https://github.com/raiden-network/light-client/compare/v0.5.2...v0.6.0
 [0.5.2]: https://github.com/raiden-network/light-client/compare/v0.5.1...v0.5.2
--- a/raiden-ts/src/transport/epics/init.ts
+++ b/raiden-ts/src/transport/epics/init.ts
@@ -281,10 +281,9 @@ function setupMatrixClient$(
   }).pipe(
     // the APIs below are authenticated, and therefore also act as validator
     mergeMap(({ matrix, server, setup }) =>
-      // ensure displayName is set even on restarts
+      // set these properties before starting sync
       merge(
         from(matrix.setDisplayName(setup.displayName)),
-        from(matrix.setPresence({ presence: 'online', status_msg: '' })),
         caps ? from(matrix.setAvatarUrl(stringifyCaps(caps))) : EMPTY,
       ).pipe(
         mapTo({ matrix, server, setup }), // return triplet again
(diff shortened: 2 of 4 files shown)


raiden-ts/src/transport/epics/presence.ts
```diff
@@ -262,3 +250,28 @@ export const matrixMonitorChannelPresenceEpic = (
...
+): Observable<never> =>
+  config$.pipe(
+    pluckDistinct('caps'),
+    skip(1), // skip replay(1) and act only on changes
+    mergeMap((caps) => matrix$.pipe(map((matrix) => [caps, matrix] as const))),
+    mergeMap(async ([caps, matrix]) =>
+      matrix.setAvatarUrl(caps && !isEmpty(caps) ? stringifyCaps(caps) : '').catch(() => {
+        /* ignore http errors */
+      }),
+    ),
```
nephix, line 275: Possible unnecessary `async` here if we don't await anything

nephix:
During tests, step #3 failed:
> See the other detect it and emit a new matrix/presence/success action with updated action.payload.caps object

The other client didn't fire a new `matrix/presence/success`

raiden-ts/CHANGELOG.md
```diff
@@ -19,8 +19,10 @@
 
 ### Changed
 - [#1462] Refactor state schema and types to be simpler and safer
+- [#1480] Update profile's caps on config.caps change and react on peers updates
 
 [#1462]: https://github.com/raiden-network/light-client/issues/1462
+[#1480]: https://github.com/raiden-network/light-client/pull/1480
```
nephix, line 25: We made last weeks release today, so this would need to go into a new unreleased section

raiden-ts/src/transport/epics/init.ts
```diff
@@ -390,10 +389,10 @@ export const matrixShutdownEpic = (
     mergeMap((matrix) =>
       action$.pipe(
         finalize(() => {
+          matrix.stopClient();
           matrix.setPresence({ presence: 'offline', status_msg: '' }).catch(() => {
```
nephix, line 393: Does it make sense to stop the client and then send a presence update?

Also, could this cause unresolved promises in e.g. our tests? Looks like `setPresence` returns a promise that we don't await for so it could be pending for quite a bit

raiden-ts/CHANGELOG.md
andrevmatos: Yes, this PR was done before the release

raiden-ts/src/transport/epics/presence.ts
andrevmatos: Some plugin on my editor suggested it when I had a function with `return`, I think even without `await`, it prefers functions returning promises to be marked as `async`

raiden-ts/src/transport/epics/init.ts
andrevmatos: `stopClient` stops the `sync` loop which makes us appear online. stopping it after marking the offline status could in rare race conditions make us appear online for another 30s (next loop), therefore I stop it before the `setPresence`. The `setPresence` call is a simple http endpoint, used just to ensure we're marked earlier as offline (usually, we'd appear as online until server doesn't see sync requests for another 30s, which could make peers keep retrying something with us while we're not listening anymore). This shouldn't cause unresolved promises in our tests, specially because we mock matrix and this endpoint should resolve immediately, while tests only complain if some promise isn't resolved 1s after tests end. We can't await anyway here, since it's a `finalize` call, performed in the subscription's teardown/shutdown/sync stop, so it's a 'best effort' call.

nephix: We don't mock matrix in integration tests

andrevmatos: Yes, I know, but there it's a local server which should reply on less than 1s. Anyway, this doesn't change the previous behavior, which did work fine, just move the call a little to prevent an hypothetical race on offline/sync loop which wouldn't cause any major issue than a small delay on clients on detecting our status.

nephix:
Test worked now on my side, forgot to `getAvailability()` in both directions instead of just one

Issue #1373, opened 2020-04-20
Update dApp connect/onboarding flow


christianbrb:
## User Story
As Chris I want to see which reward I would pay if a monitoring service submits my balance proof so that I am able to trust the system.

Split from #1352 

## User flow
1. Select Hub (do not show mint & deposit any more)
2. Open Channel (rename to "Set deposit")
3. Monitor Channel (new screen)
3a. Show available SVT tokens (move #1371 to this new screen)
3b. Mint & Deposit button
3c. Show the monitoring reward in SVT
3d. Show the exchange rate (preset but adjustable) -> related SDK feature #1369 
3e. Show the monitoring reward in the channel token = exchange rate * monitoring reward
3f. Explain the monitoring reward briefly
3g. Button should be called "Open Channel" -> not sure if we should call the screen the same

## Acceptance criteria
- Just display the value, how many RDN if the monitoring reward is used. This is not interactive so showing the value is about being transparent about it.
- There is a default for each token we use separated by net
- There is an explanation of how the user can calculate the value

## Tasks
- [ ] Design [Pre-iteration]

christianbrb:
@andrevmatos Proposal: 
- Modal, if we receive a transfer before signing the message for the MS
- Add it to the transfer history

christianbrb:
Please assign the "Ready" label after the design has been done, thanks :)

taleldayekh:
Sure thing!

christianbrb:
@taleldayekh I have written down the user flow (see above). Let me know, if you have got questions.

taleldayekh:
The way I understood this is that we want to be transparent about the amount of SVT/RDN a user might have to spend if a balance proof is sent.

Since this is not interactive (at the moment) I suggest we just add this information on the UDC screen. A user would then see how many SVT/RDN tokens they have and in relation to that also how many of those tokens they need to pay in case a balance proof is sent?
@kelsos @andrevmatos @christianbrb @weilbith

christianbrb:
The original idea was to put it in the token connection flow (see user flow). There it would be more educative / explaining for the user.

taleldayekh:
Alright, let me sit down and continue thinking about this one. I think the issue is a lot bigger than the title suggests.

taleldayekh:
![flowchart_connect](https://user-images.githubusercontent.com/43838780/83246988-2aeae600-a1a3-11ea-9f6a-69bcd98c20c4.jpg)

[...]

PR #1494, opened 2020-05-12
Channel balance-based Monitoring threshold

### LINKED ISSUE #1369: I want to set a monitoring reward calculated based on my channel capacity

andrevmatos:
Fixes #1369 

**Short description**
Implements a new config value for the SDK: `rateToSvt`, which holds a `token => wei` mapping. Only tokens present on this mapping will have monitoring enabled, and monitoring will be requested (i.e. paid to MS as rewards) only after channel's received total (off and on-chain unlocked sum), normalized through this rate, surpasses `monitoringReward` config, which default now is set to `5 SVT` (same as raiden-py,  hopefully enough for a MS to pick up our request, may be tweaked if needed). The `rateToSvt` config is empty by default, meaning receiving is still disabled by default and therefore no monitoring is applied.

**Definition of Done**

1. In served dApp, set the rate:
```ts
// rate=5e18 means 1e18wei of TKN (1TKN if decimals=18) is worth 5SVT, so 1+TKN is needed to trigger monitoring
raiden.updateConfig({ rateToSvt: { "<token_addr>": "5000000000000000000" } });
```
2. Receiving gets enabled
3. In another dApp account, transfer 0.9TKN to first client. Transfer should be received, but no monitoring signatures are requested
4. Make new transfer, worth 0.11TKN; threshold cross value of default monitoringReward of 5SVT, monitoring prompt happens 15s later and is sent to global rooms: `messageGlobalSend` actions to `meta: { roomName: "raiden_goerli_monitoring" }` containing the `type: "RequestMonitoring"` messages.

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 ## [Unreleased]
+### Added
+- [#1369] Monitoring based on channel's balance
+
+[#1369]: https://github.com/raiden-network/light-client/issues/1369
+
 ### Changed
 - [#1480] Update profile's caps on config.caps change and react on peers updates
 
--- a/raiden-ts/src/channels/reducer.ts
+++ b/raiden-ts/src/channels/reducer.ts
@@ -6,6 +6,8 @@ import { partialCombineReducers } from '../utils/redux';
 import { RaidenState, initialState } from '../state';
 import { SignatureZero } from '../constants';
 import { RaidenAction, ConfirmableActions } from '../actions';
+import { transferSecretRegister } from '../transfers/actions';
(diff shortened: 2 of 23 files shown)


christianbrb:
@kelsos Yes, it is going to be the following issue https://github.com/raiden-network/light-client/issues/1373

PR #1504, opened 2020-05-12
Fix received transfers not showing up on history

### LINKED ISSUE #1503: Received transfers aren't shown on transfers history
This actually isn't a dApp issue, but after receiving got merged, as it wasn't enabled by default anywhere, I forgot to expose the received transfers through `Raiden.transfers$` observable. Also, a proper transfer id/unique key is needed, since outgoing transfers can share secrethash with incoming ones (e.g. on mediation), so a `transferKey` is also needed, which decouples usage from transfer's internals like `secrethash`

## Steps to Reproduce
1. Enable transfers receiving after #1494 
2. Perform a transfer from another dApp account

## Expected Result
- Transfer is shown as a received transfer on history

## Actual Result
- No received transfer show up

Current master

andrevmatos:
Fixes #1503 

**Short description**
Exposes received transfers through `transfers$` observable, define `RaidenTransfer.key` as a unique string key, to be used instead of `secrethash`.

**Definition of Done**

1. Try repro at linked issue
2. See it works for received transfers:
image

### DIFF
--- a/raiden-ts/src/transfers/utils.ts
+++ b/raiden-ts/src/transfers/utils.ts
@@ -1,10 +1,10 @@
 import { concat, hexlify } from 'ethers/utils/bytes';
 import { keccak256, randomBytes, bigNumberify, sha256 } from 'ethers/utils';
 
-import { Hash, Secret, UInt, HexString } from '../utils/types';
+import { Hash, Secret, UInt, HexString, assert } from '../utils/types';
 import { encode } from '../utils/data';
 import { Lock } from '../channels/types';
-import { TransferState, RaidenTransfer, RaidenTransferStatus } from './state';
+import { TransferState, RaidenTransfer, RaidenTransferStatus, Direction } from './state';
 
 /**
  * Get the locksroot of a given array of pending locks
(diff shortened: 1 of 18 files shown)


andrevmatos:
No need for dApp changes changelog, since the issues there didn't affect users since receiving isn't enabled yet.

raiden-ts/src/transfers/utils.ts
```diff
@@ -59,12 +59,50 @@ export function makeMessageId(): UInt<8> {
...
+  return transfer.recipient === state.partner ? Direction.SENT : Direction.RECEIVED;
+}
+
+/**
+ * Get a unique key for a tranfer state or TransferId
+ *
+ * @param state - transfer to get key from, or TransferId
+ * @returns string containing a unique key for transfer
+ */
+export function transferKey(
```
kelsos, line 80: Just out of curiocity what was the problem that this key solves?

raiden-ts/src/utils/rx.ts
```diff
@@ -55,3 +56,33 @@ export function pluckDistinct<T, R>(...properties: string[]): OperatorFunction<T
...
 }
+
+/**
+ * Creates an operator to output changed values unique by key ([key, value] tuples)
+ * It's equivalent to (from(Object.entries) + distinct), but uses key to prevent memory leak
+ *
+ * @param compareFn - Function to compare equality between two values, default to === (reference)
+ * @returns Operator to map from a record to changed values (all on first)
+ */
+export function distinctRecordValues<R>(
```
kelsos, line 67: :+1:

raiden-ts/src/transfers/utils.ts
andrevmatos: Same as `channelKey`, hardcoding `secrethash` didn't account for duplications, in transfer's case, incoming and outgoing transfers could have the same secrethash (i.e. mediated transfers), and also secrethash only as a key didn't tell if it was an incoming our outgoing transfer, which was done at a time we only had outgoing transfers.
This _standardizes_ the "uniqueness" of key through this function, and if needed, we can update it in the future, to include other parameters for transfer uniqueness.

kelsos:
Like we discussed codewise looks good :) I still need to check it locally but it seems that everything started misbehaving locally :'(

PR #1538, opened 2020-05-14
[dont merge] Added lerna

nephix:
If someone wants to try/play around

```
$ git clone --recurse-submodules git@github.com:nephix/light-client.git
$ git checkout test/lerna
$ npx lerna bootstrap
$ cd raiden-dapp
$ npm run serve
```

I didn't adjust anything related to tests or CI
So it will probably fail because it doesn't run `lerna bootstrap`

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,0 +1,6 @@
+{
+  "packages": [
+    "raiden-cli", "raiden-dapp", "raiden-ts"
+  ],
+  "version": "0.0.0"
+}
--- a/package.json
+++ b/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "root",
+  "private": true,
+  "devDependencies": {
+    "lerna": "^3.21.0"
+  }
+}
--- a/raiden-cli/package-lock.json
+++ b/raiden-cli/package-lock.json
@@ -4533,9 +4533,6 @@
       "integrity": "sha1-nsYfeQSYdXB9aUFFlv2Qek1xHnM=",
       "dev": true
     },
-    "raiden-ts": {
-      "version": "file:../raiden-ts"
-    },
     "randombytes": {
       "version": "2.1.0",
       "resolved": "https://registry.npmjs.org/randombytes/-/randombytes-2.1.0.tgz",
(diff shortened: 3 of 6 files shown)


lerna.json
```diff
@@ -0,0 +1,6 @@
+{
+  "packages": [
+    "raiden-cli", "raiden-dapp", "raiden-ts"
+  ],
+  "version": "0.0.0"
```
andrevmatos, line 5: Looks like lerna can be used to bump all packages version simultaneously? Do we need to keepthis version here in sync with current version to use that on prepare_release?

nephix: Yeah it can do that, we would need to play around and investigate how/if we want to use that

nephix: It can be used for independent versions as well. If all versions are synced it seems to be called [fixed/locked mode](https://github.com/lerna/lerna#fixedlocked-mode-default)

kelsos:
I am guessing we still need some changes in the ci since `npm ci` wont' work anymore for the dApp right?

nephix:
> I am guessing we still need some changes in the ci since `npm ci` wont' work anymore for the dApp right?

Yes I haven't touched testing/integration image/CI
I thought first we could check if that's a route we want to take before we dig into the details

> I am guessing for that we have to integrate the boostrap part in the ci flow right?

Yup `npx lerna bootstrap` instead of `npm ci`

kelsos:
It seems to work good for us, and not many changes are needed to go this path. So I am certainly looking at it positively.

nephix:
We should not use `lerna publish` because it assumes that all packages are published to NPM..

So in terms of releasing `raiden-ts`, we would stick with `npm publish`

kelsos:
Yes it seems absolutely reasonable to keep publishing like before.

@andrevmatos what do you think? from my perspective it makes sense to move forward with it.

andrevmatos:
Yes, I agree, since it'll be mostly non-disruptive. IMO, it's a bit overkill to include this new dependency and workflow just because of the `raiden-ts` linking, but it's valid, so if you guys want to go ahead, I'm not against it.

kelsos:
Ok so let's finalized it. like you said it's not disruptive and I like that you can install everything with a single command.

nephix:
I friend told me yesterday that they use https://pnpm.js.org instead of lerna for their monorepo, I have no experience with it though but might be worth considering as well if lerna is perceived as overkill

kelsos:
I don't mind either, we just need a solution to make our life easier :)

PR #1542, opened 2020-05-15
Removed fixed height from amount input error label & transfer section

### LINKED ISSUE #1506: Error text below amount input field is not fully shown
## Steps to Reproduce
1. Navigate to the transfer screen
2. Enter an amount higher than your token network capacity in the amount input field

## Expected Result
- Full text is shown

## Actual Result
- Text is not fully visible
  
grafik

```
- Browser & version
- Browser console errors
- Device / OS
```

nephix:

Fixes #1506 

**Short description**

**Definition of Done**

1. Go to transfer screen
2. Insert amount that is larger than channel capacity


raiden-dapp/src/components/navigation/Transfer.vue
```diff
@@ -296,9 +296,10 @@ export default class Transfer extends Mixins(BlockieMixin, NavigationMixin) {
     &__form {
       background-color: $transfer-form-color;
       border-radius: 15px;
-      height: 175px;
```
kelsos, line 299: Would it be easy to increase the fixed size to cover two lines of text instead?
The reason it was originally fixed was to avoid having the UI jump every time extra vertical space was needed.

nephix: The error message in that case can be quite long do you think we should reserve the blank space for it or leave it with the jump?

Screenshot 2020-05-15 at 14 52 00

I feel like both is not ideal 😓

kelsos: Indeed also we should make sure that the message never goes to more than two lines. but it is hard to ensure. maybe we can keep it with jumping initially and discuss an alternative solution. @taleldayekh maybe adding an error indicator in the input like an error icon that gives you the erron on tooltip?

nephix:
Decision in daily: We will keep the version that occupies the space for now and I will create a follow up issue to rethink how we can handle input errors

PR #1543, opened 2020-05-15
Fixes handling of tokens with zero decimals

### LINKED ISSUE #1493: Error message on deposit field in open channel is wrong
## Steps to Reproduce
1. Select a token with no balance
2. Select hub
3. Open channel
4. Enter an amount larger than 0

## Expected Result
- Error message telling me, that I don't have tokens to deposit

## Actual Result
- Message `Please enter an amount larger than zero` is shown
   
   grafik

kelsos:

Fixes #1493 

**Short description**
Properly handles the display of tokens that contain 0 decimals. Previously 0 would fall into the same category as `undefined` and `18` decimals would be assumed which was wrong.

- Also fixes the token display on the transfer history

**Definition of Done**

1. Try to create a new channel with one of the scenario player tokens. 
2. For the scenario player tokens the amounts should be integer everywhere. 
3. The amount input for the scenario player should inform you if you have any decimals, and should consider valid any integer input up to `max`. I
4. Switching between `TTT` and the scenario token should switch between displaying decimals and not displaying.
5. Change between 0 and 0.00 as a place holder should also appear in the deposit screen.
6. When opening a channel on the scenario player token, you get a message about 0 decimals if you input any decimals.

### DIFF
--- a/raiden-dapp/src/store/index.ts
+++ b/raiden-dapp/src/store/index.ts
@@ -64,7 +64,7 @@ const hasNonZeroBalance = (a: Token, b: Token) =>
 const settingsLocalStorage = new VuexPersistence<RootState>({
   storage: window.localStorage,
   reducer: state => ({ settings: state.settings }),
-  filter: mutation => mutation.type == 'updateSettings',
+  filter: mutation => mutation.type == 'updateSettinSyugs',
   key: 'raiden_dapp_settings'
 });
 
--- a/raiden-dapp/src/store/index.ts
+++ b/raiden-dapp/src/store/index.ts
@@ -64,7 +64,7 @@ const hasNonZeroBalance = (a: Token, b: Token) =>
 const settingsLocalStorage = new VuexPersistence<RootState>({
   storage: window.localStorage,
   reducer: state => ({ settings: state.settings }),
-  filter: mutation => mutation.type == 'updateSettinSyugs',
+  filter: mutation => mutation.type == 'updateSettings',
   key: 'raiden_dapp_settings'
 });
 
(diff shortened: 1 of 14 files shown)


raiden-dapp/src/utils/balance-utils.ts
```diff
@@ -20,6 +24,18 @@ export class BalanceUtils {
...
+      deposit.endsWith('.')
+        ? deposit.substring(0, deposit.length - 1)
+        : deposit,
+      decimals
+    );
+  }
+}
+
+export function getDecimals(decimals?: number) {
+  if (decimals === undefined) {
```
andrevmatos, line 37: No need to change it, but this could be `decimals ?? 18`, and if it'd be this simple, maybe there'd be no need for a function

kelsos: That is a nice suggestion, I didn't think of that to be honest

raiden-dapp/src/store/index.ts
```diff
@@ -64,7 +64,7 @@ const hasNonZeroBalance = (a: Token, b: Token) =>
 const settingsLocalStorage = new VuexPersistence<RootState>({
   storage: window.localStorage,
   reducer: state => ({ settings: state.settings }),
-  filter: mutation => mutation.type == 'updateSettings',
+  filter: mutation => mutation.type == 'updateSettinSyugs',
```
andrevmatos, line 67: Did you just drop a `pacman -Syu` on your mutation name? :joy:

kelsos: OK now it makes sense why the settings where resetting :rofl:

Issue #1421, opened 2020-04-24
I want to withdraw tokens from the UDC contract


christianbrb:
## User Story
As Chris I want to withdraw tokens from the UDC contract so that I can sell them.

## Acceptance criteria
- Needs the notification area to be in place
- There is an event which helps
- There is a withdraw and deposit button (perhaps in the UDC screen)

## Tasks
- [ ] Design (Pre-Iteration design)

christianbrb:
@karlb How do I withdraw my tokens from the UDC as a user? I have already checked https://raiden-network-specification.readthedocs.io/en/latest/service_contracts.html#userdeposit, but couldn't find an answer.

karlb:
To withdraw, you have to call `planWithdraw(amount)`, wait 100 blocks and call `withdraw(amount)`. You can do this on etherscan. The Raiden client should provide a user friendly way to do that, but it currenlty doesn't. It's only [getting the capability to do that internally](https://github.com/raiden-network/raiden/pull/6091) just now.

christianbrb:
@karlb Cool, thanks for the information :) We are planning to implement this in the Light Client.

christianbrb:
@taleldayekh Please assign the "Ready" label after the design has been done, thanks :)

taleldayekh:
Sure thing!

taleldayekh:
One quick fix which also makes sense would be to add the withdrawal icon next to the balance. A user could then perform minting/depositing and withdrawal on the same screen.

<img width="591" alt="Screenshot 2020-06-02 at 10 01 59" src="https://user-images.githubusercontent.com/43838780/83495784-727db480-a4b8-11ea-962b-e45459fb0896.png">

kelsos:
Sound like a good idea to include a withdraw button there

taleldayekh:
Mockup ready. Pretty straight forward but it does the trick:
https://xd.adobe.com/view/6d1e5b87-796c-43d3-6dc0-4817716010fe-1b2f/

Not that the withdrawal icon should not be visible if there is no balance.
@kelsos @andrevmatos @weilbith

taleldayekh:
Whoever picks this issue please also center the top text. Dunno why it is not centered as it is now.

kelsos:
@taleldayekh do we have this icon already in the project?

taleldayekh:
Sure, you should already have used it when you implemented the token withdrawal screen.

Issue #357, opened 2019-09-27
Expose multi-step operations as observables


andrevmatos:
## Description

Currently, the `Raiden` public class/API exposes all its methods as async methods/Promises. This is ok for one-shot async operations (like `getTokenBalance`), but not enough for methods which have multiple steps it may be useful to expose to the user.
Example is `depositChannel` method, which consists in at least 2 steps (approve + setTotalDeposit, possibly more, like mined/confirmed pattern), and more than one of those requires user interaction (tx prompts), which makes it very important ot expose this progression to the user, so an UI can reflect the current state of the operation (by showing the user what they're about to sign).
For these operations which may be relevant to inform the user about task progression, we should return a hot observable instead of a promise, which emits descriptive elements about each step, and completes in the end or errors on failure. Getting back the current behavior then is as simple as calling `.toPromise()` on the observable.
`transfer` and most on-chain methods also may make good use of this.

## Acceptance criteria
- Any relevant public method returns a hot Observable, the simplest ones are still Promise based, as both patterns complement each other and play together

## Tasks
- [ ] Design a good way to notify progress. enums? raw actions? counters/steppers?

nephix:
> we should return a hot observable instead of a promise

Might make sense to consider giving users of the SDK an alternative to RxJS observables, by simply allowing them to provide a callback function. This way we keep it unopinionated and don't force users to implement RxJS, although it has a lot of benefits.

An example of how this could be done can be seen in Google Firebase's [onAuthStateChanged](https://firebase.google.com/docs/reference/js/firebase.auth.Auth.html#onauthstatechanged) method

christianbrb:
- We know better in which step the open channel stepper is
- Hardest part is to decide what approach to take  (technology decision)

kelsos:
Last week we had a long discussion with @nephix regarding this issue, so I would like to present some of my arguments on this matter. 

The biggest problem at the moment is that the API is half-reactive, half promise-based. The functions are exposed as promises but the `events` etc are exposed as observables. I believe that exposing a fully reactive API should be easy for us to do, and it should not require a lot of extra effort to support while bringing all the benefits reactive programming provides.

However, I can also see the points @nephix raised. I have a reactive extensions background (.NET, Android, JavaScript) and using an API based on observables feels familiar. I can understand however that learning and understanding reactive programming is not easy and forcing the users to understand reactive programming if they want to use the SDK doesn't seem right.
 
[...]

PR #1548, opened 2020-05-19
Notify when receiving transfers gets disabled

### LINKED ISSUE #1473: I want to get receiving deactivated if the SVT in the UDC is not high enough to monitor my transfers
## Description
As Chris, I want to get receiving deactivated if the SVT in the UDC is not high enough to monitor my transfers so that I don't run into a risk of losing tokens.

## Acceptance criteria
- If the SVT in the UDC is not high enough I cannot receive any more transfers until I have minted and deposited more SVT in the UDC.
- Receiving is enabled by default
- The SDK informs if receiving is enabled or disabled (see https://github.com/raiden-network/light-client/issues/1473#issuecomment-627350833)
- The user sees a warning

## User flow
1. The SDK is monitoring if the SVT in the UDC is high enough
2. If the SVT is not high enough 
2a.  Receiving will be deactivated
2b. The user gets a warning, that receiving is deactivated and he needs to deposit tokens to the UDC
udc-warning

## Design
- Modal popup, with the warning

## Monitoring Reward
- `raiden.config.monitoringReward: BigNumber | null = null`
- `raiden.updateConfig({ monitoringReward: parseUnits('5', 18) })`

## Tasks
- [ ] Design (before iteration starts)

andrevmatos:
Fixes #1473 

TODO: tests, changelog

**Short description**
A dialog is shown on any screen on dApp when/if SDK has Receiving transfers disabled. This can happen on startup or during runtime, for mainly a couple of reasons:
- `rateToSvt` config mapping is empty (it is by default, so receiving is disabled until it gets populated)
- `monitoringReward` config isn't set (it is by default, as 5SVT)
- `udcBalance` is not enough to pay monitoring reward (which can happen at startup or at runtime by a withdraw from service or user #1421).
- explictly disabled in user config with `Raiden.updateConfig({ caps: { noReceive: true } })`, which has priority

image

**Definition of Done**

1. On dApp, acquire `raiden` SDK instance and enable receiving with:
```ts
raiden.updateConfig({
  rateToSvt: { '<token_addr>': '100000000000000000000' },
  caps: { noReceive: false },
 })
```
2. Reload, check no dialog is shown
3. Disable receiving explicity while connected:
```ts
raiden.updateConfig({ caps: { noReceive: true } })
```
4. See dialog shown which says receiving is disabled
5. One can also trigger dialog shown, even if `caps['noReceive']` is unset (auto), if UDC balance is emptied, if one sets `monitoringReward` config to something higher than UDC desposit or `rateToSvt` is emptied

### DIFF
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -79,7 +79,6 @@ import { RaidenError, ErrorCodes } from './utils/error';
 export class Raiden {
   private readonly store: Store<RaidenState, RaidenAction>;
   private readonly deps: RaidenEpicDeps;
-  public config!: RaidenConfig;
 
   /**
    * action$ exposes the internal events pipeline. It's intended for debugging, and its interface
@@ -110,6 +109,11 @@ export class Raiden {
    * may be used as identifier to know which transfer got updated.
    */
   public readonly transfers$: Observable<RaidenTransfer>;
+
+  /** RaidenConfig object */
+  public config!: RaidenConfig;
+  /** RaidenConfig observable (for reactive use)  */
+  public config$: Observable<RaidenConfig>;
 
   /**
    * Expose ether's Provider.resolveName for ENS support
@@ -241,9 +245,10 @@ export class Raiden {
       },
     });
 
-    this.deps.config$.subscribe((config) => (this.config = config));
+    this.config$ = this.deps.config$;
+    this.config$.subscribe((config) => (this.config = config));
 
-    this.deps.config$
+    this.config$
       .pipe(pluckDistinct('logger'))
       .subscribe((logger) => this.log.setLevel(logger || 'silent', false));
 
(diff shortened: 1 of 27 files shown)


raiden-ts/src/raiden.ts
```diff
@@ -111,6 +110,11 @@ export class Raiden {
    */
   public readonly transfers$: Observable<RaidenTransfer>;
 
+  /** RaidenConfig object */
+  public config!: RaidenConfig;
```
nephix, line 114: Should we provide a callback'ish way as well? I guess we haven't decided yet in #357

andrevmatos: Yes, I was thinking about it, I really want to standardize these interfaces, but didn't want to make this PR more complex, so just used the same pattern we already have for `state$`, `channels$`, `transfers$` and so.. I'd say we should keep it like that for now, but I'm thinking on something like making `Raiden` an `EventEmitter` derived class and emitting these as events in a more interoperable way.

PR #1573, opened 2020-05-22
ETH transfer screen redesign

taleldayekh:

Fixes # 

**Short description**
Implements a new design in accordance with user feedback for the eth transfer in the account screen.

**Definition of Done**

1. Connect to the dApp and click to Account Screen -> Raiden Account
2. Transfer ETH from Main Account to Raiden Account
3. Transfer ETH from Raiden Account to Main Account


raiden-dapp/src/assets/logo_blue.svg
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
```
andrevmatos, line 1: Don't we already have this logo in black? Wouldn't be better if you `css filter` to change color instead of adding a new asset?

taleldayekh: We do, I could not find a way of changing it to that specific blue color via css filter alone. But maybe I can access the SVG itself without any external library?

taleldayekh: Fixed it with filters! :)

PR #1577, opened 2020-05-26
Refactor tests and fixtures

### LINKED ISSUE #1377: One fixture to rule 'em all
## Description

Part of #1268 
Supersedes #871 

The bulk of logic/behavior/protocol tests happens on epics unit tests. Epics can be fixtured and tested by providing:
- `action$: Observable<RaidenAction>` subject, to provide actions input
- `state$: Observable<RaidenState>` subject, to provide states input
- `deps: RaidenEpicDeps` object, containing all other dependencies for epics
  - `latest$: Observable<Latest>` is an important member of `deps`, which is a `ReplaySubject(1)` of a `Latest` object containing the latest emitted value of a few key reactive properties, which can be selected using `pluck` and `pluckDistinct` operators

Currently, these tests fixturing are less than ideal, as it involves too much steps to set up some specific state for each test, and also are done differently per test; usually, something along:
- `raidenEpicDeps` returns a single `depsMock: MockRaidenEpicDeps` object for a constant "we" user, as a mocked `RaidenEpicDeps` object/contracts/signers, etc
- from this object, `epicFixtures` function returns an object containing several _constant_ properties, but which are specific per run, including `action$` & `state$`, which need to be reset `beforeEach` to prevent changes done by one test of affecting others
- More often than not, on top of these values/subjects, several changes need to be applied to put state as expected by some tests; examples:
  - on tests which needs a channel in state, we need to pass through `action$` several actions, like `tokenMonitored` & `channelOpen.success` to put a channel in state
  - partner/target `matrixPresence.success` to have peers seen as online

We should instead refactor these fixtures to a single function, which would return some requested number of clients (usually up to 3, at most4), and ways to specify their state:
- monitored tokens
- channels connecting them, their state & deposits
- availability
- other state which may be needed? TBD

The main output for each client would be the 3 epics arguments, but we may want to have ways to extract common constants out of it, like configs, common addresses (token, token network, etc), matrix server, settle timeout, etc (look at output of `epicFixtures`).
With these vars, we can easily standardize how to test epics:
- create fixture, define which client is us, partner, target, etc
- optionally, set up spies
- call and subscribe to epic
  - instead of calling each epic specifically, we should always call `raidenRootEpic` and test specific epics behavior from there
  - subscription can be done through usual `subscribe` (don't forget to complete subjects and/or unsubscribe in the end) or `toPromise` with or without `toArray` to test either latest or array of all outputs respectivelly
- `next` sensible actions, state, `emit` events on matrix or provider, etc, to trigger the epic behavior we want to test; may or may not need to `action$.complete` right after emits or after a little timeout
- test output is the expected
- optionally test spies

Is this a good test architecture for heavy observables usage/async actions we have? Suggestions?

## Acceptance criteria
- A fixture which is able to set up all epics tests
- All epics tests use the above fixture

## Tasks
- [ ] Create the PR with the fixture and basic examples of how it works
- [ ] Split the PR

andrevmatos:
Fixes #1377 

**Short description**
Refactor unit mocks, fixtures and tests. For now, the new pattern was only applied to `utils.spec.ts` and `channels.spec.ts` (this was one of the oldest tests suites, and therefore, one of the most outdated).

The new testing paradigm involves a `makeRaiden` mocked function, which initializes a mocked Raiden client (actually, just a small subset of its functionality), including full epics subscription and `await` for first block actions to be synced. A `makeRaidens(n)` utility is also provided to generate `n` clients in parallel (including some pure caching, like wallets), which can then be destructured into e.g. `[raiden, partner]` pairs. These functions also queue cleanup functions to stop/unsubscribe from generated observables. For waits, `setTimeout`s for completes can be replaced by `await waitBlock`, which emits the next block and also ensures it's picked by registered clients.

For fixtures, there're now a few `ensure*` async utility functions, which should emit events and actions to the clients passed as parameter and `await` them to catch up a given state. Example: `ensureChannelIsOpen([raiden, partner])` will emit contracts events and only resolve after both us and partner fetch those events and persist them in state. Constant fixtures are also moved to `fixtures` and can be imported directly.

With this, these tests are self-contained, properly cleaned up `afterEach`, hopefully not leaking to other tests in the suite, and after we finish migrating and remove the previous mocks, we can even enable `restoreMocks` jest config option safely.

**Definition of Done**

1. Just run SDK's unit test and ensure everything pass

### DIFF
--- a/raiden-ts/tests/unit/fixtures.ts
+++ b/raiden-ts/tests/unit/fixtures.ts
@@ -4,14 +4,19 @@ import { patchEthersDefineReadOnly } from './patches';
 patchEthersDefineReadOnly();
 
 import { Subject } from 'rxjs';
-import { first, scan } from 'rxjs/operators';
+import { first, scan, pluck } from 'rxjs/operators';
 import { Wallet } from 'ethers';
-import { AddressZero } from 'ethers/constants';
-import { bigNumberify } from 'ethers/utils';
+import { AddressZero, Zero, HashZero } from 'ethers/constants';
+import { bigNumberify, defaultAbiCoder } from 'ethers/utils';
 
 import { Address, Hash, Int, UInt } from 'raiden-ts/utils/types';
 import { Processed, MessageType } from 'raiden-ts/messages/types';
-import { makeMessageId, makePaymentId } from 'raiden-ts/transfers/utils';
+import {
+  makeMessageId,
+  makePaymentId,
+  makeSecret,
+  getSecrethash,
+} from 'raiden-ts/transfers/utils';
 import { IOU } from 'raiden-ts/services/types';
 import { RaidenState } from 'raiden-ts/state';
 import { pluckDistinct } from 'raiden-ts/utils/rx';
@@ -19,8 +24,23 @@ import { RaidenAction } from 'raiden-ts/actions';
 import { raidenReducer } from 'raiden-ts/reducer';
 import { getLatest$ } from 'raiden-ts/epics';
 import { channelKey } from 'raiden-ts/channels/utils';
+import { tokenMonitored } from 'raiden-ts/channels/actions';
+import { ChannelState } from 'raiden-ts/channels';
+import { Direction } from 'raiden-ts/transfers/state';
+import { transfer, transferUnlock } from 'raiden-ts/transfers/actions';
+import { messageReceived } from 'raiden-ts/messages/actions';
 
-import { makeMatrix, MockRaidenEpicDeps } from './mocks';
+import {
+  makeMatrix,
+  MockRaidenEpicDeps,
+  MockedRaiden,
+  makeAddress,
+  makeHash,
+  waitBlock,
+  providersEmit,
+  makeLog,
+} from './mocks';
+import { Filter } from 'ethers/providers';
 
 /**
  * Composes several constants used across epics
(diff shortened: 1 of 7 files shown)


raiden-ts/tests/unit/fixtures.ts
```diff
@@ -125,3 +145,282 @@ export function epicFixtures(depsMock: MockRaidenEpicDeps) {
     state$,
   };
 }
+
+// fixture constants
+export const token = makeAddress();
+export const tokenNetwork = makeAddress();
```
kelsos, line 151: Would it make sense to use a `TOKEN_NETWORK` naming for the constacts?

kelsos: OK I saw below that you use them directly to avoid specifying the property name.

raiden-ts/tests/unit/mocks.ts
```diff
@@ -1,21 +1,32 @@
...
 
-import { AsyncSubject, of, BehaviorSubject, Subject } from 'rxjs';
+import { AsyncSubject, of, BehaviorSubject, ReplaySubject } from 'rxjs';
 import { MatrixClient } from 'matrix-js-sdk';
 import { EventEmitter } from 'events';
 import { memoize } from 'lodash';
 import logging from 'loglevel';
+import { Store, createStore, applyMiddleware } from 'redux';
 
+// TODO: remove this mock
```
kelsos, line 14: Is there some reason blocking us from removing it right now?

andrevmatos: `jest.mock` returns a mocked instance of the classes in the module. We use it on the old `raidenEpicDeps`, and need to keep it there while there're tests using these mocks. You can see the new mocks use `jest.requireActual` to get the actual class, and just mock/spy on certain methods. After we refactor the remaining tests, and can remove the old mocks and revert mocking the whole providers mock.

kelsos: Ok that makes sense, thank you.

PR #1580, opened 2020-05-26
Installs cypress and prepares dApp for integration image

### LINKED ISSUE #1404: Use cypress (frontend UI testing) and integration image for e2e/integration testing off the dApp
## Description
Use cypress (frontend UI testing) and integration image for e2e/integration testing off the dApp.

## Acceptance criteria
- We need to check if the wizard solution (raw private key) still works (it's merged in master)

## Tasks
- [ ]

kelsos:

Fixes #1404  

**Short description**

Introduces infrastructure for e2e tests for the dApp using the integration image.
The scope of this PR is the part of the test setup up to the point we have a working test that reaches the connect screen.

**Definition of Done**

This requires the integration image to run.

1.  Go to the `/integration` folder
2. Run `run-e2e.sh` this should run the docker image and start the dApp using this environment setup.
3. You should see Cypress test runner (electron app)
4. Press run tests
5. You should see the tests running and interacting with the connection until showing the "Connect new token" screen.


nephix:
Tests worked on my machine, although the test runner seems to get stuck with a matrix request in the end

kelsos:
@nephix this is the development runner so it is in `watch` mode. 
I didn't try it in headless mode though.

raiden-dapp/package.json
```diff
@@ -6,14 +6,14 @@
   "author": "brainbot labs est.",
   "scripts": {
     "serve": "vue-cli-service serve",
-    "serve:no-logging": "export NODE_ENV=nologging && vue-cli-service serve",
     "build": "vue-cli-service build",
     "test:unit": "vue-cli-service test:unit",
-    "test:e2e": "vue-cli-service test:e2e",
+    "test:e2e": "nyc vue-cli-service test:e2e --mode e2e",
```
nephix, line 11: Do you think making it `--headless` and including it in the nightly test run makes sense?
Probably not at the moment because there is not a lot tested but maybe in the future

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -120,34 +121,25 @@ export default class RaidenService {
...
+          'networkVersion' in provider &&
+          provider.networkVersion === '1' &&
+          process.env.VUE_APP_ALLOW_MAINNET === 'false'
+        ) {
+          throw new RaidenError(ErrorCodes.RDN_UNRECOGNIZED_NETWORK);
         }
 
+        raiden = await RaidenService.createRaiden(
+          provider,
+          config?.PRIVATE_KEY ?? undefined,
```
nephix, line 138: Doesn't `config?` already return `undefined`

raiden-dapp/package.json
kelsos: Absolutely. As soon as we have more tests. Maybe depending on the runtime we could also have it in the regular workflow parallel to the other tests. Basically I was thinking the same for the sdk integration. The current runtime is around 3-4minutes so if we make it parallel to the other tests it should not increase the overall execution time.

raiden-dapp/src/services/raiden-service.ts
kelsos: Yes this `??` is redundant basically but I missed it. Nice catch :D

raiden-dapp/package.json
nephix, line 11: There is a package called [start-server-and-test](https://www.npmjs.com/package/start-server-and-test) which starts the server and as soon as it gets a 200 OK from the URL also starts the test. Might make sense to use this for headless

kelsos: Yes, I am aware of it. But here it would only make sense for running tests locally right? 
instead of starting the image by hand (using the script) we could have the package start in instead. Is that what you had in mind?

In CI it doesn't matter because cypress will run inside the container itself.

PR #1611, opened 2020-05-29
Disables minting on mainnet

### LINKED ISSUE #1579: Don't mint (only deposit) to UDC when not on a test net
## Steps to Reproduce
1. Serve dApp
2. Connect (not to a test net)
3. Click on a token
4. Acquire some SVT
5. Try to deposit in dApp

## Expected Result
- We already have SVT in the account and can deposit 

## Actual Result
- Currently it should fail because minting is only allowed on test nets
- Depositing may fail as well because minting failed

kelsos:
Fixes #1579  

**Short description**
Disable minting on mainnet and adjusts the text to remove minting from display when the network is mained

**Definition of Done**

1. Select mainnet in metamask
2. Go to try to connected a token
3. Go to select hub. The token should not have a mint button. UDC tooltip and dialog should only show info about deposit
4. UDC screen doesn't mention mint anywhere


raiden-dapp/src/store/index.ts
```diff
@@ -180,6 +180,9 @@ const store: StoreOptions<RootState> = {
     network: (state: RootState) => {
       return state.network.name || `Chain ${state.network.chainId}`;
```
andrevmatos, line 181: Notice `network.name` is always defined in ether's `Network`. With unknown network, it's simply `unknown`, that's why we have [this approach](https://github.com/raiden-network/light-client/blob/6f65f2d6ba0221a8ecc774e207c6782e71c1ab02/raiden-ts/src/utils/ethers.ts#L164) on SDK, which you can use or copy, as it's simple enough.

kelsos: It would make sense to replace it with the `getNetwork` function instead of duplicating the code.

PR #1612, opened 2020-06-01
Dependencies upgrades and eslint7 fixes

andrevmatos:
**Short description**
Upgrade all dependencies to major versions and fixes new linting errors.
Notice also that this fixes our deprecated usage of `typechain-target-ethers` since `typechain@2`, and replace it with `@typechain/ethers-v4`. No usage change beyond the right parameter to typechain script.

**Definition of Done**

1. Just test fresh install (`npm ci`) on the 3 packages works, and tests pass


raiden-dapp/.eslintrc.js
```diff
@@ -3,11 +3,9 @@ const path = require('path');
 module.exports = {
   root: true,
   env: {
-    node: true,
+    node: true
```
kelsos, line 6: Hmm, looking at this, it doesn't make sense. We don't run the dApp on a node environment, right? we target browsers instead.

raiden-ts/src/state.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable @typescript-eslint/no-explicit-any,@typescript-eslint/camelcase */
```
kelsos, line 1: Did we completely remove the camelcase requirement or something else changed? could you point to the change?

andrevmatos: The rule was [completely removed](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/camelcase.md), and replaced by [`@typescript-eslint/naming-convention`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md), which isn't (yet?) part of `@typescript-eslint/recommended` set of rules we follow, so I had to remove the disables to suppress errors about the unknown rule.

raiden-dapp/.eslintrc.js
andrevmatos: Well, this was just an `eslint --fix`, the value was already there, but tsc/babel uses some types/logic from Node, maybe that's the reason. [Here's the docs](https://eslint.org/docs/user-guide/configuring#specifying-environments) about these env, feel free to reevaluate.

PR #1623, opened 2020-06-03
Renames action tags from camel case to path format

### LINKED ISSUE #837: Rename action tags from camel case to path format
## Description

On #806 , with async actions, we introduced a new naming schema for actions, which looks like paths in a domain driven structure
example:
Before:
- `channelOpen` (request)
- `channelOpened` (success)
- `channelOpenFailed` (failure)

After:
- `channel/open/request`
- `channel/open/success`
- `channel/open/failure`

We want to be consistent with this tagging schema for the rest of the actions as well.
Examples:
- `block/new`
- `token/monitor`
- `iou/clear`

These literals should also be used only on the definition of the actions, and anywhere else, `action.type` should be used instead of the string.

## Acceptance criteria
- 

## Tasks
- [ ]

kelsos:

Fixes #837  

**Short description**

**Definition of Done**

1. Run the tests for the sdk
2. Run the tests for the dApp

### DIFF
--- a/raiden-ts/src/channels/actions.ts
+++ b/raiden-ts/src/channels/actions.ts
@@ -12,7 +12,7 @@ const ChannelId = t.type({
 });
 
 /* A new head in the blockchain is detected by provider */
-export const newBlock = createAction('newBlock', t.type({ blockNumber: t.number }));
+export const newBlock = createAction('new/block', t.type({ blockNumber: t.number }));
 export interface newBlock extends ActionType<typeof newBlock> {}
 
 /**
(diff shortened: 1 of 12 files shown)


raiden-ts/src/channels/actions.ts
```diff
@@ -12,15 +12,15 @@ const ChannelId = t.type({
 });
 
 /* A new head in the blockchain is detected by provider */
-export const newBlock = createAction('newBlock', t.type({ blockNumber: t.number }));
+export const newBlock = createAction('new/block', t.type({ blockNumber: t.number }));
```
andrevmatos, line 15: As the actions are like `domain/action`, maybe this should be
```suggestion
export const newBlock = createAction('block/new', t.type({ blockNumber: t.number }));
```

kelsos: That makes absolute sense, let me fix that.

Issue #173, opened 2019-06-09
Raiden API


christianbrb:
## Description

API which mimics the [Raiden API](https://raiden-network.readthedocs.io/en/stable/rest_api.html) to enable an integration test with the scenario player.

## Acceptance criteria
- Offers same API endpoints as Raiden python
- Endpoints take the same parameters as Raiden Python

## Tasks
- [ ] 

## Story Points

kelsos:
This would be a node.js app build on top of the SDK, using something like [Express](https://expressjs.com/)?

andrevmatos:
I think yes, and therefore, it'd be a second app for the SDK, now a node.js daemon providing the functionality over an HTTP REST API. As Raiden API is quite large, we'd need to prioritize which APIs are essential for the scenario player and implement them first.

eorituz:
That is basically the same that I tried to do during my last berlin trip @kelsos, right?
So An node+express app that mimics the raiden API functionality

kelsos:
@eorituz pretty much yes.

manuelwedler:
I created this private repo for the API express app that I am using for the demo train integration: https://github.com/raiden-network/light-client-rest-api
We can move it into the light-client repo once we added TypeScript and the existing functionality exactly mimics the Raiden API.

kelsos:
@manuelwedler I would prefer to keep it as it's own repo and probably add a link on that instead.

christianbrb:
Let's have a meeting with @manuelwedler to split the issue

nephix:
Another option might be https://www.fastify.io which is an express alternative (but the API is very similar) with the goal of being faster and allowing more throughput

weilbith:
Just throw in another biased opinion which framework to use: [KoaJS](https://koajs.com/) is a more modular and light weight version in comparison to Express and was developed by the same team. I developed with both frameworks already and _KoaJS_ was more pleasant to work with. :man_shrugging:

weilbith:
I did a quick comparison for [light-client-rest-apiu](https://github.com/raiden-network/light-client-rest-api) its current state in comparison to the official [Raiden API](https://raiden-network.readthedocs.io/en/stable/rest_api.html). Stroked endpoints are already implemented. Also mark the light client exclusive endpoints. 


### Channels

- _GET_ `/api/(version)/channels`
- _GET_ `/api/(version)/channels/(token\_address`
- _GET_ `/api/(version)/channels/(token\_address)/(partner\_address)`
[...]

PR #1636, opened 2020-06-04
Feature/transform cli to rest api

weilbith:
Related to: #173

**Short description**
This PR is not implementing the full Raiden API specification. Rather it is implementing the basics for the application, attempting to do the heavy lifting for future development. The idea is to get some early feedback for continuing to developing the rest of the API. This PR is already huge enough (the poor reviewer :wink:).
Having this base of code should also allow us to parallelize work and set new priorities for the different endpoints (like for the train demo). Nevertheless there are already some endpoints implemented. User parameter validation is established for all endpoints of the spec.

Please watch-out for some `TODO` markers. They refer to "issues" or lack of knowledge with/of the SDK. Feel free to simply fix them. In general I'm totally with initiations to improve the code yourself if you have a strong opinion.

I'm sorry that especially my first PR in first project starts so big. But I'm happy about the solution to split this epic issue now. I hope the code is good to read, it was my primary goal. The commit history is not actually meant to get merged like this, since the story of this issue was relatively chaotic. I guess we should do a final squash to keep our history clean. Future issues should be smaller and PRs can be more focused, contributing to a clean history.

**Definition of Done**


raiden-cli/src/controllers/channels.ts
```diff
@@ -0,0 +1,81 @@
+import { Request, Response, NextFunction } from 'express';
+import { ChannelState } from 'raiden-ts';
+import _ from 'lodash';
```
kelsos, line 3: In the SDK/dApp we always import the functions directly e.g. for `map` that would be `import map from 'lodash/map';`. This is to enable tree-shaking anything unused when bundling instead of pulling the whole `lodash` in the final bundle

raiden-cli/src/utils/validation.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export function validate(validations: ValidationChain[]) {
+  return async (request: Request, response: Response, next: NextFunction): Promise<void> => {
+    await Promise.all(validations.map((validation) => validation.run(request)));
+    const errors = validationResult(request).formatWith(formatValdiationErrors);
+
+    if (errors.isEmpty()) {
+      next();
+    } else {
+      response.status(404).json({ errors: errors.array() });
```
kelsos, line 25: Since this is a parameter validation error shouldn't that be a 400?

weilbith: I agree with that. The point is that the Raiden API defines it like that. But I think even them are inconsistent in which error code they throw for which invalid parameter. For example do they define [here](https://raiden-network.readthedocs.io/en/latest/rest_api.html#put--api-(version)-tokens-(token_address)) for adding a token, that the API should return a `404`. And they do this all other the place. For URL parameter as well as JSON encoded bodys. So I was basically just following their specification. :man_shrugging:

raiden-cli/src/controllers/channels.ts
weilbith: That's a good point. Sometimes (especially for such generic names) I like to have a reference while reading the code where this comes from. But you are right, having the whole lodash in the package would be nasty. I'll fix that. :+1:

weilbith: :heavy_check_mark:

raiden-cli/src/utils/validation.ts
weilbith: If you agree with the approach to implement the API specification straight forward, then simply close this conversation. Else I keep it open for further discussions.

raiden-cli/package.json
```diff
@@ -38,9 +39,13 @@
     "webpack-cli": "^3.3.11"
   },
   "dependencies": {
+    "@types/http-errors": "^1.6.3",
     "ethers": "^4.0.47",
     "express": "^4.17.1",
+    "express-validator": "^6.5.0",
```
andrevmatos, line 45: You can validate the inputs, but just would like to note that the SDK is already quite strict about input validation (address, bignumbers, etc), so don't spend too much time on cli's side validation. We can improve errors typing if you need to catch it on client's side, and I'd say this would be better than user validation.

raiden-cli/src/app.ts
```diff
@@ -0,0 +1,42 @@
...
+  next(createError(404, 'Undefined resource'));
+}
+
+function internalErrorHandler(
+  error: Errback,
+  _request: Request,
+  _response: Response,
+  next: NextFunction,
+) {
+  console.log(error); // TODO: proper error logging
```
andrevmatos, line 19: Please, avoid `console.*` calls directly, and prefer to use `loglevel`. Current cli version sets this as a global `Logger` object defaulting to root logger, and setting it to a account-specific logger after account is available. You could use that in an independent shared module ` let logger` export/import with some `setLogger` function to change that to the account-specific logger (which isn't much used currently but is nice to provide future options, like per-logger level and/or prepending the line with logger name, in case of multiple clients running in parallel)

andrevmatos: Or maybe accessing an explicit logger in some object, to avoid the global/module export.

raiden-cli/src/app.ts
```diff
@@ -0,0 +1,42 @@
...
+  next(createError(500, 'Internal Raiden node error'));
+}
+
+const app = express();
```
andrevmatos, line 23: Global/module level variables are how we did it so far, but I'd say for this improvement/rewriting, it could be worth it to have these instances in some context/object, instead of being global, which could allow multiple/independent clients to be run in a single Node instance, needed e.g. for testing (in the future). Some functions could need some context binding at some point, but this could be useful.
What do you think, @kelsos ?

raiden-cli/src/utils/validation.ts
andrevmatos: I agree our main target is Raiden API's seamless compatibility, but it'd be worthy putting a FIXME or whatever to discuss proper/better error codes on these.

raiden-cli/src/raiden.ts
```diff
@@ -0,0 +1,127 @@
...
+import { Subscription } from 'rxjs';
+
+const DEFAULT_CONFIG = {
+  matrixServer: 'https://raidentransport.test001.env.raiden.network',
+  pfs: 'https://pfs.raidentransport.test001.env.raiden.network',
+  pfsSafetyMargin: 1.1,
+  caps: { noDelivery: true, webRTC: true },
+};
+
+export default class RaidenService {
```
andrevmatos, line 13: This would be my strongest suggestion so far: I'd like to avoid a new service/wrapper. I know it's usually good practice to separate implementation from libs specifics, but since CLIs would also be a reference implementation, I'd rather have the SDK's `Raiden` instance used as close to the "surface" as possible, possibly even directly by the route handlers!
And also to be as stateless possible outside of the `Raiden` instance, so it should provide you with all state/variables/info needed at any time.
This way, we get quickly notified if any change breaks something, have simple/direct/straightforward/local views of what needs to be done to get anything in or out of the SDK (methods, data), and can use these requirements as informative requests for changing the SDK's APIs and improving it (e.g. using this to drive our decisions on #357 and other public API patterns), we'll change whatever needed on the SDK, and maybe this may drive improvements/simplifications in the dApp as well, which already wraps quite a lot of data/method calls on its vuex state/actions/services.

raiden-cli/src/index.ts
```diff
@@ -231,98 +44,71 @@ async function main() {
...
+function initLogging(address: string): void {
+  log = logging.getLogger(`cli:${address}`);
+  log.info('Address:', address);
+}
 
-  const init = raidenInit(argv);
-  raiden.start();
-  // if events$ completes, it means raiden shut down
-  stopSub = raiden.events$.subscribe(undefined, raidenStop, raidenStop);
+function shutdown(): void {
```
andrevmatos, line 80: Careful with shutdown, you're losing something here: currently,  we use `raiden?.started` to check if we already stopped it once (i.e. it's a second `Ctrl+C`), and force exit on it. It means on first SIGINT we call stop and try to wait for it to exit gracefully (node to finish any pending http request, scheduled timeout handler, etc), and if it's slow/stuck, the user can hit it again and it exit(1). Your shutdown here may lose that.

raiden-cli/src/raiden.ts
```diff
@@ -0,0 +1,127 @@
...
+  private _channels: RaidenChannels;
+  private _subscriptions: Subscription[];
+
+  constructor(raiden: Raiden) {
+    this._raiden = raiden;
+    this._channels = {};
+    this._subscriptions = [];
+  }
+
+  static getInstance(): RaidenService {
```
andrevmatos, line 26: Again, related to the 'global instances or contextual ones', these static methods also wouldn't allow for multiple instances running on same Node process, so with the suggestion get rid of services, also try to use some local instance or Raiden instead of static methods/properties.

raiden-cli/src/custom.d.ts
```diff
@@ -0,0 +1,9 @@
+import { RaidenChannel } from 'raiden-ts';
+
+declare global {
```
andrevmatos, line 3: Would this better go on a `global.d.ts` file? For consistency

raiden-cli/src/app.ts
kelsos: As I mentioned to @weilbith we already have an example for `loglevel` with `express` in the raiden-hub code.

raiden-cli/src/app.ts
kelsos: I also prefer scoping as much as possible and exposing the bare minimum to the global level.

raiden-cli/package.json
weilbith: Okay. The important point is that we should avoid to validate user data twice. This is just a bad habit and can lead to unexpected behavior. So I agree with you. 

Code this means we need to establish a proper parsing of SDK errors. We would simply forward the user parameter to the SDK and then respond to the user depending on the output/error. This means the SDK must provide us with really good error types and messages. The optimal case would be that we could do a switch case depending on the errors type and the errors message can be forwarded directly. Do you think the SDK provides these requirements already?

raiden-cli/src/app.ts
weilbith: Okay, so what does that mean now. What would you expect the module to expose and how should these instances be stored/handled. I must admit that from this conversation I don't know how to improve.

raiden-cli/src/raiden.ts
weilbith: I'm totally with you and understand the arguments. My simple point is: I think
it is not possible how it is now and the goal is to provide the API. So this
would require to lengthen the whole API development while adapting the SDK. My
goal was to provide a first compatible solution for the user (like Olli with the
train demo). Afterwards we can tinker the internals without the user
recognizing. 

Accessing the SDKs directly from the routes sounds awesome. Unfortunately it is designed in a way that someone must instantiate it, but also share the instance. This sharing of an instance is the main feature this wrapper provides. Furthermore I would not know where you would do the subscriptions and on what these subscriptors interact with. 
As I said, I share your vision. I just don't know how it should work right now.
:shrug:

raiden-cli/src/index.ts
weilbith: Oh, thanks that you caught that! I'll need to fix that. :+1:

raiden-cli/src/raiden.ts
weilbith: I like the idea to have the possibility for multiple instances. Lets see in the above discussion how we resolve that.

raiden-cli/src/custom.d.ts
weilbith: :heavy_checkmark:

raiden-cli/src/utils/validation.ts
weilbith: :heavy_check_mark:

raiden-cli/src/index.ts
weilbith: I pushed a fix for that. Could you please verify that it is correct? Just to be sure. Thanks!

raiden-cli/src/app.ts
weilbith: :heavy_check_mark:

weilbith:
@andrevmartos there are some `TODO`s at `./utils/formatting.ts` according to
format a `RaidenChannel` to what the API defines. Could you please check them
out and provide answers? Thanks. :smiley:

raiden-cli/src/raiden.ts
andrevmatos: So, just to leave it here for reference, you could go with the [shared] data as something like an interface in the form:
```ts
export interface Cli {
  raiden: Raiden;
  app: Express;
  server: Server;
  router: Router; // needed?
  log: logging.Logger; // default it to logging root logger
}
```
Then you create a simple const object matching this interface, and this will contain the variables which currently are on global scope. You can pass it to several functions, either as a first, explicit `cli: Cli` parameter and access it like `cli.raiden.start()`, or pass it as `this` parameter, declaring like:
```ts
function shutdown(this: Cli, ...args: any[]) {
  this.raiden.stop(); // this is properly typed
}
```

You can then call it then like `shutdown.apply(cli, other_params_array)`.
If you need to pass this function ahead as a callback to some place which may not bind `this` (e.g. route handler), you can pass it like:
```ts
async function handleChannelOpen(this: Cli, request: Request, _response: Response, next: NextFunction) {
  await this.raiden.openChannel(...)
}
export function setupChannelsApis(this: Cli) {
  this.router.get('/channels', handleChannelOpen.bind(this));
}
```
You can reassign members (e.g. `this.log = logging.getLogger(`cli:${this.raiden.address}`);`) if needed without problems, as long as you always access them as members from the `cli` object reference (instead of passing the members directly somewhere)
If some function needs to be called when some of these instances are not ready yet, you can pass it like `function(this: Pick<Cli, 'app' | 'server'>)`, and this will require a subset of the interface where only `app` and `server` properties are defined, not others possibly undefined members.
Try as much as possible to have a first `setupCli` or whatever entrypoint, so this can be easily used to setup a CLI instance wherever needed, and this would be the entrypoint to setup everything on the `main` function.

andrevmatos: At last, you could have your implementation directly on the router definitions, using directly the SDK instance (`this.raiden.`...), and take out only shared logic like filtering or object schema mapping/transform. This way, one can reference directly the endpoint as a reference implementation to how to perform some operation, what specific errors need to be handled, etc.

raiden-cli/src/index.ts
andrevmatos: Looks right now, besides the RaidenService discussion. Thanks!

raiden-cli/src/raiden.ts
andrevmatos: Discussed above. Thanks

raiden-cli/src/routes/api.v1.ts
```diff
@@ -0,0 +1,39 @@
...
+import { Router, Request, Response } from 'express';
+import RaidenService from '../raiden';
+import channelsRouter from './channels';
+import tokensRouter from './tokens';
+import pendingTransfersRouter from './pendingTransfers';
+import connectionsRouter from './connections';
+import paymentsRouter from './payments';
+import testingRouter from './testing';
+
+const router = Router();
```
andrevmatos, line 11: Just a headsup: global variable (and respective method calls bellow), may be encapsulated on the context bindings or such as we discussed

raiden-cli/src/utils/formatting.ts
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable @typescript-eslint/camelcase */
+import { RaidenChannel } from 'raiden-ts';
+import { ApiChannel } from '../types';
+
+export function transformChannelFormatSdkToApi(channel: RaidenChannel): ApiChannel {
```
andrevmatos, line 5: I like pure functions, no bind if not needed

raiden-cli/src/utils/formatting.ts
```diff
@@ -0,0 +1,17 @@
...
+  return {
+    channel_identifier: channel.id ?? 0, // TODO: how can this be undefined?
```
andrevmatos, line 7: As we discussed, this was a leftover of when channel state could be `opening` and no `id` would be available then. Nowadays, first channel state is `open`, and id,  settleTimeout, etc, are always defined. I'll fix.

raiden-cli/src/utils/formatting.ts
```diff
@@ -0,0 +1,17 @@
...
+    token_network_address: channel.tokenNetwork,
+    partner_address: channel.partner,
+    token_address: channel.token,
+    balance: channel.balance.toString(),
+    total_deposit: channel.ownDeposit.toString(), // TODO: must this add the partnerDeposit?
```
andrevmatos, line 12: Should be `ownDeposit`, it's correct. Thanks

raiden-cli/src/utils/formatting.ts
```diff
@@ -0,0 +1,17 @@
...
+    state: channel.state,
+    settle_timeout: channel.settleTimeout ?? 0,
+    reveal_timeout: 0, // TODO: why not defined?
```
andrevmatos, line 15: As dsicussed, this comes from `config.revealTimeout`, this config isn't specific per channel, and at some point should actually be replaced by a dynamic value based on gasPrice police and blockchain congestion, not a static/constant value, and definitely not specific per channel.

raiden-cli/src/utils/logging.ts
```diff
@@ -0,0 +1,25 @@
...
+  };
+  if (process.env.NODE_ENV === 'production') {
+    log.setLevel('INFO');
+  } else {
+    log.setLevel('DEBUG');
+  }
+  return log;
+}
+
+const logger = setupLoglevel();
```
andrevmatos, line 24: Headsup: global var

raiden-cli/src/app.ts
```diff
@@ -0,0 +1,42 @@
+import { Server } from 'http';
+import express, { Request, Response, NextFunction, Errback } from 'express';
+import createError from 'http-errors';
+import { log } from './utils/logging';
+import apiV1 from './routes/api.v1';
+
+let server: Server;
```
andrevmatos, line 7: Headsup: global var

raiden-cli/src/app.ts
andrevmatos: Already discussed. Just a headsup: global var

raiden-cli/package.json
weilbith: Just a status update here from me. We decided that the error handling in the SDK for the parameter validation must become better. This is what [that PR](https://github.com/raiden-network/light-client/pull/1649?notification_referrer_id=MDE4Ok5vdGlmaWNhdGlvblRocmVhZDkyNDczNjgxNjoxMjU0MzY0Nw%3D%3D) provides.
Anyways I close this discussion, since there will be a major rework according our decisions. They will require a new review by themselves.

raiden-cli/src/app.ts
weilbith: I close this discussion due to a rework we have decided on. This will require a new review anyways.

raiden-cli/src/raiden.ts
weilbith: Wow thank you so  much that you did not only developed this idea during our team call, but also left this long written explanation here! :pray:

As for the other discussion will close this one here as well. They lead to a major rework that will require a new review afterwards.

raiden-cli/src/routes/api.v1.ts
weilbith: :heavy_check_mark:

raiden-cli/src/utils/formatting.ts
weilbith: :heavy_check_mark:

raiden-cli/src/utils/formatting.ts
weilbith: :heavy_check_mark:

raiden-cli/src/utils/formatting.ts
weilbith: Yes. I think this is more complex than it might look on a first gaze. So since this could be critical once we join the scenario player, I opened [an issue](https://github.com/raiden-network/light-client/issues/1656) for this, that is up for discussion. I fear it is not the last endpoint in the spec where we need to do such a compatibility decision.

raiden-cli/src/utils/logging.ts
weilbith: :heavy_check_mark:

raiden-cli/src/app.ts
weilbith: :heavy_check_mark:

weilbith:
@andrevmatos I'm not sure how to solve the CI issue by `codeclimate`. For me it looks correct. :shrug:

kelsos:
> @andrevmatos I'm not sure how to solve the CI issue by `codeclimate`. For me it looks correct. 
shrug

We can ignore it.

kelsos:
I went and manually approved but maybe we have to adjust our configuration to 5. Let's discuss it tomorrow.

raiden-cli/src/controllers/channels.ts
```diff
@@ -1,6 +1,6 @@
 import { Request, Response, NextFunction } from 'express';
 import { ChannelState } from 'raiden-ts';
-import _ from 'lodash';
+import { map } from 'lodash';
```
andrevmatos: It should be:
```suggestion
import map from 'lodash/map';
```
As lodash isn't ES module by default (you'd need `lodash-es` for that), just destructuring won't get it tree-shaken, but importing the independent modules will.

raiden-cli/src/raiden.ts
```diff
@@ -1,6 +1,6 @@
 import { Raiden, RaidenChannel, RaidenChannels, ChannelState } from 'raiden-ts';
 import { LocalStorage } from 'node-localstorage';
-import _ from 'lodash';
+import { flatMap, values, filter } from 'lodash';
```
andrevmatos: Same. Unfortunatelly, one line for each, but you get used to ;)

raiden-cli/src/raiden.ts
```diff
@@ -86,16 +86,13 @@ export default class RaidenService {
     let filteredChannels = this.allChannels;
 
     if (state) {
-      filteredChannels = _.filter(filteredChannels, (channel) => channel.state === state);
+      filteredChannels = filter(filteredChannels, (channel) => channel.state === state);
```
andrevmatos: Do you need lodash's `filter` for that? Why not simply `Array.prototype.filter`?

raiden-cli/src/raiden.ts
```diff
@@ -63,7 +63,7 @@ export default class RaidenService {
 
   get allChannels(): RaidenChannel[] {
     // To resolve structure {token: {partner: [channel..], partner:...}, token...}
-    return _.flatMap(_.values(this._channels), (partnerChannels) => _.values(partnerChannels));
+    return flatMap(values(this._channels), (partnerChannels) => values(partnerChannels));
```
andrevmatos: `flatMap` here can also be replaced by `Array.prototype.reduce` to avoid the import as well... something like:
```ts
Object.entries(this._channels).reduce((acc, partnerChannels) => acc.concat(Object.values(partnerChannels)), [] as RaidenChannel[]);
```
Lodash is very useful sometimes, but there's a lot of things we don't need it for on ES6 or even old JS.

raiden-cli/src/utils/validation.ts
```diff
@@ -0,0 +1,19 @@
+import { Request, Response, NextFunction } from 'express';
+import { Address } from 'raiden-ts';
+
+export function validateAddressParameter(
+  this: string,
+  request: Request,
+  response: Response,
+  next: NextFunction,
+) {
+  const addressParameter = request.params[this];
```
andrevmatos, line 10: Nice usage ;)

raiden-cli/src/index.ts
```diff
@@ -231,96 +49,66 @@ async function main() {
...
-  raiden.start();
-  // if events$ completes, it means raiden shut down
-  stopSub = raiden.events$.subscribe(undefined, raidenStop, raidenStop);
-  try {
-    await init;
-    if (argv.serve) await setupApi(argv.serve);
-  } finally {
-    if (!argv.serve) raidenStop();
-  }
+  const cli = await makeCli(log, raiden, argv.port);
```
andrevmatos, line 109: Looks like `makeCli` is a very small function/module, which actually do only 2 things besides creating the `cli`. Wouldn't be worth it to have these 2 lines (app and server assignment) done in the end of `makeApp` (maybe rename it to `setupApp`), and then have `cli` object creation here, just folowed by a call to `setupApp(cli)`? Looks like `main` already knows about `log` and `raiden` members anyway, so I'd say it'd not deviate much from this. Nitpicking anyway, feel free to ignore this if you think it's worth to keep this separation.

raiden-cli/src/utils/logging.ts
```diff
@@ -0,0 +1,22 @@
...
+      const prefix = `${new Date(Date.now()).toISOString()} [${methodName}]`;
+      rawMethod(prefix, ...message);
+    };
+  };
+  if (process.env.NODE_ENV === 'production') {
+    log.setLevel('INFO');
+  } else {
+    log.setLevel('DEBUG');
+  }
+  return log;
```
andrevmatos, line 21: Careful: you're actually mutating and returning the global instance/root logger here. It's ok to have it in a function you're sure is called only once (e.g. on beginning of main, as you do), but please, don't return nor use it directly in the `cli`. Instead, return `void` here, and make `log` there be a `log: logging.Logger = logging.getLogger(<some string unique per cli based on address>);`. Also, if possible, for consistency, rename `log` here to `logging`, so we remember this is the root logger and not some child logger.

raiden-cli/src/types.ts
```diff
@@ -0,0 +1,34 @@
+import { Raiden } from 'raiden-ts';
```
andrevmatos, line 1: Nitpick: could you, please, on this module, import like:
```suggestion
import type { Raiden } from 'raiden-ts';
```
This is TypeScript 3.9 way of telling this import will actually be removed from transpiled output, as these imports are used only for typing, and this is an interface/types-only file.
We still need to apply this concept to our codebase, so feel free to ignore (as all nitpicks), but is more as a headsup of this nice recent language addition.

raiden-cli/src/index.ts
```diff
@@ -1,206 +1,24 @@
...
-async function main() {
-  logging.setLevel(logging.levels.DEBUG);
-  const argv = yargs
-    .usage('Usage: $0 -k <private_json_path> -e <node_url> --serve <port>')
+import { Raiden } from 'raiden-ts';
+import { CliArguments, Cli } from './types';
+import { makeCli } from './cli';
+import { setupLoglevel } from './utils/logging';
+
+const DEFAULT_RAIDEN_CONFIG = {
```
andrevmatos, line 12: Nitpick: it'd be interesting to have this in a `config.json` file, imported directly from typescript, like `import DEFAULT_RAIDEN_CONFIG from './config.json';` or something like that. Just a small QoL improvement ;)

raiden-cli/package.json
andrevmatos: Is this dependency still needed? Didn't see it imported anywhere. If not, may you clean it up, please?

andrevmatos:
Ah, don't forget to rebase before merging ;)

raiden-cli/src/controllers/channels.ts
weilbith: Oh, sorry. That was stupid. :see_no_evil:
Anyways this file does not exit anymore. Actually already when you reviewed. :thinking:

raiden-cli/src/index.ts
weilbith: Yes of course. I love it. Always hated that this was in the code. I was just focusing on other stuff...

raiden-cli/src/raiden.ts
weilbith: Replace `entries` with `values` and it works. :grimacing:

weilbith: Replace `entries` with `values` and it should work. :D

weilbith: :heavy_check_mark:

raiden-cli/src/raiden.ts
weilbith: :heavy_check_mark:

raiden-cli/src/raiden.ts
weilbith: I got rid of lodash. This file was already deleted anyways :D

raiden-cli/src/index.ts
weilbith: :heavy_check_mark:

raiden-cli/src/index.ts
weilbith: To be honest would like to keep it. I think this way the modules are more decoupled from each other. So yes the `cli` module is tiny. My intention was to use it as an accumulator for the REST server. By that it could be used everywhere, providing the arguments as needed. So it is only used by the caller of the command line with its arguments. So the `raiden` and `log` instance are retrieved from the users input. But in other cases this could be created by anyone for any need. So he can simply call `makeCli` and get his own server, that is independent of the other one(s). We may can improve that later? Maybe have a discussion. Shouldn't be much work to change later and from the user perspective nothing will change.

raiden-cli/src/utils/logging.ts
weilbith: :heavy_check_mark:

PR #1645, opened 2020-06-08
Extend git commit rules for CONTRIBUTION

weilbith:
This relates to the recent discussion we had as a team according to the suggestion by me. Please verify that the explanation is good enough and if you agree with the decided scope names. Feel free to change spelling or content on your behalf.

**Note:**
The deleted and added empty lines are caused by the auto-formatting on local machine. Unfortunately I wasn't very precise while staging the chunks. I can remove it if you dislike it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,7 +142,6 @@ For the sdk you have to run the following:
 ```bash
 $ cd raiden-ts
 $ npm run test
-
 ```
 
 For the dApp:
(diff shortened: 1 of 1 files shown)


weilbith:
One left question from my side is how we handle commits that are actually not focused on these scoped. E.g. this commit. I think there is some general stuff (templates, license, etc.) that is hard to put in a specific scope. The current approach would be to not prefix it. Not sure if this is the best solution. :shrug:

PR #1660, opened 2020-06-10
sdk: Introduces withdraw form the UDC functionality

kelsos:

SDK part #1421  

**Short description**

**Definition of Done**

1. Run the dApp
2. Open Vue dev tools
3. Select App (so that it attaches to $vm0)
4. Go to console
5. `raiden = $vm0.$raiden.raiden`
6. `await raiden.planUdcWithdraw(200000)`
7. wait for 100 blocks for the withdraw to happen

### DIFF
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -11,8 +11,17 @@ import { createLogger } from 'redux-logger';
 
 import constant from 'lodash/constant';
 import memoize from 'lodash/memoize';
-import { Observable, AsyncSubject, merge, defer, EMPTY, ReplaySubject, of } from 'rxjs';
-import { first, filter, map, mergeMap, skip, pluck } from 'rxjs/operators';
+import {
+  Observable,
+  AsyncSubject,
+  merge,
+  defer,
+  EMPTY,
+  ReplaySubject,
+  of,
+  throwError,
+} from 'rxjs';
+import { first, filter, map, mergeMap, skip, pluck, switchMap } from 'rxjs/operators';
 import logging from 'loglevel';
 
 import { TokenNetworkRegistryFactory } from './contracts/TokenNetworkRegistryFactory';
@@ -58,7 +67,7 @@ import {
   transferKey,
   transferKeyToMeta,
 } from './transfers/utils';
-import { pathFind } from './services/actions';
+import { pathFind, udcWithdraw, udcWithdrawPlanned } from './services/actions';
 import { Paths, RaidenPaths, PFS, RaidenPFS, IOU } from './services/types';
 import { pfsListInfo } from './services/utils';
 import { Address, Secret, Storage, Hash, UInt, decode, isntNil } from './utils/types';
@@ -1255,6 +1264,25 @@ export class Raiden {
       { log: this.log },
     );
     return receipt.transactionHash as Hash;
+  }
+
+  public async planUdcWithdraw(value: BigNumberish): Promise<Hash> {
+    const withdrawPlan = merge(
+      this.action$.pipe(
+        filter(udcWithdraw.failure.is),
+        switchMap((action) => throwError(action.payload)),
+      ),
+      this.action$.pipe(
+        filter(udcWithdrawPlanned.is),
+        filter((action) => !!action.meta.txHash),
+        map((action) => action.meta.txHash!),
+      ),
+    )
+      .pipe(first())
+      .toPromise();
+
+    this.store.dispatch(udcWithdraw.request({ amount: bigNumberify(value) as UInt<32> }, {}));
+    return withdrawPlan;
   }
 }
 
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -1259,7 +1259,7 @@ export class Raiden {
 
   public async planUdcWithdraw(value: BigNumberish): Promise<Hash> {
     const meta = { amount: bigNumberify(value) as UInt<32> };
-    const promise = asyncActionToPromise(udcWithdraw, meta, this.action$, false).then(
+    const promise = asyncActionToPromise(udcWithdraw, meta, this.action$, true).then(
       ({ txHash }) => txHash!,
     );
     this.store.dispatch(udcWithdraw.request(undefined, meta));
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -11,17 +11,8 @@ import { createLogger } from 'redux-logger';
 
 import constant from 'lodash/constant';
 import memoize from 'lodash/memoize';
-import {
-  Observable,
-  AsyncSubject,
-  merge,
-  defer,
-  EMPTY,
-  ReplaySubject,
-  of,
-  throwError,
-} from 'rxjs';
-import { first, filter, map, mergeMap, skip, pluck, switchMap } from 'rxjs/operators';
+import { Observable, AsyncSubject, merge, defer, EMPTY, ReplaySubject, of } from 'rxjs';
+import { first, filter, map, mergeMap, skip, pluck } from 'rxjs/operators';
 import logging from 'loglevel';
 
 import { TokenNetworkRegistryFactory } from './contracts/TokenNetworkRegistryFactory';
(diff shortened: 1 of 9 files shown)


raiden-ts/src/services/actions.ts
```diff
@@ -45,3 +45,26 @@ export interface iouClear extends ActionType<typeof iouClear> {}
...
 export const udcDeposited = createAction('udc/deposited', UInt(32));
 export interface udcDeposited extends ActionType<typeof udcDeposited> {}
+
+export const udcWithdraw = createAsyncAction(
+  t.partial({}),
+  'udc/withdraw/request',
+  'udc/withdraw/success',
+  'udc/withdraw/failure',
+  t.type({ amount: UInt(32) }),
+  t.type({ txHash: Hash, amount: UInt(32) }),
```
andrevmatos, line 55: Is this a `ConfirmableAction`? It be good to stick to the schema (txHash, txBlock, confirmed)

kelsos: it is kind of weird, not really a confirmable action I just wait inside before I fire the withdrawPlanned

raiden-ts/src/services/actions.ts
```diff
@@ -45,3 +45,26 @@ export interface iouClear extends ActionType<typeof iouClear> {}
...
+export namespace udcWithdraw {
+  export interface request extends ActionType<typeof udcWithdraw.request> {}
+  export interface success extends ActionType<typeof udcWithdraw.success> {}
+  export interface failure extends ActionType<typeof udcWithdraw.failure> {}
+}
+
+export const udcWithdrawPlanned = createAction(
+  'udc/withdraw/planned',
+  t.type({ amount: UInt(32), block: UInt(32) }),
+  t.partial({ txHash: Hash }),
```
andrevmatos, line 67: The meta here is the txHash, right? This doesn't seem descriptive enough as an 'identifier' of the planning, since it would only be available after request is mined (so can't be shared with `udcWithdraw`). Maybe you could have `amount` as a general `UdcWithdrawId` codec for `meta`, and use it for `udcWithdraw` and `udcWithdrawPlanned`, and the rest of the parameters in the `payload`. Also, `block` here is a small number, it'd be more consistent with the rest of the SDK to have it as `t.number`, and maybe call it `txBlock`, for consistency, even if it isn't a confirmable action (but could be as well).

raiden-ts/src/services/epics.ts
```diff
@@ -695,3 +706,146 @@ export const monitorRequestEpic = (
...
   );
+
+export const udcWithdrawRequestEpic = (
+  action$: Observable<RaidenAction>,
+  state$: Observable<RaidenState>,
+  { userDepositContract, address, log, signer, config$, provider }: RaidenEpicDeps,
+): Observable<udcWithdrawPlanned | udcWithdraw.failure> => {
+  return action$.pipe(
+    filter(udcWithdraw.request.is),
+    withLatestFrom(from(userDepositContract.functions.balances(address))),
```
andrevmatos, line 717: This won't work. `withLatestFrom` would fire only once (at subscription time), not stay up to date, and if a request comes right away, this could suppress it because the result would still be undefined.
Instead, what you want os something like:
```suggestion
    mergeMap((action) => from(userDepositContract.functions.balances(address).then((balance) => [action, balance] as const))),
```

raiden-ts/src/services/actions.ts
kelsos: Ok, can you provide an explanation on when to put something in meta and when in the payload? We had that discussion in the past but a lot of time has passed.

raiden-ts/src/services/epics.ts
kelsos: OK that makes sense thank you

raiden-ts/src/services/actions.ts
andrevmatos: `meta` is intended to be data that 'uniquely' identifies actions of a family between different 'instances'. E.g. when a channel action go through (request, event, etc), the meta is `{ tokenNetwork, partner }`, because this uniquely identifies on which channel this action acts. Other actions of the same type but on different channel would not conflict even if happening in parallel. So, for your udcWithdraw actions, which data/property identifies a time series of actions which belong to a specific request? This goes on meta, the rest is specific to each individual action and go on payload (including confirmable params, etc)

raiden-ts/src/services/epics.ts
```diff
@@ -695,3 +706,146 @@ export const monitorRequestEpic = (
...
+export const udcCheckWithdrawPlannedEpic = (
+  action$: Observable<RaidenAction>,
+  state$: Observable<RaidenState>,
+  { userDepositContract, address }: RaidenEpicDeps,
+): Observable<udcWithdrawPlanned> => {
+  return action$.pipe(
+    filter(newBlock.is),
+    first(),
+    concatMap(() =>
+      from(userDepositContract.functions.withdraw_plans(address)).pipe(
```
andrevmatos, line 787: As you're not using newBlock, and do want to check this on subscription, may I suggest you put `defer(() => userDepositContract.functions.wihtdraw_plans(address)).pipe(...` actually as the top-level observable? No need for action$ here, but `defer` is important to call the promise only at subscription.

raiden-ts/src/services/actions.ts
andrevmatos: Forget about `txBlock` on planned action, just saw it's not the tx block, but the withdraw block... it still can be a number though =)

raiden-ts/src/services/epics.ts
```diff
@@ -695,3 +706,146 @@ export const monitorRequestEpic = (
...
+  action$: Observable<RaidenAction>,
+  state$: Observable<RaidenState>,
+  { log, userDepositContract, address, signer }: RaidenEpicDeps,
+): Observable<udcWithdraw.success | udcWithdraw.failure> => {
+  return action$.pipe(filter(udcWithdrawPlanned.is)).pipe(
+    mergeMap((action) =>
+      state$.pipe(
+        pluck('blockNumber'),
+        first((blockNumber) => action.payload.block.lte(blockNumber)),
+        map(() => action),
```
andrevmatos, line 813:
```suggestion
        mapTo(action),
```

raiden-ts/src/services/epics.ts
```diff
@@ -695,3 +706,146 @@ export const monitorRequestEpic = (
...
+      ),
+    ),
+    withLatestFrom(from(userDepositContract.functions.balances(address))),
```
andrevmatos, line 816: `withLatestFrom` again..

kelsos: The same as above with mergeMap?:

raiden-ts/src/services/epics.ts
```diff
@@ -695,3 +706,153 @@ export const monitorRequestEpic = (
...
+          ErrorCodes.UDC_PLAN_WITHDRAW_NO_BALANCE,
+          {
+            balance: balance.toString(),
+          },
+        ]);
+      } catch (e) {
+        return of(udcWithdraw.failure(e, { amount: action.meta.amount }));
+      }
+      const contract = getContractWithSigner(userDepositContract, signer);
+      return from(contract.functions.withdraw(action.meta.amount)).pipe(
```
andrevmatos, line 834: Don't use `from` inside `concatMap`, as it calls the function and creates the promise at creation time, instead of subscription time (which is the whole point of using concatMap instead of e.g. mergeMap)`. Use `defer(() => promise)` instead. Maybe the `balances` map & filter could/should also be under `defer` to ensure this evaluation happens when subscribed instead of before and to get an up-to-date view.

raiden-ts/src/services/epics.ts
andrevmatos: Yes, but also see the `defer` comment below

raiden-ts/src/raiden.ts
```diff
@@ -1256,6 +1265,25 @@ export class Raiden {
     );
     return receipt.transactionHash as Hash;
   }
+
+  public async planUdcWithdraw(value: BigNumberish): Promise<Hash> {
+    const withdrawPlan = merge(
+      this.action$.pipe(
+        filter(udcWithdraw.failure.is),
```
andrevmatos, line 1272: If you have `.success` action as the role currently occupied by `Planned`, you can use `asyncActionToPromise` utility to wait for the success/failure response. It may be important to have a meaningful `meta` for it, but would work with empty object as well.

raiden-ts/src/services/actions.ts
```diff
@@ -45,3 +45,30 @@ export interface iouClear extends ActionType<typeof iouClear> {}
...
+const UdcWithdrawId = t.type({
+  amount: UInt(32),
+});
+
+export const udcWithdraw = createAsyncAction(
+  UdcWithdrawId,
+  'udc/withdraw/request',
+  'udc/withdraw/success',
+  'udc/withdraw/failure',
+  t.type({}),
```
andrevmatos, line 58: Nitpick: when an action doesn't accept a payload, I usually use `t.undefined` instead of an empty type

PR #1694, opened 2020-06-12
Filters out service tokens from token select

taleldayekh:

Fixes # 

**Short description**
This PR takes care of the bug that displays the service token in the token list view and select hub view.

**Definition of Done**

1. Connect to the dApp on mainnet
2. Choose to connect new token

### DIFF
--- a/raiden-dapp/src/store/index.ts
+++ b/raiden-dapp/src/store/index.ts
@@ -41,6 +41,7 @@ const _defaultState: RootState = {
   accessDenied: DeniedReason.UNDEFINED,
   channels: {},
   tokens: {},
+  tokenAddresses: [],
   transfers: {},
   presences: {},
   network: PlaceHolderNetwork,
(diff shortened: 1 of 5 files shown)


raiden-dapp/src/store/index.ts
```diff
@@ -156,13 +160,16 @@ const store: StoreOptions<RootState> = {
...
+      Object.values(state.tokens)
+        .sort((a: Token, b: Token) => {
+          if (hasNonZeroBalance(a, b)) {
+            return (b.balance! as BigNumber).gt(a.balance! as BigNumber)
+              ? 1
+              : -1;
+          }
+          return a.symbol && b.symbol ? a.symbol.localeCompare(b.symbol) : 0;
+        })
+        .filter(token => state.tokenAddresses.includes(token.address)),
```
kelsos, line 172: Wouldn't it be better to first filter and then sort? 

In this case, it is just an extra item so it is not too big of an issue, but in general, I would put the filter first to avoid sorting items that won't be needed

taleldayekh: Alrighty! Thx for that feedback.

Issue #1646, opened 2020-06-08
Implement missing API endpoints for the train demo


christianbrb:
## Description

From https://github.com/raiden-network/light-client/issues/173#issuecomment-640456211

> The CLI works perfectly fine, lightning-fast and provides almost every needed endpoint. It only misses the GET .../api/(version)/payments/:token/:partner_address.
This is currently a blocker for the demo train, as I need this endpoint to verify that the train sent a payment with a unique given identifier.

CC @eorituz 

## Acceptance criteria
- endpoint **PUT** `api/v1/channels` is implemented
- endpoint **POST** `api/v1/payments/:token/:receiver_address` is implemented
- endpoint **GET** `api/v1/payments/:token/:partner_address` is implemented

## Tasks
- [ ]

christianbrb:
For testing the integration image could get used

weilbith:
Finally the last PR was merged. Sorry for the delay.
Now, everything for the train demo should be there. Looking forward to get feedback. In this case would be no feedback the best, but... :wink:

PR #1701, opened 2020-06-15
Implement open channel endpoint + extend decode error handling

weilbith:
Related to: #1646

**Short description**

This PR consists of two changes that go hand in hand.
First the decode error handling by the SDK has been extended. This has been
already taking advantage of for the `openChannel` method.
Second has the **PUT** `/api/v1/channels` endpoint become implemented with
a first version of the error validation
Checkout the according commit messages to get more insights.

**Definition of Done**

Example `curl` request: 

```
curl -i -X PUT http://localhost:5001/api/v1/channels -H 'Content-Type: application/json' --data-raw '{"token_address": "0x0f84DB873066d53379eC99048609d51b6239c409", "partner_address": "0x1F916ab5cf1B30B22f24Ebf435f53Ee665344Acf", "total_deposit": "1"}'
```

### DIFF
--- a/raiden-cli/src/routes/channels.ts
+++ b/raiden-cli/src/routes/channels.ts
@@ -23,7 +23,7 @@ function isConflictError(error: RaidenError): boolean {
 async function getAllChannels(this: Cli, _request: Request, response: Response) {
   const channelDictionary = await this.raiden.channels$.pipe(first()).toPromise();
   const channelList = flattenChannelDictionary(channelDictionary);
-  const formattedChannels = transformSdkChannelFormatToApi(channelList);
+  const formattedChannels = channelList.map((channel) => transformSdkChannelFormatToApi(channel));
   response.json(formattedChannels);
 }
 
@@ -31,22 +31,18 @@ async function getChannelsForToken(this: Cli, request: Request, response: Respon
   const channelDictionary = await this.raiden.channels$.pipe(first()).toPromise();
   const channelList = flattenChannelDictionary(channelDictionary);
   const filteredChannels = filterChannels(channelList, request.params.tokenAddress);
-  const formattedChannels = transformSdkChannelFormatToApi(filteredChannels);
+  const formattedChannels = filteredChannels.map((channel) =>
+    transformSdkChannelFormatToApi(channel),
+  );
   response.json(formattedChannels);
 }
 
 async function getChannelsForTokenAndPartner(this: Cli, request: Request, response: Response) {
   const channelDictionary = await this.raiden.channels$.pipe(first()).toPromise();
-  const channelList = flattenChannelDictionary(channelDictionary);
-  const filteredChannels = filterChannels(
-    channelList,
-    request.params.tokenAddress,
-    request.params.partnerAddress,
-  );
-  const formattedChannels = transformSdkChannelFormatToApi(filteredChannels);
+  const channel = channelDictionary[request.params.tokenAddress]?.[request.params.partnerAddress];
 
-  if (formattedChannels.length) {
-    response.json(formattedChannels[0]);
+  if (channel) {
+    response.json(transformSdkChannelFormatToApi(channel));
   } else {
     response.status(404).send('The channel does not exist');
   }
@@ -61,8 +57,9 @@ async function openChannel(this: Cli, request: Request, response: Response, next
       deposit: request.body.total_deposit,
     });
     const channelDictionary = await this.raiden.channels$.pipe(first()).toPromise();
-    const newChannel = channelDictionary[request.body.token_address][request.body.partner_address];
-    response.status(201).json(newChannel);
+    const newChannel =
+      channelDictionary[request.body.token_address]?.[request.body.partner_address];
+    response.status(201).json(transformSdkChannelFormatToApi(newChannel));
   } catch (error) {
     console.log(error.code);
     if (isInvalidParameterError(error)) {
--- a/raiden-cli/src/routes/channels.ts
+++ b/raiden-cli/src/routes/channels.ts
@@ -20,75 +20,82 @@ function isConflictError(error: RaidenError): boolean {
   );
 }
 
+async function getAllChannels(this: Cli, _request: Request, response: Response) {
+  const channelDictionary = await this.raiden.channels$.pipe(first()).toPromise();
+  const channelList = flattenChannelDictionary(channelDictionary);
+  const formattedChannels = transformSdkChannelFormatToApi(channelList);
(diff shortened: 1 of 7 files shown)


weilbith:
@kelsos how can we fix the issue with `codeclimate` of too complex/long functions? Due to our functional like approach we need to make each module export itself as a function explicitly. So this is indented to fail I would say. Splitting each route entry to its own function would become horribly readable, wouldn't it? :thinking:

weilbith:
The workflow is failing for the moment, because I introduced an incompatible
decoding type that is not expected. So the issue is that the spec defines
integers can be send as strings as well. But `io-ts` can't decode a string to
a number. So we eventually need to brand the `t.number`. I thought that `UInt`
would be suitable, but I think this is too much of a change, since this affects
a mass of code.

kelsos:
> @kelsos how can we fix the issue with `codeclimate` of too complex/long functions? Due to our functional like approach we need to make each module export itself as a function explicitly. So this is indented to fail I would say. Splitting each route entry to its own function would become horribly readable, wouldn't it?

I am not sure how hard to read it would be. To be honest, I would probably go through the "a function" for each route approach. I don't know how easy it would be to have exceptions for `codeclimate`, but the 50 was a sensible reasonable length we decided after much thought.

raiden-cli/src/utils/channels.ts
```diff
@@ -27,18 +27,16 @@ export function filterChannels(
...
+    channel_identifier: channel.id ?? 0, // FIXME: old "bug" in the SDK
+    token_network_address: channel.tokenNetwork,
+    partner_address: channel.partner,
+    token_address: channel.token,
+    balance: channel.balance.toString(),
+    total_deposit: channel.ownDeposit.toString(),
+    state: channel.state,
+    settle_timeout: channel.settleTimeout ?? 0,
+    reveal_timeout: 0, // FIXME: Not defined here. Python client handles reveal timeout differently,
+  } as ApiChannel;
```
andrevmatos, line 41: Nitpick: don't type-assert here, this could hide issues if something is wrong with your object-literal declaration. Instead, have the type declared on an variable assignment or return type (as you did) and return/assign the object-literal directly, this way TSC will complain if something isn't right, and you can fix it explicitly.

raiden-cli/src/routes/channels.ts
```diff
@@ -1,60 +1,97 @@
...
+    transformSdkChannelFormatToApi(channel),
+  );
+  response.json(formattedChannels);
+}
+
+async function getChannelsForTokenAndPartner(this: Cli, request: Request, response: Response) {
+  const channelDictionary = await this.raiden.channels$.pipe(first()).toPromise();
+  const channel = channelDictionary[request.params.tokenAddress]?.[request.params.partnerAddress];
+
+  if (channel === undefined) {
```
andrevmatos, line 44:
```suggestion
  if (channel) {
```
`channel === undefined` would be the `else` condition. Also, for objects, no need to be explicit about `undefined`, as simple boolean conditional check is enough unless you'd have other falsy values in your union (which isn't the case here)

raiden-ts/src/raiden.ts
```diff
@@ -592,15 +592,31 @@ export class Raiden {
     assert(Address.is(partner), [ErrorCodes.DTA_INVALID_ADDRESS, { partner }], this.log.info);
     const tokenNetwork = await this.monitorToken(token);
     assert(!options.subkey || this.deps.main, ErrorCodes.RDN_SUBKEY_NOT_SET, this.log.info);
-    const deposit = options.deposit === undefined ? undefined : decode(UInt(32), options.deposit);
+
+    // Note that we use the advantage of the UInt decoding here, but immediately
+    // convert it to a plain number again.
+    const settleTimeout =
+      options.settleTimeout === undefined
```
andrevmatos, line 599:
```suggestion
      !options.settleTimeout
```
Only other falsy values are also invalid (`0`, `''`)

raiden-ts/src/raiden.ts
```diff
@@ -592,15 +592,31 @@ export class Raiden {
...
+        ? undefined
+        : decode(
+            UInt(4),
+            options.settleTimeout,
+            ErrorCodes.DTA_INVALID_TIMEOUT,
+            this.log.info,
+          ).toNumber();
+
+    const deposit =
+      options.deposit === undefined
```
andrevmatos, line 609:
```suggestion
      !options.deposit
```

raiden-ts/src/utils/types.ts
```diff
@@ -22,12 +23,29 @@ function reporterAssert<T>(value: Either<t.Errors, T>): asserts value is Right<T
...
+  log?: (...args: any[]) => void,
+): C['_A'] {
+  try {
+    const decoded = codec.decode(data);
+    reporterAssert(decoded);
+    return decoded.right;
+  } catch (originalError) {
+    log?.('__decode failed:', codec, data);
+
+    if (customError === undefined) {
```
andrevmatos, line 43:
```suggestion
    if (!customError) {
```

raiden-cli/src/routes/channels.ts
weilbith: Oh damn. See? We need some kind of tests I can run quickly if I do such code rework. I can't do this manually all the time. :expressionless:

Thanks that you caught that!

raiden-ts/src/raiden.ts
weilbith: I was actually re-using your code. :laughing: 
But sure it is an improvement. I guess you could argue that the user should be forced to be more explicit, but...

raiden-cli/src/utils/channels.ts
weilbith: :heavy_check_mark:

raiden-cli/src/routes/channels.ts
weilbith: :heavy_check_mark:

raiden-ts/src/raiden.ts
weilbith: :heavy_check_mark:

raiden-ts/src/utils/types.ts
weilbith: :heavy_check_mark:

raiden-ts/src/raiden.ts
weilbith: :heavy_check_mark:

weilbith:
I addressed all review comments. I'm about to rebase the fixup commits. If CI doesn't complain I'll merge afterwards.

Issue #1374, opened 2020-04-20
I want to see if a monitoring service has submitted my balance proof


christianbrb:
## User Story
As Chris I want to see if a monitoring service has submitted my balance proof so that I know, that a monitoring service has acted on my behalf.

Split from #1352 

- If possible show how much the user has paid the monitoring service to submit the balance proof. If not, split into a separate story.

## Tasks
- SDK monitors monitoring service contract and fire events if some channel of us got updated by a MS
- Display if balance proof has been submitted as a first notification in the notification panel #168 .

christianbrb:
For the SDK part. There are events in the contract. We need to monitor them.

andrevmatos:
@taleldayekh SDK part was done in above PR. dApp part is still missing, but as you're the writer of the notification area, maybe it'd be easier for you to look at the event and see how that event would fit with your pattern?

PR #1714, opened 2020-06-17
Monitors MS contract and emit event when it acts

andrevmatos:
SDK part of #1374 

**Short description**
Monitors MS contract and fires a `ms/balanceProof/sent` event through `Raiden.events$` observable if a Monitoring Service acted while we were offline and updated one of our channels.
This event's payload contain the following fields:
- `tokenNetwork`: token network from which the channel belongs
- `partner`: partner with which we had this channel
- `id`: channel Id
- `reward`: reward collected by MS from our UDC deposit for this service
- `nonce`: order of the state change on partner's channel state end which got updated
- `monitoringService`: address of the MS which acted and collected the reward
- `txHash`: hash of the transaction on-chain
- `txBlock`: when the transaction took place
- `confirmed`: if the transaction already got confirmed

**Definition of Done**

Not that easy to test, but this worked here:
1. Ensure receiving is enabled for any channel value: e.g. `rateToSvt: { "<token_addr>": MaxUint256 }`
2. Ensure there's enough UDC deposit to pay for MS, e.g. 10SVT
3. While online, receive a transfer, check `RequestMonitoring` message was sent
4. Go offline
5. On partner (sender)'s side, close the channel
6. Don't come back online with receiver and wait `settleTimeout= 500` (default) blocks to pass
7. Channel becomes settleable, MS should have sent an update tx
8. Come back online with receiving node, see `ms/balanceProof/sent` action be detected and emitted
9. Either sender or receiver can then settle the channels, receiver should receive its due tokens

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -8,17 +8,19 @@
 - [#1651] Fix PFS being disabled if passed an undefined default config
 
 ### Added
+- [#1374] Monitors MonitoringService contract and emit event when MS acts
 - [#1421] Adds support for withdrawing tokens from the UDC
 - [#1642] Check token's allowance before deposit and skip approve
 - [#1701] Allow parameter decoding to throw and log customized errors
 - [#1701] Add and extend error codes for user parameter validation
 
 ### Changed
-- [#1610] Updates smart contracts to v0.37.0 (Alderaan)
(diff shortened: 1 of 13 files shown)


andrevmatos:
Also, as part of this issue, some `unit/epics/raiden.spec.ts` tests started failing, and it was easier to convert it to the new mocks pattern than to fix this rather old test suite. All previous tests were rewritten 1:1.

PR #1733, opened 2020-06-18
Adds CLI to the CircleCI configuration

kelsos:
**Short description**

Adds cli to CirclCI

**Definition of Done**

1.
2.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,8 @@ var_2: &sdk_cache_key raiden-ts-{{ .Branch }}-{{ checksum "package-lock.json" }}
 var_3: &dapp_cache_key raiden-dapp-{{ .Branch }}-{{ checksum "package-lock.json" }}
 var_4: &sdk_working_dir ~/src/raiden-ts
 var_5: &dapp_working_dir ~/src/raiden-dapp
+var_6: &cli_cache_key raiden-cli-{{ .Branch }}-{{ checksum "package-lock.json" }}
+var_7: &cli_working_dir ~/src/raiden-cli
 
 anchor_1: &sdk_package_lock_key
   key: *sdk_cache_key
@@ -21,6 +23,9 @@ anchor_2: &attach_options
 
 anchor_3: &dapp_package_lock_key
   key: *dapp_cache_key
+
+anchor_4: &cli_package_lock_key
+  key: *cli_cache_key
 
 executors:
   base-executor:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -63,6 +68,9 @@ jobs:
       - run:
           name: Check dApp Licenses
           command: cd ~/src/raiden-dapp && license-checker --production --onlyAllow 'MIT;BSD;Apache;MPL;LGPL;ZPL;ISC;WTFPL;Unlicense' --excludePackages 'uuid@2.0.1' --excludePrivatePackages # uuid@2.0.1 has an issue with the license information but it is licensed under MIT
+      - run:
+          name: Check CLI Licenses
+          command: cd ~/src/raiden-cli && license-checker --production --onlyAllow 'MIT;BSD;Apache;MPL;LGPL;ZPL;ISC;WTFPL;Unlicense' --excludePackages 'uuid@2.0.1' --excludePrivatePackages # uuid@2.0.1 has an issue with the license information but it is licensed under MIT
```
weilbith, line 73: Nitpick:
We have basically the same line 3 times with the same comment copied. Since this is a mono repo we can't know if there isn't more to come. But even these three are enough in my opinion to make this somewhat re-usable. If you agree and like the idea I would recommand to create a [custom command](https://circleci.com/docs/2.0/reusing-config/#authoring-reusable-commands) and add the directory as a parameter for this command. This approach might be nice for some other places as well. :shrug:

.circleci/config.yml
```diff
@@ -97,6 +105,22 @@ jobs:
           paths:
             - ./*
 
+  install_cli:
+    executor: base-executor
+    working_directory: *cli_working_dir
+    steps:
+      - attach_workspace: *attach_options
+      - restore_cache: *cli_package_lock_key
+      - run: npm ci
```
weilbith, line 114: I must admit that I'm a bit unexperienced with this command. But doesn't this require to have `npm install` first? So this might work as long as CircleCI can restore a cache. But for the first time on a new cache version this would fail, wouldn't it? :thinking:

kelsos: This is a clean install. It deletes `node_modules` and then installs everything as it is in `package-lock.json`

https://docs.npmjs.com/cli/ci.html

.circleci/config.yml
kelsos: Sure let's open an issue for that :)

weilbith: Issue: #1739

PR #1778, opened 2020-06-24
Fix PFSFeeUpdates not being sent and preventing mediation

### LINKED ISSUE #1777: SDK doesn't send PFSFeeUpdates
Currently, `pfsFeeUpdateEpic` try to send `PFSFeeUpdate` messages through `messageGlobalSend` actions as soon as it gets subscribed and a state goes through (which is expected), but these early actions does not reactively detect the global room because they are filtered out and don't come on sync, and, as as workaround, we set them directly on the state (to be used/picked up later) but don't `emit` the respective `Room` matrix event, which prevents the waiting `getRoom$` utility to use this room to send the message. Later (delayed) global room messages (like the more common `PFSCapacityUpdate`) are sent because then the global room would already be available per `roomId` on client.
Just emitting the `Room` object though triggered another bug, where the `matrixLeaveUnknownRoomsEpic` was not able to detect a room was global (by alias) and was leaving it after 30s.
The initial bug affects **mediation** through LC only, but is quite relevant to be fixed, as it could hide other bugs.

## Steps to Reproduce
1. Enable and try to medaite through LC

## Expected Result
- LC sends the `PFSFeeUpdate` and becomes eligible to be picked by PFS to mediate the transfer

## Actual Result
- LC doesn't send the message on pre-existing channels and therefore can't be used to mediate

andrevmatos:
Fixes #1777

**Description**
Fixes the described issue.
Additionally, this includes some very relevant improvements to debouncing behafor for `PFSCapacityUpdate` and `RequestMonitoring` epics. Instead of some absolute debouncing, now `pairwise` operator helps compare a channel state with the previous one. This allows debouncing **only** when a pending transfer gets reduced (meaning there's a big chance of it being completed on next seconds) and allowing us to save some signatures and messages/http roundtrips (original intent of the debounces), while being smarter about when a transfer got completed instead or a deposit was made, skipping the debounce timeout and acting right away. There're tests for these specific behaviors.

**Definition of Done**

1. Check issue doens't happen when mediation is enabled
2. PFSCapacityUpdate and MonitoringRequest are performed immediatelly when a transfer succeeds or a deposit gets confirmed, and are debounced on transfer pending.

### DIFF
--- a/raiden-ts/src/services/epics.ts
+++ b/raiden-ts/src/services/epics.ts
@@ -1,5 +1,5 @@
 import * as t from 'io-ts';
-import { defer, EMPTY, from, merge, Observable, of, combineLatest } from 'rxjs';
+import { defer, EMPTY, from, merge, Observable, of, combineLatest, timer } from 'rxjs';
 import {
   catchError,
   concatMap,
@@ -21,6 +21,8 @@ import {
   skip,
   take,
   mapTo,
+  debounce,
+  pairwise,
 } from 'rxjs/operators';
 import { fromFetch } from 'rxjs/fetch';
 import { Event } from 'ethers/contract';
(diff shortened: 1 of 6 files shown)


andrevmatos:
Changelog isn't needed, since mediation isn't exposed to the users yet.

PR #1822, opened 2020-06-27
Refactor TokenNetwork events fetching

### LINKED ISSUE #1780: Open Channel Modal is not closing
## Steps to Reproduce
1. Navigate to Select Hub
2. Enter hub.raiden.eth
3. Click on Select Hub
4. Enter an amount
5. Click on Open Channel
6. Sign all the messages (Last one: Set Total Deposit - Confirmed)

## Expected Result
- I am getting redirected to the transfer screen

## Actual Result
- The modal screen is open and the spinner spinning

Logs: 
[Todays_logs.txt](https://github.com/raiden-network/light-client/files/4823760/Todays_logs.txt)

```
- Firefox
```

### LINKED ISSUE #1787: Channel Opening not succesfull - event not returned by MetaMask
## Steps to Reproduce
1. Open a channel 
2. Select a hub

## Expected Result
- Light client should Open a channel along with deposit

## Actual Result
- The channel opening is successful but deposit is not successful
- The light client also does not show the channel which has been opened, this is 
valid both when channel is opened from and to the light client.
- There are no errors on the console.

[raiden_2020-06-24T12_02_33.749Z.log](https://github.com/raiden-network/light-client/files/4825229/raiden_2020-06-24T12_02_33.749Z.log)

```
- Browser & version
- Browser console errors
- Device / OS
```

### LINKED ISSUE #1792: Root Cause Analysis for bugs
## Description
Doing a [root cause analysis ](https://smartbear.com/learn/performance-monitoring/what-is-root-cause-analysis/?l=ua) for the latest bugs, which have been piling up.

## Acceptance criteria
- 

## Tasks
- [ ]

andrevmatos:
Fix #1787 
Fix #1780 
Should fix #1792 

**Short description**
Complete rewrite of TokenNetwork events monitoring logic.
This should not have drawbacks, and UX shouldn't be changed (besides things just working™)

Before:
- `tokenMonitoredEpic` was responsible for monitoring ChannelOpened events, was stateful on which blocks it had already seen and couldn't recover it network errors made it lose events (issue 1780 & 1787)
  - it was also inneficient because although it did only 1 getLogs call, it was too wide (any ChannelOpened event with anyone) and filtered client-side, could be thousands of events on first launch e.g. on mainnet
- `channelMonitoredEpic` then, for each channel, did a getLogs call on every startup (narrowed by channelId) and kept 1 filter

Now:
- `channelEventsEpic`: a single epic per TokenNetwork
  - 2-3 getLogs on each startup, but which are ranged from latest seen event blockNumber, and are strictly limited to channels of interest (with us), so even on initial sync, won't need to fetch thousands of irrelevant channels as before
  - 1 single filter per TokenNetwork (for all channels), only on new blocks it's a catch-all filter (filtering client-side), which can't be too many events but are more performant on number of network requests required
  - If network issues which can't be detected makes it not get some logs, it'll scan again from the narrowest safe range possible (since latest event seen), and will pick up any missed event

Costs per TokenNetwork monitored (N is number of channels):
Before: `(1 + N) * (getLogs, filters)` (plus first TN getLogs could be very expensive)
Now: `2..3 * getLogs + 1 * filter` (all getLogs are as specific as possible)

**Definition of Done**

1. Not much to say, just test past and new events are detected

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -9,6 +9,7 @@
 - [#1690] Fix LockExpired with empty balanceHash verification
 - [#1698] Fix estimateGas errors on channelOpen not properly being handled
 - [#1761] Fix deposit error on openChannel not rejecting promise
+- [#1787] Fix TokenNetwork monitoring losing events
 
 ### Added
 - [#1374] Monitors MonitoringService contract and emit event when MS acts
@@ -25,6 +26,7 @@
 - [#1657] Expose RaidenChannel's id,settleTimeout,openBlock as required properties
 - [#1708] Expose RaidenTransfer's secret as optional property
 - [#1705] All transfers become monitored per default to make receiving transfers safe
+- [#1822] Refactor and optimize TokenNetwork events monitoring: one filter per Tokennetwork
 
 [#837]: https://github.com/raiden-network/light-client/issues/837
 [#1374]: https://github.com/raiden-network/light-client/issues/1374
@@ -44,6 +46,8 @@
 [#1705]: https://github.com/raiden-network/light-client/issues/1705
 [#1711]: https://github.com/raiden-network/light-client/pull/1711
 [#1761]: https://github.com/raiden-network/light-client/issues/1761
+[#1787]: https://github.com/raiden-network/light-client/issues/1787
+[#1822]: https://github.com/raiden-network/light-client/pull/1822
 
 ## [0.9.0] - 2020-05-28
 ### Added
--- a/raiden-ts/src/channels/actions.ts
+++ b/raiden-ts/src/channels/actions.ts
@@ -28,6 +28,7 @@ export const tokenMonitored = createAction(
     }),
     t.partial({
       fromBlock: t.number,
+      toBlock: t.number,
     }),
   ]),
 );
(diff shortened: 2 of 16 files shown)


andrevmatos:
Ok, testing this PR by hand, it definitely improved the situation by ensuring we're able to **recover** from lost events upon restarts, which is nice to have/keep, but not essential if we get the events fetching working properly, plus the good performance improvements.

Unfortunately, while testing it, I've seen the event loss happening even on top of this PR. I've tested with a local ethereum node (geth.goerli.ethnodes...) and couldn't reproduce it there, only on Metamask + Infura.

My best bet is that this is caused by how `ethers` (v4 at least) does filters subscriptions: instead of web3js filters logic of registering remotely (which actually doesn't work very well, and I think not at all with Infura), it registers it locally, and upon every new block, it does a `getLogs` under the hood from **previous block to current**. This work well with a local node which is the only one consistently being used, but Infura AFAIK actually load-balance these requests between several nodes, and since it doesn't work with the concept of confirmation blocks, the different Infura nodes could see the mined transaction at different points in time, which then would give weird behaviors when relying on getLogs over only latest block.

Fortunately, if the above is the case, there's a nice way to fix it: implement my own algorithm on top of ether's provider to register filters locally and, on new blocks, perform `getLogs` on a larger range, e.g. `[blockNumber - confirmationBlocks, blockNumber]`, with local/client-side event deduplication. Let's do it.

kelsos:
@andrevmatos How do we proceed with this PR, do we merge it and you do the new algorithm implemented in a new one?

It would make sense for that to go to a new PR since this is already big.

andrevmatos:
Ok, I agree, let's merge this and do a follow-up PR with the additional fixes, they're complementary anyway. Thanks

weilbith:
Oh my review took to long... :joy:

PR #1836, opened 2020-06-30
dapp: extend gitignore for server PR cache

weilbith:
The tool to serve the dApp for a specific PR does store some cache of
the CircleCI artifacts locally. These should be ignored by the SCM.

@andrevmatos it was too lazy to check there could be also other name schemes. Feel free to extend if there are more.


.gitignore
```diff
@@ -1 +1,2 @@
 node_modules
+serve_pr/[0-9]*/
```
kelsos, line 2: this is for running the script directly right?

weilbith: Yes, exactly.

weilbith:
@andrevmatos thanks, good point. Accidentally missed that. Adopted for it.

weilbith:
I'll rebase to merge the fixup commit. Then it is good to get merged.

PR #1866, opened 2020-07-02
cli: add missing event property for payment object

### LINKED ISSUE #1865: CLI `GET /payment/:token/:sender` doesn't replicate python client fully
## Expected Result
- When querying the endpoint `GET /payment/:token/:sender` the python client returns:
```
[
    {
        "event": "EventPaymentReceivedSuccess",
        "amount": "5",
        "initiator": "0x82641569b2062B545431cF6D7F0A418582865ba7",
        "identifier": "1",
        "log_time": "2018-10-30T07:03:52.193",
        "token_address" : "0x5a2d2b9b015b46b8eaff7bffdc5db0051db7439b"
    }
]
```

## Actual Result
The CLI rutrns atm:
```
[
    {
        "amount": "1",
        "identifier": 0,
        "initiator_address": "0x00D384EF74575E97884215e9f39142228c7ACfa8",
        "secret": "",
        "secret_hash": "0x52d05daa7f7ffa5e6283ae523b54fb4de879b6904c9a000217700befc21f5c7f",
        "target_address": "0x00673b5556Db5Fe0DdDB0875bF565a3a4AE51Dcb",
        "token_address": "0x59105441977ecD9d805A4f5b060E34676F50F806"
    }
]
```

weilbith:
Fixes #1865

### DIFF
--- a/raiden-cli/src/utils/payments.ts
+++ b/raiden-cli/src/utils/payments.ts
@@ -2,13 +2,15 @@ import { RaidenTransfer } from 'raiden-ts';
 import { ApiPayment, ApiPaymentEvents } from '../types';
 
 export function transformSdkTransferToApiPayment(transfer: RaidenTransfer): ApiPayment {
+  // TODO: The payment object representation of the API spec is not clear
+  // enough. We need to clarify this and agree on a single representation.
   return {
     event: ApiPaymentEvents[transfer.direction],
     initiator_address: transfer.initiator,
     target_address: transfer.target,
     token_address: transfer.token,
     amount: transfer.value.toString(),
-    identifier: transfer.paymentId.toNumber(),
+    identifier: transfer.paymentId.toString(),
     secret: '', // FIXME: must be first exposed by SDK (#1708)
     secret_hash: transfer.secrethash,
   };
--- a/raiden-cli/src/utils/payments.ts
+++ b/raiden-cli/src/utils/payments.ts
@@ -13,5 +13,6 @@ export function transformSdkTransferToApiPayment(transfer: RaidenTransfer): ApiP
     identifier: transfer.paymentId.toString(),
     secret: '', // FIXME: must be first exposed by SDK (#1708)
     secret_hash: transfer.secrethash,
+    log_time: transfer.changedAt.toISOString(),
   };
 }
--- a/raiden-cli/src/utils/payments.ts
+++ b/raiden-cli/src/utils/payments.ts
@@ -1,8 +1,9 @@
 import { RaidenTransfer } from 'raiden-ts';
-import { ApiPayment } from '../types';
+import { ApiPayment, ApiPaymentEvents } from '../types';
 
 export function transformSdkTransferToApiPayment(transfer: RaidenTransfer): ApiPayment {
   return {
+    event: ApiPaymentEvents[transfer.direction],
     initiator_address: transfer.initiator,
     target_address: transfer.target,
     token_address: transfer.token,
--- a/raiden-cli/src/types.ts
+++ b/raiden-cli/src/types.ts
@@ -43,7 +43,7 @@ export interface ApiPayment {
   target_address: string;
   token_address: string;
   amount: string;
-  identifier: number;
+  identifier: string;
   secret: string;
   secret_hash: string;
 }
--- a/raiden-cli/src/types.ts
+++ b/raiden-cli/src/types.ts
@@ -32,7 +32,13 @@ export interface ApiChannel {
   reveal_timeout: number;
 }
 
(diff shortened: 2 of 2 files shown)


raiden-cli/src/utils/payments.ts
```diff
@@ -1,8 +1,17 @@
 import { RaidenTransfer } from 'raiden-ts';
-import { ApiPayment } from '../types';
+import { ApiPayment, ApiPaymentEvents } from '../types';
+
+function transformTransferDirectionToPaymentEvent(
+  direction: RaidenTransfer['direction'],
+  // @ts-ignore: Function lacks ending return statement and return type does not include 'undefined'
+): ApiPaymentEvents {
+  if (direction === 'sent') return ApiPaymentEvents.sent;
```
andrevmatos, line 8: As there's just 2 possibilities, you could simply use if/else, a conditional expression on return, or possibly (as the `ApiPaymentEvents` is an enum and `direction` is literal), something like `return ApiPaymentEvents[direction];` (not sure if this will work here, but it should with enums+literals.

weilbith: > ApiPaymentEvents[direction];

That looks clean to me. So I can get rid of this whole stupid function at all and code remains slim.

weilbith: :heavy_check_mark:

weilbith:
> Notice their implementation probably differs from the spec you based on; compatibility with their implementation is priority over spec, but we should also take notes on those to update the spec with the current schemas used.

I see your point. But I must admit that this makes me worrying. That I can't follow their specification causes issue. It should be the medium for definition and the implementation just align. Unfortunately it appears to been born in reverse. So the spec is not a spec but was once written do document the already implemented API. And then after a while it got not updated anymore according to the changes in the implementation. This is at least how I imagine it.

Anyways I recognized a bigger issue. So I was wondering why I missed all these things. And the simple answer is that they use different representations for their payment objects. So I first wrote the endpoint to do a payment. According to that I wrote the object conversion to how they defined the response there. Afterwards I added the getter endpoint. But I simply used the same data scheme. But according to their docs they differ. They are similar enough, so I suspect I simply read over it.
This API claims to be RESTful. We can't simply use different object representations. If we go for your approach, and you insist on that, we can write an awkward handling in the CLI to convert the transfers into different kind of payment schemes. In my opinion I would rather fix only what our customer Oliver needs. So we remain with minimal impact. Afterwards we open an issue to discuss with the TransCore team about the spec. I would dislike to be calm and simply redo their slips without talking about it.

Sorry for being so extensive here. But I really fear that if we start to write dirty fixes already now, we won't come back. It will simply work and nobody cares about clean code or good architecture. And such include a well defined API specification.
On the other side I see the good reason that we must act immediately so that Oliver can continue with his demo. So I'm willed to find the compromise I wrote. But this is my very own opinion. Open for any decision.

andrevmatos:
Yes, I feel your pain and agree 100% with you, but I just try to balance the exposed API coherence with our need to have a SP-compatible client sooner than later. @christianbrb ?

andrevmatos:
@weilbith Maybe a compromise (if you want to call that, but which would deviate from spec anyway), would be to provide all properties in our returned object (like both `initiator` and `initiator_address`. Usually, no implementation should mind spare properties, as long as the expected ones are present, so you could make both usages happy with a single schema.

weilbith:
My idea for now was to satisfy Oliver according to [his response](https://github.com/raiden-network/light-client/issues/1865#issuecomment-654103804). Then we can merge this PR and afterwards open an issue for the spec. That sounds like to most simple and straight forward solution to me for the moment. Is that okay for you? :thinking:

weilbith:
According to Oliver [his comment](https://github.com/raiden-network/light-client/issues/1865#issuecomment-654103804) I fulfilled the following requirements by the exposed payment object:
- `identifier` (which is now a string)
- `event` (newly added and artifically derived from direction)
- `initiator_address` (inconsistent naming in spec)
- `target_address` (inconsistent naming in spec)
- `log_time` (newly added, nice-to-have requirement by Olli)

@andrevmatos please validate this PR by these requirements. Thank you! :pray:

weilbith:
@andrevmatos why does the _codecov_ for the SDK complain? I haven't touched it...
I'll do a rebase on `master. I need to merge the fixup commits anyways. I hope that will resolve the problem.

weilbith:
Okay happily _codecov_ is working now again.
@andrevmatos could you please have a quick look and if you agree with my strategy give an approval? Thank you!

weilbith:
[Here](https://github.com/raiden-network/raiden/issues/6368) is the issue I opened for the API spec.

PR #1881, opened 2020-07-07
Feature/sdk check target presence for mediated transfers

### LINKED ISSUE #1835: SDK should autmatically check node availability for mediated transfers
## Description

The CLI struggles to do a mediated transfers because the SDK always returns that the target node is offline. This is actually incorrect and means that the SDK does not know the presence state of the target, since it only monitors the presence of channel partners. The dApp does therefore a manual presence check to fix that issue. But this is not intuitive, the error leads to wasted hours of investigation and everyone new to use the transfer function will run into this issue. Therefore the SDK itself should check if the targets presence is already monitored, else doing so by itself. The latter case will by slightly slower, while the first case is as fast as usual. After all the SDK will become more easy to use.

**Note:** This is kinda urgent for the train demonstration project using the CLI.

## Acceptance criteria
- SDK can to a mediated transfer to a target which presence is not yet monitored without an error message

## Tasks
- [x] @andrevmatos creates the design for the enhancement
- [ ] @weilbith implementation
   - [ ] adopt SDK 
   - [ ] remove now useless/time wasting presence call in the dApp

weilbith:
Fixes #1835

**Short description**

Before validating a route for a mediated transfer the `pathFindServiceEpic` first makes sure that the presence of the routes target is actually known. Before a missing presence lead to the assumption that the target is not available. The workaround so far was to do a manual presence request by the user before doing a transfer.

_Attention:_
The epic for touched for this feature was too huge when starting to work on it. It was kinda part of this PR to first refactor its code to reduce the complexity of each single function and improve the code quality especially in regards of readability. This was necessary to do this extension without loosing the overview. After all would have this changes break the _codeclimate_ rules. I may have taken this refactoring quite far. I recommend to review the commits individually. So the actual main implementation of this feature is within a single commit and is better to understand when looking at it focused. Sorry for the inconvenience. 

**Definition of Done**

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -21,6 +21,7 @@
 - [#1701] Allow parameter decoding to throw and log customized errors
 - [#1701] Add and extend error codes for user parameter validation for open channel
 - [#1711] Add and extend error codes for user parameter validation for transfer
+- [#1835] The presence knowledge for a payment routes target is secured automatically
 
 ### Changed
 - [#837] Changes the action tags from camel to path format. This change affects the event types exposed through the public API.
@@ -55,6 +56,7 @@
 [#1822]: https://github.com/raiden-network/light-client/pull/1822
 [#1830]: https://github.com/raiden-network/light-client/issues/1830
 [#1832]: https://github.com/raiden-network/light-client/pull/1832
+[#1835]: https://github.com/raiden-network/light-client/pull/1835
 [#1848]: https://github.com/raiden-network/light-client/issues/1848
 
 ## [0.9.0] - 2020-05-28
--- a/raiden-ts/src/services/epics.ts
+++ b/raiden-ts/src/services/epics.ts
@@ -39,13 +39,14 @@ import { MessageTypeId, signMessage, createBalanceHash } from '../messages/utils
 import { ChannelState, Channel } from '../channels/state';
 import { assertTx, channelAmounts, groupChannel$ } from '../channels/utils';
 import { Address, decode, Int, Signature, Signed, UInt, isntNil, Hash } from '../utils/types';
-import { isActionOf } from '../utils/actions';
+import { isActionOf, isResponseOf } from '../utils/actions';
 import { encode, losslessParse, losslessStringify } from '../utils/data';
 import { getEventsStream } from '../utils/ethers';
 import { RaidenError, ErrorCodes, assert } from '../utils/error';
 import { pluckDistinct } from '../utils/rx';
 import { Capabilities } from '../constants';
 import { getContractWithSigner } from '../helpers';
+import { matrixPresence } from '../transport/actions';
 
 import {
   iouClear,
@@ -172,16 +173,37 @@ export const pathFindServiceEpic = (
   action$: Observable<RaidenAction>,
   {}: Observable<RaidenState>,
   deps: RaidenEpicDeps,
-): Observable<pathFind.success | pathFind.failure | iouPersist | iouClear> => {
+): Observable<
+  matrixPresence.request | pathFind.success | pathFind.failure | iouPersist | iouClear
+> => {
   return action$.pipe(
     filter(isActionOf(pathFind.request)),
     concatMap((action) =>
       deps.latest$.pipe(
         first(),
-        mergeMap((latest) => getRoute$(action, deps, latest)),
-        withLatestFrom(deps.latest$),
-        mergeMap(([route, latest]) => validateRoute$(action, deps, route, latest)),
-        catchError((err) => of(pathFind.failure(err, action.meta))),
+        mergeMap((latest) => {
+          const { target } = action.meta;
+          let presenceRequest: Observable<matrixPresence.request>;
+          let latestWithTargetInPresences: Observable<Latest>;
+
+          if (target in latest.presences) {
+            presenceRequest = EMPTY;
+            latestWithTargetInPresences = of(latest);
+          } else {
+            presenceRequest = of(matrixPresence.request(undefined, { address: target }));
+            latestWithTargetInPresences = waitForMatrixPresenceResponse$(action$, deps, target);
+          }
+
+          return merge(
+            latestWithTargetInPresences.pipe(
+              mergeMap((latest) => getRoute$(action, deps, latest)),
+              withLatestFrom(deps.latest$),
+              mergeMap(([route, latest]) => validateRoute$(action, deps, route, latest)),
+              catchError((err) => of(pathFind.failure(err, action.meta))),
+            ),
+            presenceRequest,
+          );
+        }),
       ),
     ),
   );
@@ -728,6 +750,26 @@ export const msMonitorNewBPEpic = (
   );
 };
 
+function waitForMatrixPresenceResponse$(
+  action$: Observable<RaidenAction>,
+  deps: RaidenEpicDeps,
+  target: Address,
+): Observable<Latest> {
+  return action$.pipe(
+    filter(
+      isResponseOf<typeof matrixPresence>(matrixPresence, { address: target }),
+    ),
+    take(1),
+    mergeMap((matrixPresenceResponse) => {
+      if (matrixPresence.success.is(matrixPresenceResponse)) {
+        return deps.latest$.pipe(first(({ presences }) => target in presences));
+      } else {
+        throw matrixPresenceResponse.payload;
+      }
+    }),
+  );
+}
+
 function getRoute$(
   action: pathFind.request,
   deps: RaidenEpicDeps,
@@ -735,9 +777,12 @@ function getRoute$(
 ): Observable<{ paths?: Paths; iou: Signed<IOU> | undefined; error?: PathError }> {
   validateRouteTargetAndEventuallyThrow(action, latest);
 
+  const { tokenNetwork, target, value } = action.meta;
+  const { state, presences } = latest;
+
   if (action.payload.paths) {
     return of({ paths: action.payload.paths, iou: undefined });
-  } else if (directTransferIsPossible(action, latest)) {
+  } else if (channelCanRoute(state, presences, tokenNetwork, target, target, value) === true) {
     return of({
       paths: [{ path: [deps.address, action.meta.target], fee: Zero as Int<32> }],
       iou: undefined,
@@ -802,7 +847,7 @@ function validateRouteTargetAndEventuallyThrow(action: pathFind.request, latest:
   if (!Object.values(state.tokens).includes(tokenNetwork))
     throw new RaidenError(ErrorCodes.PFS_UNKNOWN_TOKEN_NETWORK, { tokenNetwork });
 
-  if (!(target in presences) || !presences[target].payload.available)
+  if (!presences[target].payload.available)
     throw new RaidenError(ErrorCodes.PFS_TARGET_OFFLINE, { target });
 
   if (presences[target].payload.caps?.[Capabilities.NO_RECEIVE])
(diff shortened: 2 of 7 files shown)


raiden-ts/tests/e2e/raiden.spec.ts
```diff
@@ -762,13 +762,6 @@ describe('Raiden', () => {
       ).rejects.toThrowError(/Invalid path parameter./i);
     });
 
-    test('target not available', async () => {
-      expect.assertions(1);
-      await expect(raiden.transfer(token, partner, 21)).rejects.toThrowError(
```
andrevmatos, line 767: Hmm, do we wanna remove this test? Isn't it expected to reject if the target is offline?

weilbith: I did already in the recent fixup commit. :+1: 
I also think it is of no value anymore and must not be replaced.

raiden-ts/src/services/epics.ts
```diff
@@ -164,200 +173,37 @@ export const pathFindServiceEpic = (
...
+          if (target in latest.presences) {
+            presenceRequest = EMPTY;
+            latestWithTargetInPresences = of(latest);
+          } else {
+            presenceRequest = of(matrixPresence.request(undefined, { address: target }));
+            latestWithTargetInPresences = waitForMatrixPresenceResponse$(action$, deps, target);
+          }
+
+          return merge(
+            presenceRequest,
```
andrevmatos, line 198: I'd suggest you put the request in the end of `merge`. Just to be safe, in case the synchronous calls would respond to your request before your observable below would be subscribed.

raiden-ts/src/services/epics.ts
```diff
@@ -903,3 +749,297 @@ export const msMonitorNewBPEpic = (
...
+      : undefined,
+  });
+
+  return fromFetch(`${pfs.url}/api/v1/${tokenNetwork}/paths`, {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body,
+  }).pipe(
+    timeout(httpTimeout),
+    map((pfsResponse) => ({ pfsResponse, iou })),
```
andrevmatos, line 987: Nitpick: no need for this map, if you just pass iou to use it below, it's already available in the context

raiden-ts/src/services/epics.ts
```diff
@@ -903,3 +749,297 @@ export const msMonitorNewBPEpic = (
...
+  return getPsfInfo$(action.payload.pfs, latest.config.pfs, deps).pipe(
+    mergeMap((pfs) => getIouForPfs(pfs, tokenNetwork, deps)),
+    mergeMap(({ pfs, iou }) =>
+      requestPfs$(
+        pfs,
+        iou,
+        tokenNetwork,
+        deps.address,
+        target,
+        UInt(32).encode(value),
```
andrevmatos, line 876: Nitpick: can you pass the actual `value`, and encode it inside `requestPfs$`, like the others, just so we get it typeguarded there as well?

raiden-ts/src/services/epics.ts
```diff
@@ -903,3 +749,297 @@ export const msMonitorNewBPEpic = (
...
+  if (paths) {
+    for (const { path, fee } of paths) {
+      const cleanPath = getCleanPath(path, address);
+      const recipient = cleanPath[0];
+      let shouldSelectPath = false;
+      let reasonToNotSelect = '';
+
+      if (invalidatedRecipients.has(recipient)) continue;
+      if (filteredPaths.length === 0) {
+        shouldSelectPath = directTransferIsPossible(action, latest, recipient);
```
andrevmatos, line 905: In this case, if `directTransferIsPossible` returns falsy, you don't get a `reasonToNotSelect`

raiden-ts/src/services/epics.ts
weilbith: I could swear I took that from the old code-base. :laughing:

weilbith: :heavy_check_mark:

raiden-ts/src/services/epics.ts
weilbith: :heavy_check_mark:

raiden-ts/src/services/epics.ts
weilbith: Actually I wanted to do so first. My issue was/is that `value` is of type `t.Branded<BigNumber, UIntB<32>>`. This is a bit awkward to define as function parameter. And since `value` is retrieved from the `meta` object of a `pathFind.request`, I have no chance to use that for the function signature. At least I always get the the error `Property 'meta' does not exist on type 'ActionCreator<"path/find/request"...` But probably my knowledge is just too limited here. Could you help me?

raiden-ts/src/services/epics.ts
weilbith: Oh f***. This is pretty bad on a new level. :see_no_evil:

weilbith: Okay I reverted the introduction of this utility function. The `channelCanRoute` function is not easy to handle and an awkward approach to me. Anyways I'm using it directly again and somehow handle its result. This is now getting the reason again.

weilbith: :heavy_check_mark:

weilbith:
> Also, not sure if the tests cover everything, the removed tests could be considered relevant.
From what I could see, everything seems to be in place, just left some nitpicks comments. I still want to test it live to be sure, but I'm pre-approving as everything seems fine. Thank you

I agree. Actually I felt a bit bad about just removing them. I need think about how to add new cases. So which case actually makes sense to test. We only have one new cases added so far. Maybe you could think of more?
I'll to a practical test soon with the `serve_pr` tool. So I would open a channel from two different light clients to the hub, so that the presence of the according other light client is not known. Then I'll do a transfer to the other light client. Since the dApp doesn't do a manual presence lookup anymore this should be a valid scenario.

weilbith:
Practical test did work for me. :tada: :laughing: 

I will do a rebase to resolve the conflicts with `master` and integrate the fixup commits. But I keep the branch as it is for now. Thereby you can checkout the fixup commits and we can find a solution for the last review conversation. Maybe adding another test case as well. :thinking:

raiden-ts/src/services/epics.ts
andrevmatos: There would be a few ways to extract the type directly ton the action creator or type interface, some of them could require you to use `typeof`, but in this case, it's not needed, `t.Branded<BigNumber, UIntB<32>>` is actually `UInt<32>`, and they're inter-assignable. Actually, since #1870 , the type of `value` should effectively be `UInt<32>`, you can use it in the param of the function.

andrevmatos:
> I agree. Actually I felt a bit bad about just removing them. I need think about how to add new cases. So which case actually makes sense to test. We only have one new cases added so far. Maybe you could think of more?

I think the removed tests need to stay, but you may need to adapt them a bit.
We still want to assert it rejects/throws/errors if the target presence isn't known, so for that you may need to wait a bit after dispatching the `pathFind.request`, verify a `matrixPresence.request` was emitted for target, dispatch a `matrixPresence.failure` and then assert the error also is forwarded to the output of the epic.
Then, we also need the equivalent of above, but where you emit a `matrixPresence.success`, but with `payload.available` as false (i.e. there was a presence found, but the target is offline); this also needs to throw
And finally, the success case for this workflow, where you emit a `matrixPresence.success` with `payload.available` as true, and assert it succeeds.

raiden-ts/src/services/epics.ts
weilbith: :heavy_check_mark:

weilbith:
> We still want to assert it rejects/throws/errors if the target presence isn't known, so for that you may need to wait a bit after dispatching the pathFind.request, verify a matrixPresence.request was emitted for target, dispatch a matrixPresence.failure and then assert the error also is forwarded to the output of the epic.

The part that we check if there gets a matrixPresence.request dispatched if the presence is missing is already tested. I added that back then.

> dispatch a matrixPresence.failure and then assert the error also is forwarded to the output of the epic

Can we actually do that? I mean yes probably we can but I don't know. So the pathFind.request will emit a matrixPresence.request automatically. So how can I inject here and manipulate the response of this request?

> Then, we also need the equivalent of above, but where you emit a matrixPresence.success, but with payload.available as false (i.e. there was a presence found, but the target is offline); this also needs to throw

I thought that was already tested. But I'll check that again.

---

I think I need to figure out how I prevent the default actions as response to the matrixPresence.request and how to mock it. All you test cases refer to that. The test I have atm is pretty simple.

weilbith:
> We still want to assert it rejects/throws/errors if the target presence isn't known, so for that you may need to wait a bit after dispatching the pathFind.request, verify a matrixPresence.request was emitted for target, dispatch a matrixPresence.failure and then assert the error also is forwarded to the output of the epic.

The part that we check if there gets a matrixPresence.request dispatched if the presence is missing is already tested. I added that back then.

> dispatch a matrixPresence.failure and then assert the error also is forwarded to the output of the epic

Can we actually do that? I mean yes probably we can but I don't know. So the pathFind.request will emit a matrixPresence.request automatically. So how can I inject here and manipulate the response of this request?

> Then, we also need the equivalent of above, but where you emit a matrixPresence.success, but with payload.available as false (i.e. there was a presence found, but the target is offline); this also needs to throw

I thought that was already tested. But I'll check that again.

---

I think I need to figure out how I prevent the default actions as response to the matrixPresence.request and how to mock it. All you test cases refer to that. The test I have atm is pretty simple, covering the last case.

weilbith:
We can actually improve a lot of test cases since they don't need to provide the manual `matrixPresence.success` actions anymore. So they can be more focused on what they actually want to test.

andrevmatos:
> Can we actually do that? I mean yes probably we can but I don't know. So the pathFind.request will emit a matrixPresence.request automatically. So how can I inject here and manipulate the response of this request?

Yes, you can. Unfortunately, the `pathFindServiceEpic` tests are still in the old pattern (which uses explicit `action$`), so it could be worth it if you want to convert to the new pattern/mocks, with `const [raiden, partner] = await makeRaidens(2);`, but that would involve rewrite all tests these tests from scratch, although way simpler than they are now. The `pfsCapacityUpdateEpic` tests (right below) are on the new pattern.
If you don't want to refactor these tests, you can still do it in the previous. Basically, you subscribe or `tap` on the output of the epics on the tests you want to write, and if it outputs an action of interest, e.g. `matrixPresence.request`, then you `action$.next` the respective `matrixPresence.success` or `matrixPresence.failure` you want to test, and then wait to see in the output the respective `pathFind.success` or `pathFind.failure`. You "inject" the response actions by `action$.next`ing each upon the respective request out epic's output.

> I thought that was already tested. But I'll check that again.

I think what we test is it target is already in presence, either online or offline; you may rely on that as well, assuming you're sure the epic continues when the respective matrixPresence response goes through.

> I think I need to figure out how I prevent the default actions as response to the matrixPresence.request and how to mock it. All you test cases refer to that. The test I have atm is pretty simple, covering the last case.

On the old tests, you don't need to prevent, they're not going to respond because you subscribe only to the specific epic. On the new tests, though, all epics are always subscribed (the whole/actual state machine), so instead of injecting the response (success/failure), you should instead mock one of the matrix methods which are used by `searchAddressPresence$` function in the `matrixMonitorPresenceEpic` to return empty, fail or succeed with the respective values of interest. If this becomes a pattern, we can have a new mock function to ensure the respective presence response.

> We can actually improve a lot of test cases since they don't need to provide the manual `matrixPresence.success` actions anymore. So they can be more focused on what they actually want to test.

Yes, definitely, specially with the new mocks/test patterns, this is going to help a lot. We need to finish this test refactoring, it's already annoying me, but time, time.......

weilbith:
Okay I kept it simple for the moment and implemented all test cases the old way. There are way more test cases that would need refactoring with the new fixtures. So want to keep that out of this PR here.

I'll now rebase to integrate the fixup commits and resolve merge conflicts.

raiden-ts/CHANGELOG.md
```diff
@@ -230,31 +271,36 @@
...
 - Add getTokenBalance and getTokenInfo to public API.
-- Add network and events$ to the public API.
+- Add network and events\$ to the public API.
 - Add account change and network change monitoring.
 - Add matrix sdk/transport integration.
 - Add protocol message implementation.
 
-
-[Unreleased]: https://github.com/raiden-network/light-client/compare/v0.9.0...HEAD
+[unreleased]: https://github.com/raiden-network/light-client/compare/v0.9.0...HEAD
```
andrevmatos, line 303: Why lowercase?

weilbith: Ah damn I forgot to disable auto-formatting for this file. There a lot of new new-lines etc. Need to revert that.

weilbith: :heavy_check_mark:

weilbith:
I fixed the formatting again.
I'll ignore the changelog reminder for the dApp. It was basically a one line change and removing a test case for this line. I would not really know what to write for this that would be interesting in any kind.

PR #1883, opened 2020-07-09
sdk: try to fix integration tests

andrevmatos:
Fixes #

**Short description**
Nightly integration tests are failing because `pnpm` doesn't seem to be building SDK on `prepare`. So, fix it by explicitly calling the build script, and also make some of the tests failures more descriptive.

**Definition of Done**

1. Just let it run tonight

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -211,7 +211,7 @@ jobs:
       - attach_workspace: *attach_options
       - run:
           name: Restore dependencies
-          command: pnpm install --prefer-frozen-lockfile
+          command: pnpm install --prefer-frozen-lockfile && pnpm run build --filter raiden-ts
       - run:
           name: Run integration tests
           command: source /etc/profile.d/smartcontracts.sh && pnpm run test:integration --filter raiden-ts -- --ci --runInBand --reporters=default --reporters=jest-junit
--- a/raiden-ts/tests/integration/integration.spec.ts
+++ b/raiden-ts/tests/integration/integration.spec.ts
@@ -10,7 +10,7 @@ import { filter } from 'rxjs/operators';
 import { Signer, Wallet } from 'ethers';
 import { RaidenPaths } from 'raiden-ts/services/types';
 
-jest.setTimeout(80000);
+jest.setTimeout(120000);
 
 const ethBalance = '100000000000000000000000';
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -211,7 +211,7 @@ jobs:
       - attach_workspace: *attach_options
       - run:
           name: Restore dependencies
-          command: pnpm install --prefer-frozen-lockfile
+          command: pnpm install --prefer-frozen-lockfile && pnpm run build --filter raiden-ts
```
weilbith, line 214: Why chain the command here? This will just clutter the output of the run and makes it harder to identify issues.

raiden-ts/tests/integration/integration.spec.ts
```diff
@@ -73,13 +73,12 @@ describe('integration', () => {
...
   afterAll((done) => {
-    raiden.stop();
     raiden.events$.pipe(filter((value) => value.type === 'raiden/shutdown')).subscribe(done);
+    raiden.stop();
   });
 
   test('account is funded', async () => {
-    const balance = await raiden.getBalance(raiden.address);
-    expect(balance.eq(bigNumberify(ethBalance))).toBe(true);
+    await expect(raiden.getBalance(raiden.address)).resolves.toEqual(bigNumberify(ethBalance));
```
weilbith, line 81: What was actually the issue here? Are we sure this is formatted? Looks preeeetty long this line and I actually liked the two lines before.

.circleci/config.yml
andrevmatos: The build was to be run in the `install` phase, it's intended for that as well, but for some unknown reason, `pnpm` on `circleCI` called but did nothing to build it. This chained command is just a complement to that, and it makes no sense to have it in its own `run`, and also, no one checks these preparetory steps outputs anyway unless something goes wrong, and it isn't too much output, some ~30L more.

raiden-ts/tests/integration/integration.spec.ts
andrevmatos: This is actually intended to have any thrown exception by `getBalance` as part of the error output, so it makes the error case more readable. The same with switching to a `toEqual` comparison, so it should show up as `rejected instead of resolved` and both values in case of mismatch, instead of just `expected true, got false`.

PR #1887, opened 2020-07-10
sdk: accept falsy paymentId parameters

### LINKED ISSUE #1882: SDK: optional paymentId is only used when being truely in TypeScript
## Steps to Reproduce
1.  Provide `{ paymentId: 0 }` as `options` for a `RaidenService.transfer()` call
2.  Provide `{ paymentId: "0" }` as `options` for a `RaidenService.transfer()` call

## Expected Result
- both calls should work the same

## Actual Result
- the first call will consider the `paymentId` as not given and generates one
- the second call uses the given `paymentId` and parse it as a `UInt<8>`

This is based on the "truelyness" in TypeScript (or JavaScript) of numbers in contrast to a string. So the number `0` is considered `false`, while the string `"0"` is a non-empty string and therefore `true`.

@eorituz thanks for discovering that. As I understood you are now using the string approach to get your zero identifier.

weilbith:
Fixes #1882

@andrevmatos I did not add a test, but still extended the changelog. :stuck_out_tongue_winking_eye: 
After all a user detected this problem for him, so I think this entry is not completely worthless. :wink:

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -12,6 +12,7 @@
 - [#1787] Fix TokenNetwork monitoring losing events
 - [#1830] Fix a nonce race when openining + depositing concurrently
 - [#1848] Fix a Metamask error by retry on deposit
+- [#1882] Fix paymentId gets ignored when being falsie (e.g. `0`)
 
 ### Added
 - [#249] Add withdraw functionality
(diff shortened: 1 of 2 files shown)


weilbith:
Sorry for the amended commit. Just fixed a typo. :see_no_evil:

Issue #1691, opened 2020-06-12
Heavy and stress testing with BF6


christianbrb:
## Description
Currently we have got the unit, e2e, integration and manuel tests. The scenario player tests (#1412) will not be ready soon.

To be confident we would like to add additional (semi-)manual tests how we have done them for Red Eyes (a lot of transfers / load, kill the dApp / SDK during transfers, etc.)

## Acceptance criteria
- Scenario Player runs BF6
- Fixes to run BF6 are implemented
- Documentation and code on how to run BF6 is done

## Tasks
- [ ] @andrevmatos Do research on the various parts needed to split this story

christianbrb:
Maybe we could also use the integration test for this.

christianbrb:
Discussion with @andrevmatos 

It is more important to build a set of scripts to test for Goerli:
- 10 Nodes
- With mediated transfers
- Include PFS delays
- Killing nodes
- CLI Nodes and Python Nodes

Maybe using the scenario player or integration test script to set-up a Raiden Network (just Python Clients).
- [ ] @andrevmatos is checking out the scenario player
     - Scenario Player https://github.com/raiden-network/scenario-player
     - Scenarios: https://github.com/raiden-network/raiden/tree/develop/raiden/tests/scenarios

andrevmatos:
Scenario Player integration depends on extending the CLI compatibility, aka #1692 , so I'm pulling that

christianbrb:
Related #1412

PR #1899, opened 2020-07-14
cli: make GET /payments params optional

### LINKED ISSUE #1880: CLI: read payments endpoint has actually optional filter parameter
## Description

The endpoint has the following scheme **GET** `/api/(version)/payments/(token_address)/(target_address)`. This is also how it is currently implemented in the CLI.  But since we just recently discovered some issues in the spec with the payment endpoints I read again more carefully.  And I discovered that in once sentence they say that this actually is a abbreviation for three endpoints. In fact you can get all payments unfiltered, filtered by token network and filtered by token network and target address. We don't support that. Only the last and most restricted endpoint is implemented.  

## Acceptance criteria
- endpoint to get all payments unfiltered got added
- endpoint to get payments in specific token network got added

## Tasks
- [ ] -"-

andrevmatos:
Fixes #1880 
Part of #1691 

**Short description**
First of a series of PRs to improve CLI to make it compatible with SP, to try to use it for stress tests.

**Definition of Done**

1. Check /payments and /payments/<token> routes behave as expected

### DIFF
--- a/raiden-cli/src/routes/payments.ts
+++ b/raiden-cli/src/routes/payments.ts
@@ -1,9 +1,13 @@
 import { Router, Request, Response, NextFunction } from 'express';
 import { ErrorCodes, RaidenError, RaidenTransfer } from 'raiden-ts';
 import { timer } from 'rxjs';
-import { toArray, takeUntil, first } from 'rxjs/operators';
+import { toArray, takeUntil, first, filter, map } from 'rxjs/operators';
 import { Cli } from '../types';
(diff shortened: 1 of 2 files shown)


andrevmatos:
Yah, `takeUntil(timer(0))` isn't very nice, but it's the tradeoff to be able to have this pattern of emitting only the changed/updated transfer. We may need to think in something better later. Yes, I think it also applies to channels endpoint, I'm proceeding to it soon, to implement the missing functionality there, and will look over it. There should be several PRs which I'll try to split in smaller chunks like this one. Thanks

Issue #1692, opened 2020-06-12
Complete Raiden API endpoint implementations


weilbith:
## Description

This issue acts as extension to #173 and works in addition to #1646.
The target is to have the cli implementation to provide all endpoints the [Raiden API specification](https://raiden-network.readthedocs.io/en/latest/rest_api.html) defines. While all routes are already defined, the according implementations with the SDK are missing.

## Acceptance criteria
- [ ] cli implements all Raiden API endpoints that are missing to completion

## Tasks
- [ ] implement route logic

andrevmatos:
Done in the several PRs above. REST API should be complete. We're doing CLI parameters compliance now.

PR #1903, opened 2020-07-15
cli: implement channel deposit, close, settle and fix some issues with channels interface

andrevmatos:
Part of #1692 

**Short description**
Extend the channel part of the CLI:
- `balance` field contains what is called `capacity` on the SDK, i.e. the amount which can still be transferred on the channel
- `state` contains a different set of strings: `opened`, `closed`, `settled`, so we need to adapt our more descriptive state to these
- all numbes are serialized as strings, including small ones like `id`, `settleTimeout` and `revealTimeout`
- implement `PATCH` method on the channels endpoint:
  - body with `{ "state": "closed"|"settled" }` respectively closes and settles the channel
  - body with `{ "total_deposit": "<number>" }` deposits amout which takes the total deposit to the requested amount
  - body with `{ "total_withdral": "<number>" }` withdraws amount which brings total withdraw to the requested amount

**Definition of Done**

1. Test calling actions on a channel with the respective PATCH on the API

### DIFF
--- a/raiden-cli/src/routes/channels.ts
+++ b/raiden-cli/src/routes/channels.ts
@@ -1,17 +1,13 @@
 import { Router, Request, Response, NextFunction } from 'express';
-import { first } from 'rxjs/operators';
-import { ErrorCodes, RaidenError } from 'raiden-ts';
+import { first, pluck } from 'rxjs/operators';
+import { ErrorCodes, RaidenError, isntNil } from 'raiden-ts';
 import { Cli } from '../types';
 import {
-  validateAddressParameter,
   isInvalidParameterError,
   isTransactionWouldFailError,
+  validateOptionalAddressParameter,
 } from '../utils/validation';
-import {
-  flattenChannelDictionary,
-  transformSdkChannelFormatToApi,
-  filterChannels,
-} from '../utils/channels';
+import { flattenChannelDictionary, transformSdkChannelFormatToApi } from '../utils/channels';
 
 function isConflictError(error: RaidenError): boolean {
   return (
@@ -20,46 +16,39 @@ function isConflictError(error: RaidenError): boolean {
   );
 }
 
-async function getAllChannels(this: Cli, _request: Request, response: Response) {
-  const channelDictionary = await this.raiden.channels$.pipe(first()).toPromise();
-  const channelList = flattenChannelDictionary(channelDictionary);
-  const formattedChannels = channelList.map((channel) => transformSdkChannelFormatToApi(channel));
-  response.json(formattedChannels);
-}
-
-async function getChannelsForToken(this: Cli, request: Request, response: Response) {
-  const channelDictionary = await this.raiden.channels$.pipe(first()).toPromise();
-  const channelList = flattenChannelDictionary(channelDictionary);
-  const filteredChannels = filterChannels(channelList, request.params.tokenAddress);
-  const formattedChannels = filteredChannels.map((channel) =>
-    transformSdkChannelFormatToApi(channel),
-  );
-  response.json(formattedChannels);
-}
-
-async function getChannelsForTokenAndPartner(this: Cli, request: Request, response: Response) {
-  const channelDictionary = await this.raiden.channels$.pipe(first()).toPromise();
-  const channel = channelDictionary[request.params.tokenAddress]?.[request.params.partnerAddress];
-
-  if (channel) {
-    response.json(transformSdkChannelFormatToApi(channel));
+async function getChannels(this: Cli, request: Request, response: Response) {
+  const channelsDict = await this.raiden.channels$.pipe(first()).toPromise();
+  const token: string | undefined = request.params.tokenAddress;
+  const partner: string | undefined = request.params.partnerAddress;
+  if (token && partner) {
+    const channel = channelsDict[token]?.[partner];
+    if (channel) {
+      response.json(transformSdkChannelFormatToApi(channel));
+    } else {
+      response.status(404).send('The channel does not exist');
+    }
   } else {
-    response.status(404).send('The channel does not exist');
+    let channelsList = flattenChannelDictionary(channelsDict);
+    if (token) channelsList = channelsList.filter((channel) => channel.token === token);
+    if (partner) channelsList = channelsList.filter((channel) => channel.token === partner);
+    response.json(channelsList.map(transformSdkChannelFormatToApi));
   }
 }
 
 async function openChannel(this: Cli, request: Request, response: Response, next: NextFunction) {
+  const token: string = request.body.token_address;
+  const partner: string = request.body.partner_address;
   try {
     // TODO: We ignore the provided `reveal_timeout` until #1656 provides
     // a better solution.
-    await this.raiden.openChannel(request.body.token_address, request.body.partner_address, {
+    await this.raiden.openChannel(token, partner, {
       settleTimeout: request.body.settle_timeout,
       deposit: request.body.total_deposit,
     });
-    const channelDictionary = await this.raiden.channels$.pipe(first()).toPromise();
-    const newChannel =
-      channelDictionary[request.body.token_address]?.[request.body.partner_address];
-    response.status(201).json(transformSdkChannelFormatToApi(newChannel));
+    const channel = await this.raiden.channels$
+      .pipe(pluck(token, partner), first(isntNil))
+      .toPromise();
+    response.status(201).json(transformSdkChannelFormatToApi(channel));
   } catch (error) {
     if (isInvalidParameterError(error)) {
       response.status(400).send(error.message);
@@ -76,19 +65,11 @@ async function openChannel(this: Cli, request: Request, response: Response, next
 export function makeChannelsRouter(this: Cli): Router {
   const router = Router();
 
-  router.get('/', getAllChannels.bind(this));
-
   router.get(
-    '/:tokenAddress',
-    validateAddressParameter.bind('tokenAddress'),
-    getChannelsForToken.bind(this),
-  );
-
-  router.get(
-    '/:tokenAddress/:partnerAddress',
-    validateAddressParameter.bind('tokenAddress'),
-    validateAddressParameter.bind('partnerAddress'),
-    getChannelsForTokenAndPartner.bind(this),
+    '/:tokenAddress?/:partnerAddress?',
+    validateOptionalAddressParameter.bind('tokenAddress'),
+    validateOptionalAddressParameter.bind('partnerAddress'),
+    getChannels.bind(this),
   );
 
   router.put('/', openChannel.bind(this));
--- a/raiden-cli/src/routes/channels.ts
+++ b/raiden-cli/src/routes/channels.ts
@@ -1,7 +1,7 @@
 import { Router, Request, Response, NextFunction } from 'express';
 import { first, pluck } from 'rxjs/operators';
-import { ErrorCodes, RaidenError, isntNil } from 'raiden-ts';
-import { Cli } from '../types';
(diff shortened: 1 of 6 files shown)


raiden-cli/src/routes/channels.ts
```diff
@@ -1,21 +1,33 @@
...
+function isConflictError(error: Error): boolean {
   return (
     [ErrorCodes.RDN_UNKNOWN_TOKEN_NETWORK, ErrorCodes.CNL_INVALID_STATE].includes(error.message) ||
     isTransactionWouldFailError(error)
   );
 }
 
+function isInsuficientFundsError(error: { message: string; code?: string | number }): boolean {
+  return (
+    error.code === 'INSUFFICIENT_FUNDS' ||
```
weilbith, line 21: Is this thrown by EthersJS? :thinking:

raiden-cli/src/routes/channels.ts
```diff
@@ -1,21 +1,33 @@
...
+    [
+      ErrorCodes.DTA_INVALID_AMOUNT,
+      ErrorCodes.DTA_INVALID_DEPOSIT,
```
weilbith, line 24: Isn't this more a `Bad Request` error? If I provide an invalid value to deposit this error would be not correct. Because it can't know if I actually have enough funds. This function is used to determine if we should send a `402 - Payment Required` (official HTTP name) error. These error codes are actually part of the definition for `isInvalidParameterError()` (as all `DTA_*` codes are since this is their shared property). This means depending on how you order the if-else conditions at the route, this will result in a different response.

raiden-cli/src/routes/channels.ts
```diff
@@ -52,7 +64,91 @@ async function openChannel(this: Cli, request: Request, response: Response, next
...
+function validateChannelUpdateBody(request: Request, response: Response, next: NextFunction) {
+  const intersec = Object.keys(request.body).filter((k) => allowedUpdateKeys.has(k));
+  if (intersec.length < 1)
+    return response
+      .status(400)
+      .send('one of "state" | "total_deposit" | "total_withdraw" operations required');
+  else if (intersec.length > 1)
+    return response
+      .status(409)
+      .send('more than one of "state" | "total_deposit" | "total_withdraw" requested');
```
weilbith, line 87: Prone to errors: Maybe rather use something like `Array.from(allowedUpdatedKeys.values())`?

raiden-cli/src/routes/channels.ts
```diff
@@ -52,7 +64,91 @@ async function openChannel(this: Cli, request: Request, response: Response, next
...
+  if (
+    request.body.state &&
+    ![ApiChannelState.closed, ApiChannelState.settled].includes(request.body.state)
+  )
+    return response
+      .status(400)
+      .send('invalid "state" requested: must be one of "closed" | "settled"');
```
weilbith, line 94: Prone to errors. Simply use the enumeration values directly. No need to differ/replicate between the check and the related message. Even it might be longer code for the message it is more correct. To be even more correct and avoid any kind of mistakes on future changes I would put the accepted states into a list as you did for the keys above. Then you can declare it once and use it twice, making sure these both lists will never deviate from each other.

I know this are nitpicks, but I think this brings the code to another level of quality. Most of the time you can say it is super unlikely that this happens and is not necessary. I think code should attempt to be as self-contained as possible, deleting the instable human factor. Also does it come with some piece in mind. You can code free without worrying to break anything. This might be not an actual break, but many of these small mistakes lead to bad software over time.

raiden-cli/src/routes/channels.ts
```diff
@@ -52,7 +64,91 @@ async function openChannel(this: Cli, request: Request, response: Response, next
...
+  return next();
+}
+
+async function updateChannel(this: Cli, request: Request, response: Response, next: NextFunction) {
```
weilbith, line 98: This function mixes different kind of abstractions. The higher level purpose of this function is to get the channel at its current state, choose the correct kind of update and return the updated channel. Instead you have to deal with all the details for each kind of update while reading it. I propose to export all the kind of updates (change state, increase deposit, withdraw) into their own functions while each of them will return the updated channel or throw some errors. This function here will feed them with the channel and the query parameter to update  (e.g. the target state or the deposit amount). At the end this function is responsible to evaluate the thrown error and respond to the user in general.

Sorry that I always bug you with splitting your functions. But I think there are many simple rules to achieve qualitative good code. And I barely see good reasons why not to try following them. I mean there are also way more strict rules. Robert C. Martin for examples thinks that a function should be so focused that even error handling (i.e. try-catch) is a single responsibility. I wouldn't go so far. I'm just asking for some basics to keep the code better readable, since I think this is key to good software.

raiden-cli/src/routes/channels.ts
```diff
@@ -52,7 +64,91 @@ async function openChannel(this: Cli, request: Request, response: Response, next
...
+    let channel = await this.raiden.channels$.pipe(first(), pluck(token, partner)).toPromise();
+    if (!channel) return response.status(404).send('channel not found');
+    if (request.body.state) {
+      if (request.body.state === ApiChannelState.closed) {
+        await this.raiden.closeChannel(token, partner);
+        channel = await this.raiden.channels$
+          .pipe(
+            pluck(token, partner),
+            first((channel) =>
+              [ChannelState.closed, ChannelState.settleable, ChannelState.settling].includes(
```
weilbith, line 111: Shouldn't we rather filter for the target state we want to reach?

raiden-cli/src/routes/channels.ts
```diff
@@ -52,7 +64,91 @@ async function openChannel(this: Cli, request: Request, response: Response, next
...
+      } else if (request.body.state === ApiChannelState.settled) {
+        const promise = this.raiden.settleChannel(token, partner);
+        channel = await this.raiden.channels$.pipe(pluck(token, partner), first()).toPromise();
+        await promise;
+      }
+    } else if (request.body.total_deposit) {
+      await this.raiden.depositChannel(
+        token,
+        partner,
+        channel.ownDeposit.sub(request.body.total_deposit).mul(-1),
```
weilbith, line 126: Wait what. Why? Please add a comment if code can't express it. I think this is an important detail.
(same goes for the withdrawl below)

raiden-cli/src/routes/channels.ts
```diff
@@ -52,7 +64,91 @@ async function openChannel(this: Cli, request: Request, response: Response, next
...
+      );
+      channel = await this.raiden.channels$
+        .pipe(
+          pluck(token, partner),
+          first((channel) => channel.ownDeposit.gte(request.body.total_deposit)),
```
weilbith, line 131: Again: why don't filter for the actual target deposit? Could it be that the target doesn't get reached by 100%? What happens if we chain two deposit requests? Couldn't it then be that the second gets the same response as the first one and never the actual update he requested for?
(same goes for the withdrawal below)

raiden-cli/src/routes/channels.ts
```diff
@@ -74,9 +170,13 @@ export function makeChannelsRouter(this: Cli): Router {
 
   router.put('/', openChannel.bind(this));
 
-  router.patch('/:tokenAddress/:partnerAddress', (_request: Request, response: Response) => {
-    response.status(404).send('Not implemented yet');
-  });
+  router.patch(
+    '/:tokenAddress/:partnerAddress',
+    validateOptionalAddressParameter.bind('tokenAddress'),
+    validateOptionalAddressParameter.bind('partnerAddress'),
```
weilbith, line 176: I think here the parameters are not optional. How to get the exact channel else?

raiden-cli/src/utils/channels.ts
```diff
@@ -9,6 +9,25 @@ export function flattenChannelDictionary(channelDict: RaidenChannels): RaidenCha
   );
 }
 
+function transformSdkChannelStateToApi(state: ChannelState): ApiChannelState {
+  let apiState;
+  switch (state) {
+    case ChannelState.open:
+    case ChannelState.closing:
```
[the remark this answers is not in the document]
andrevmatos: Kinda.. we don't receive nor send transfers on it, but some action which is only possible on open channels are still possible here, namely closing it (or attempting to do so, in case the previous attempt failed for some recoverable reason). The python API doesn't have these intermediary states, always assume close call will succeed and I think it is some fatal error or some undefined state if it doesn't (we had to be more reliable here due to interactive signing and the possibility of user rejecting the prompt).

raiden-cli/src/routes/channels.ts
andrevmatos: I don't know where it came from, it was already there, I just kept it. EthersJS error code's usually are numeric, afaik

raiden-cli/src/routes/channels.ts
andrevmatos: Hmm, yes, I was not that concerned with the specifics of the error codes. Unfortunately, we don't have a specific check to fail early on the deposit if not enough funds, instead the transaction just fails on `estimateGas`. I'll remove the errors here and create an issue to fix it on the SDK.

raiden-cli/src/routes/channels.ts
andrevmatos: The `settle` case is an exception for that: settled channels aren't emitted in the `channels$` observable, as they're cleaned right away (actually moved to `oldChannels` on state, but nonetheless not available publicly anymore), so we need to filter to whatever state we can see before that.

raiden-cli/src/routes/channels.ts
andrevmatos: I'll try to do that in code: SDK approach deals with deposit and withdraw amounts, not with totals. This here calculates that by e.g. subtracting `ownDeposit` from `total_deposit`, but since `ownDeposit` is already a `BigNumber`, and `total_deposit` is (expectedly) a `BigNumberish`, it's simpler to subtract in the reverse order and invert then, to avoid importing `bigNumberify` only for a call which there's already a method for.

raiden-cli/src/routes/channels.ts
andrevmatos: Concurrent deposit is kind of not supported. These could cause some boring races which won't put the SDK in an unstable state, but will make one of the requests fail. This is expected, and this `gte` is just to ensure we either succeed (instead of getting stuck) if such race happens, or the user will get the rejected exception. There's no easy way to prevent the user calling it concurrently in the dApp, we would need to introduce locks or some kind of flags on the SDK, and I'd rather avoid that and leave it up to the user to know what they're doing, and possibly concurrent calls may succeed.

raiden-cli/src/routes/channels.ts
andrevmatos: Ops, brain fart. Will fix.

raiden-cli/src/routes/channels.ts
weilbith: Oh okay. But when a state is `settleable` and we call to settle it, will we then actually get a result here? Sounds to me like the request will timeout. :thinking:

andrevmatos: Hmm, you're right.. not settleable, but if it happens that the settle confirmation cleans the channel before this kicks in, it'll hang forever.. I'll fix to also pass if channel isn't present anymore.

raiden-cli/src/routes/channels.ts
weilbith: If you could link the issue here after creation would be great. Just close the conversation then.

andrevmatos: https://github.com/raiden-network/light-client/issues/1905
Already working on it.

Issue #1576, opened 2020-05-25
Provide token network deployment functionality to the SDK


kelsos:
## Description

Currently we can only use token networks that have been deployed by the python client. 
To reach feature parity the light client SDK should also provide this functionality.

We can discuss if it makes sense to further include that in the dApp but I can certainly see the need for it in the SDK/cli part of the light client.

## Acceptance criteria
- Being able to register a new token network using the SDK

## Tasks
- [ ] SDK implementation
- [ ] Provide endpoint in the CLI

PR #1904, opened 2020-07-15
cli: implement /tokens endpoints

andrevmatos:
Part of #1692 

**Short description**
Title.
Only deploying token networks is left, since it's not supported by the SDK yet (#1576)

**Definition of Done**

1. Test the respective endpoints are working properly in the CLI

### DIFF
--- a/raiden-cli/src/routes/tokens.ts
+++ b/raiden-cli/src/routes/tokens.ts
@@ -1,20 +1,52 @@
 import { Router, Request, Response } from 'express';
+import { first } from 'rxjs/operators';
+
 import { Cli } from '../types';
+import { validateAddressParameter } from '../utils/validation';
+
+async function getAllTokens(this: Cli, _request: Request, response: Response) {
+  response.json(await this.raiden.getTokenList());
+}
+
+async function getTokenNetwork(this: Cli, request: Request, response: Response) {
+  // Attention: this also enables monitoring token
+  try {
+    response.json(await this.raiden.monitorToken(request.params.tokenAddress));
+  } catch (error) {
+    response.status(404).send(error.message);
+  }
+}
+
+async function getTokenPartners(this: Cli, request: Request, response: Response) {
+  const token: string = request.params.tokenAddress;
+  const channelsDict = await this.raiden.channels$.pipe(first()).toPromise();
+  const baseUrl = request.baseUrl.replace(/\/\w+$/, '');
+  if (!(token in channelsDict)) response.status(404).send('Unknown token or no partner');
+  else
+    response.json(
+      Object.values(channelsDict[token]).map((channel) => ({
+        partner_address: channel.partner,
+        channel: `${baseUrl}/channels/${token}/${channel.partner}`,
+      })),
+    );
+}
 
 export function makeTokensRouter(this: Cli): Router {
   const router = Router();
 
-  router.get('/', function (_request: Request, response: Response): void {
-    response.status(404).send('Not implemented yet');
-  });
+  router.get('/', getAllTokens.bind(this));
 
-  router.get('/:tokenAddress', (_request: Request, response: Response) => {
-    response.status(404).send('Not implemented yet');
-  });
+  router.get(
+    '/:tokenAddress',
+    validateAddressParameter.bind('tokenAddress'),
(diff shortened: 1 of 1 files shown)


raiden-cli/src/routes/tokens.ts
```diff
@@ -1,20 +1,52 @@
...
+import { validateAddressParameter } from '../utils/validation';
+
+async function getAllTokens(this: Cli, _request: Request, response: Response) {
+  response.json(await this.raiden.getTokenList());
+}
+
+async function getTokenNetwork(this: Cli, request: Request, response: Response) {
+  // Attention: this also enables monitoring token
+  try {
+    response.json(await this.raiden.monitorToken(request.params.tokenAddress));
```
weilbith, line 14: Hmm. I'm not sure if this is an expected side effect. I would maybe just add it when it becomes necessary for the scenario player or so. Until then I would stick with the spec.

raiden-cli/src/routes/tokens.ts
```diff
@@ -1,20 +1,52 @@
...
+  } catch (error) {
+    response.status(404).send(error.message);
+  }
+}
+
+async function getTokenPartners(this: Cli, request: Request, response: Response) {
+  const token: string = request.params.tokenAddress;
+  const channelsDict = await this.raiden.channels$.pipe(first()).toPromise();
+  const baseUrl = request.baseUrl.replace(/\/\w+$/, '');
+  if (!(token in channelsDict)) response.status(404).send('Unknown token or no partner');
```
weilbith, line 24: Having no partner in a network results into an empty list not an error. Please separate these two cases.

raiden-cli/src/routes/tokens.ts
```diff
@@ -1,20 +1,52 @@
...
+    response.json(await this.raiden.monitorToken(request.params.tokenAddress));
+  } catch (error) {
+    response.status(404).send(error.message);
+  }
+}
+
+async function getTokenPartners(this: Cli, request: Request, response: Response) {
+  const token: string = request.params.tokenAddress;
+  const channelsDict = await this.raiden.channels$.pipe(first()).toPromise();
+  const baseUrl = request.baseUrl.replace(/\/\w+$/, '');
```
weilbith, line 23: Are you 100% sure that this works? I thought `baseUrl` just refers to the routers mounting point. So in this case `transfers`. Are you sure this works recursive and get the parent routers mount points?

andrevmatos: Yes, I've tested, it works as intended, it includes the whole mountpoint, including parent's. This is the only way I found besides exporting a `const` somewhere for api base url, and I prefer this way.

raiden-cli/src/routes/tokens.ts
andrevmatos: I am sticking to the spec. The spec has no notion of 'monitored' tokens or not, Raiden python always monitors all tokens. We just don't before as an optimization, but the only method to extract the `tokenNetwork` address is by also monitoring it, and I think it's sensible, monitoring a token don't take us farther from the python behavior, but instead closer.

raiden-cli/src/routes/tokens.ts
```diff
@@ -1,20 +1,52 @@
...
 
-  router.get('/', function (_request: Request, response: Response): void {
-    response.status(404).send('Not implemented yet');
-  });
+  router.get('/', getAllTokens.bind(this));
 
-  router.get('/:tokenAddress', (_request: Request, response: Response) => {
-    response.status(404).send('Not implemented yet');
-  });
+  router.get(
```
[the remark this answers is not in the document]
andrevmatos: These are two different endpoints, which completely different implementations (opposing the other cases where one was a subset of the other). There's no sense IMO to combine these together.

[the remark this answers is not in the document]
weilbith: You are actually right. Two times stupid by me.

raiden-cli/src/routes/tokens.ts
weilbith: Hmm okay. :shrug: 
Thanks that you answered with a good argumentation. :pray:

PR #1906, opened 2020-07-15
Feature/notify channel settled

### LINKED ISSUE #1824: Notification when other party settled channel
## Description

When the other party of a channel has settled, the channel disappears from the channel list. Therefore he has to go to the token withdraw screen of his account to get the tokens back on his main account. This requires advanced knowledge of
the dApp.
A nice solution would be to get a notification of a settled channel. This notification allows then to get forwarded to the withdraw screen. Thereby this event gets not missed by the user and he gets pointed where to withdraw.

## Acceptance criteria
- user receives notification when other party has settled a channel
- notification somehow points the user to how claiming his tokens

## Tasks
- [ ] Need to have the snackbar pr closed before we start more work on notifications
- [ ] Expose event in the SDK
- [ ] In the notification, inform the user where on the account screen to get tokens back

weilbith:
Fixes #1824 

**Short description**
The notifications looks ugly due to the long addresses...

**Definition of Done**

1.
2.


raiden-dapp/src/services/raiden-service.ts
```diff
@@ -373,6 +389,46 @@ export default class RaidenService {
...
+    if (this._subkey) {
+      description +=
+        ' ' + i18n.t('notifications.settlement.success.withdrawal_hint');
+    }
+    await this.store.dispatch('notifications/notify', {
+      title: i18n.t('notifications.settlement.success.title'),
+      description,
+      context: NotificationContext.INFO,
+      importance: NotificationImportance.HIGH,
+    } as NotificationPayload);
```
andrevmatos, line 412: `as` is a type assertion, it may hide errors.. You may leave them here, since this is the pattern already accepted, but this should be improved in the future, either by having these action payloads typesafe (not easy right now with vuex, should be possible on Vue3) or by declaring the payload beforehand with `: NotificationPayload` type, then TS will complain if the object literal doesn't match the expected type.

weilbith: Actually typing get removed on Vuex > 4.? So maybe we should then take some time to make the whole store stuff properly typed. I have there some ideas. :grimacing:

weilbith:
Looks good again. I don't know what the issue was. I will resolve the conflicts and integrate the fixups now. Expect a rebase.

weilbith:
The `codeclimate` actions seem to not work anymore. I never responds. Therefore I'll force the merge.

PR #1907, opened 2020-07-15
sdk: check token balance before trying to deposit

### LINKED ISSUE #1905: Fail earlier and with specific code if there's not enough funds to deposit
## Description
The deposit (be it on channel open or its own action) isn't checked beforehand, and will fail only upon `setTotalDeposit`'s `estimateGas` (which happens only after `approve`).
We should check balance early and fail with specific ErrorCode in this case, to avoid wasting gas and be more predictable when this error is thrown. This will also make it easier to catch and reply with 402 errors on the respective endpoints on CLI.
See https://github.com/raiden-network/light-client/pull/1903#discussion_r454993099

## Acceptance criteria
- 

## Tasks
- [ ]

andrevmatos:
Fixes #1905 

**Short description**
Title.

**Definition of Done**

1. Check you get an early `ErrorCodes.RDN_INSUFFICIENT_BALANCE` error if you try to deposit more than you have of balance

### DIFF
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -618,29 +618,36 @@ const makeDeposit$ = (
 ): Observable<channelDeposit.failure> => {
   if (!deposit?.gt(Zero)) return EMPTY;
 
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  function stopRetryIfRejected(err: any, count: number) {
+  function stopRetryIfRejected(err: { message: string; code?: string | number }, count: number) {
     // 4001 is Metamask's code for rejected/denied signature prompt
     const metamaskRejectedErrorCode = 4001;
     return count >= 10 || err?.code === metamaskRejectedErrorCode;
   }
 
-  return defer(() => tokenContract.functions.allowance(sender, tokenNetworkContract.address)).pipe(
-    mergeMap((allowance) =>
-      allowance.gte(deposit)
-        ? of(true)
-        : retryAsync$(
-            () =>
-              tokenContract.functions.approve(tokenNetworkContract.address, deposit, {
-                nonce: approveNonce,
-              }),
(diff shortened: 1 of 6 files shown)


raiden-ts/src/channels/epics.ts
```diff
@@ -618,29 +618,36 @@ const makeDeposit$ = (
...
+      return retryAsync$(
+        () =>
+          tokenContract.functions.approve(tokenNetworkContract.address, deposit, {
+            nonce: approveNonce,
+          }),
+        (tokenContract.provider as JsonRpcProvider).pollingInterval,
+        stopRetryIfRejected,
+      ).pipe(
+        // if needed, send approveTx and wait/assert it before proceeding
+        assertTx('approve', ErrorCodes.CNL_APPROVE_TRANSACTION_FAILED, { log }),
```
weilbith, line 648: So this has different return types if I understand this correctly? When the allowance is great enough we resolve with a `true`. And else the final operator `assertTx` returns something I don't actually understand. This slightly confuses me. I think after all everything above is meant to complete before the the next `mergeMapTo` operator. Is this correct? Sorry maybe I need another coffee but for the moment the reactive structure is not 100% clear to me. I think a explaining comment by you will already help.

andrevmatos: Yes, notice that this `pipe` applies only to the `approve` observable, so it's going to be an observable which performs the `approve` tx, waits for it to be mined and only then resolves to the `Receipt` (as you can see both by the type of the `assertTx` custom operator and its docstring. If allowance is enough, it returns just a simple `of(true)` instead, which emits and completes immediately. The `mergeMapTo` only proceeds when the previous observable emits, so it's either going to kick in immediately or wait for the `approve`, mine and receipt.
In practice, if you inspect on your text editor the type of the mergeMap's callback, you'll see its return type is `Observable<true | TransactionReceipt>`, which is well descriptive enough IMO.

PR #1909, opened 2020-07-15
cli: implement pending_transfers endpoint and fix /status

andrevmatos:
Part of #1692 

**Short description**
Title.

**Definition of Done**

1. Test `/api/v1/pending_transfers/` endpoints work as per spec

### DIFF
--- a/raiden-cli/src/routes/pendingTransfers.ts
+++ b/raiden-cli/src/routes/pendingTransfers.ts
@@ -1,20 +1,67 @@
 import { Router, Request, Response } from 'express';
+import { timer } from 'rxjs';
+import { filter, takeUntil, map, toArray } from 'rxjs/operators';
+import { RaidenTransfer, isntNil } from 'raiden-ts';
+
 import { Cli } from '../types';
+import { validateOptionalAddressParameter } from '../utils/validation';
+
+enum ApiRole {
+  initiator = 'initiator',
+  mediator = 'mediator',
+  target = 'target,',
+}
+
+function transformSdkToApiPendingTransfer(this: Cli, transfer: RaidenTransfer) {
+  let role: ApiRole;
+  if (transfer.direction === 'sent' && transfer.initiator === this.raiden.address)
+    role = ApiRole.initiator;
+  else if (transfer.direction === 'received' && transfer.target === this.raiden.address)
+    role = ApiRole.target;
+  else if (transfer.direction === 'sent') role = ApiRole.mediator;
+  else return; // filter out duplicated 'received' side of mediated transfers
+  return {
+    channel_identifier: transfer.channelId.toString(),
+    initiator: transfer.initiator,
+    locked_amount: transfer.amount.toString(),
+    payment_identifier: transfer.paymentId.toString(),
+    role,
+    target: transfer.target,
+    token_address: transfer.token,
+    token_network_address: transfer.tokenNetwork,
+    transferred_amount: transfer.amount.sub(transfer.fee).toString(),
+  };
+}
+
+async function getPendingTransfers(this: Cli, request: Request, response: Response) {
+  this.raiden.transfers$
+    .pipe(
+      filter((transfer) => !transfer.completed),
+      filter(
+        (transfer) =>
+          !request.params.tokenAddress || request.params.tokenAddress === transfer.token,
+      ),
+      filter(
+        (transfer) =>
+          !request.params.partnerAddress || request.params.partnerAddress === transfer.partner,
+      ),
+      takeUntil(timer(0)),
+      map(transformSdkToApiPendingTransfer.bind(this)),
+      filter(isntNil),
+      toArray(),
+    )
+    .subscribe((transfers) => response.json(transfers));
+}
 
 export function makePendingTransfersRouter(this: Cli): Router {
   const router = Router();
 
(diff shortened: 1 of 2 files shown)


raiden-cli/src/routes/pendingTransfers.ts
```diff
@@ -1,20 +1,61 @@
...
+import { timer } from 'rxjs';
+import { filter, takeUntil, map, toArray } from 'rxjs/operators';
+import { RaidenTransfer, isntNil } from 'raiden-ts';
+
 import { Cli } from '../types';
+import { validateOptionalAddressParameter } from '../utils/validation';
 
-export function makePendingTransfersRouter(this: Cli): Router {
-  const router = Router();
+enum ApiRole {
```
weilbith, line 9: So far we put this into the `types.ts`. But I must admit that I still miss experience about where to define enumerations, interfaces, types... in TypeScript. Just want to say that we should not start to mix these styles. So either adjust the others to do the same or export this from here.

raiden-cli/src/routes/pendingTransfers.ts
```diff
@@ -1,20 +1,61 @@
...
+  initiator = 'initiator',
+  mediator = 'mediator',
+  target = 'target,',
+}
 
-  router.get('/', function (_request: Request, response: Response): void {
-    response.status(404).send('Not implemented yet');
-  });
+function transformSdkToApiPendingTransfer(this: Cli, transfer: RaidenTransfer) {
```
weilbith, line 15: Similar to the comment above. So far these conversations where exported into a separate module. The intention was to keep the routes slim and focused on retrieving data from the SDK or trigger events. Such details just clutter it. 
But I actually don't have too hard feelings about that. Just that we should attempt to develop a uniform code-base. So again: either do it like we do for the other resources already or adjust all other resources.

raiden-cli/src/routes/pendingTransfers.ts
```diff
@@ -1,20 +1,61 @@
...
+  let role: ApiRole;
+  if (transfer.direction === 'sent' && transfer.initiator === this.raiden.address)
+    role = ApiRole.initiator;
+  else if (transfer.direction === 'received' && transfer.target === this.raiden.address)
+    role = ApiRole.target;
+  else if (transfer.direction === 'sent') role = ApiRole.mediator;
+  else return; // filter out duplicated 'received' side of mediated transfers
+  return {
```
weilbith, line 23: Usually we had an interface defined for these objects that get shared via the API. Either remove the old ones or extend here please.

raiden-cli/src/routes/pendingTransfers.ts
```diff
@@ -1,20 +1,61 @@
...
 
-  router.get('/:tokenAddress', (_request: Request, response: Response) => {
-    response.status(404).send('Not implemented yet');
-  });
+async function getPendingTransfers(this: Cli, request: Request, response: Response) {
+  this.raiden.transfers$
+    .pipe(
+      filter((t) => !t.completed),
+      filter((t) => !request.params.tokenAddress || request.params.tokenAddress === t.token),
+      filter((t) => !request.params.partnerAddress || request.params.partnerAddress === t.partner),
```
weilbith, line 41: Nitpick: Why? Does this lead to ugly code formatting or what is the intention? Code gets incomparable often read then written. But I'm willed to accept it here... Just want to achieve some awareness for such tiny things. Event code navigation, highlighting of references, renaming etc are just bad with single character variable names.

raiden-cli/src/routes/pendingTransfers.ts
andrevmatos: So I think the only types/enums which need to be exported in its own modules are those which may be reusable. We prefer to have single-use types local on where they're needed. The same goes for helper/utility functions: if they're generic and/or used multiple times, they can be split in a specific module, but IMO it's more readable to not need to be jumping back and forth on other files if a function/type/enum is only used and maybe only relevant in that context, than to split it by "type"... this is more of a Domain Driven Development, which we tend to follow on the light-client, as per @christianbrb 's lead. These types and functions belong to the same semantic domain. If it becomes too big, we can split, but the way we did with the epics, having them on separate files but domain-localized in a private module instead of the public/global ones.
I didn't want to refactor the others here, since I'm trying to do small PRs, but I'll make a follow up PR to adjust those.

raiden-cli/src/routes/pendingTransfers.ts
andrevmatos: Such details are utility functions only useful to these local routes, not relevant to anyone else, so I don't think it gets cluttered by them. We should decide on this and do this refactoring on a separate PR.

raiden-cli/src/routes/pendingTransfers.ts
andrevmatos: Fixed.

raiden-cli/src/routes/pendingTransfers.ts
weilbith: We agreed on the D3 approach in the daily call. :+1:

raiden-cli/src/routes/pendingTransfers.ts
weilbith: Checkout conversations above.

PR #1910, opened 2020-07-15
sdk: option to allow mint to specific address with mintFor

andrevmatos:
Needed for `_testing`'s mint functionality of #1692 

**Short description**
Add an optional param to specificy the beneficiary of the `mint` transaction.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -1018,7 +1018,7 @@ export class Raiden {
    * Throws an error, if
    * <ol>
    *  <li>Executed on main net</li>
-   *  <li>`token` is not a valid address</li>
+   *  <li>`token` or `options.to` is not a valid address</li>
    *  <li>Token could not be minted</li>
    * </ol>
    *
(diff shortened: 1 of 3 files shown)


raiden-ts/src/raiden.ts
```diff
@@ -1043,14 +1044,21 @@ export class Raiden {
...
+      Address.is(beneficiary),
+      [ErrorCodes.DTA_INVALID_ADDRESS, { beneficiary }],
+      this.log.info,
+    );
+
     const receipt = await callAndWaitMined(
       customTokenContract,
-      'mint',
-      [decode(UInt(32), amount)],
+      'mintFor',
```
weilbith, line 1060: Hmm the docs say that the token contract can either define `mint(address, uint256)`, `mintFor(uint256, address)` or `increaseSupply(uint256, address)`. The last seem to be not relevant. But from the first two: what is the difference in behavior except the signature? Is there definition exclusive? Should we maybe check for both? Or is `mintFor` the only one that does what we want now?

raiden-ts/src/raiden.ts
```diff
@@ -1043,14 +1044,21 @@ export class Raiden {
...
+      [decode(UInt(32), amount, ErrorCodes.DTA_INVALID_AMOUNT), beneficiary],
```
weilbith, line 1061: Please pull out the decoding. Must not be mixed in here. I think we don't do so elsewhere. Anyways I would say this is a bad habit. :wink:

raiden-ts/src/raiden.ts
andrevmatos: We don't support all these interfaces. If you look into the source of the CustomToken contract, you'll see it supports just 2 minting methods: `mint(uint256)`, which mints for self, and `mintFor(uint256, address)`, which allows minting for a 3rd party. This is a `_testing` function, used only on our testnets, and these are always deployed with this contract. It'd be hard for the SDK to support multiple contracts interfaces, so we went with what is actually used.

PR #1911, opened 2020-07-15
cli: implement /_testing endpoint for minting tokens

andrevmatos:
Based on top of #1910 
Part of #1692 

**Short description**
Implements minting tokens from the CLI API.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-cli/src/routes/testing.ts
+++ b/raiden-cli/src/routes/testing.ts
@@ -1,12 +1,31 @@
-import { Router, Request, Response } from 'express';
+import { Router, Request, Response, NextFunction } from 'express';
+import { ErrorCodes } from 'raiden-ts';
+import { validateAddressParameter } from '../utils/validation';
 import { Cli } from '../types';
+
+async function mintTokens(this: Cli, request: Request, response: Response, next: NextFunction) {
+  try {
+    const transactionHash = await this.raiden.mint(
+      request.params.tokenAddress,
+      request.body.value,
+      { to: request.body.to },
+    );
+    response.json({ transaction_hash: transactionHash });
+  } catch (error) {
+    if ([ErrorCodes.DTA_INVALID_ADDRESS, ErrorCodes.DTA_INVALID_AMOUNT].includes(error.message))
+      response.status(400).send(error.message);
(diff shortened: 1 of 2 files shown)


raiden-cli/src/routes/testing.ts
```diff
@@ -1,12 +1,30 @@
...
+async function mintTokens(this: Cli, request: Request, response: Response, next: NextFunction) {
+  try {
+    response.json({
+      transaction_hash: await this.raiden.mint(request.params.tokenAddress, request.body.value, {
+        to: request.body.to,
+      }),
+    });
+  } catch (error) {
+    if ([ErrorCodes.DTA_INVALID_ADDRESS, ErrorCodes.DTA_INVALID_AMOUNT].includes(error.message))
+      response.status(400).send('Invalid "value" or "to" fields.');
```
weilbith, line 15: Usually we directly forward the error message here. So this if-else is basically in first place to determine the error code to use. Else we should avoid to attempt introducing new error messages and simply use the probably way better messages by the SDK.

raiden-cli/src/routes/testing.ts
```diff
@@ -1,12 +1,30 @@
...
+import { validateAddressParameter } from '../utils/validation';
 import { Cli } from '../types';
 
+async function mintTokens(this: Cli, request: Request, response: Response, next: NextFunction) {
+  try {
+    response.json({
+      transaction_hash: await this.raiden.mint(request.params.tokenAddress, request.body.value, {
+        to: request.body.to,
+      }),
+    });
```
weilbith, line 12: Nitpick: For me the idea of this endpoint is to mint the tokens (`POST` as active request) and then return the result. This is how my brain works: declarative. So I would first await the minting and then respond. Makes it easy to follow the code. Personally I would even "construct" the response object separately but I wouldn't push for it.

andrevmatos: Done.

raiden-cli/src/routes/testing.ts
andrevmatos: Done.

PR #1913, opened 2020-07-15
sdk,cli: expose contracts info in SDK and provide /contracts endpoint in CLI

andrevmatos:
Part of #1692 

**Short description**
Implements `/contracts` endpoint. Still undocumented on spec, but offered by python API.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-cli/src/routes/api.v1.ts
+++ b/raiden-cli/src/routes/api.v1.ts
@@ -19,15 +19,24 @@ export function makeApiV1Router(this: Cli): Router {
   router.use('/testing', makeTestingRouter.call(this));
 
   router.get('/version', (_request: Request, response: Response) => {
+    response.json({ version: Raiden.version });
+  });
+
+  router.get('/contracts', (_request: Request, response: Response) => {
+    const contracts = this.raiden.contractsInfo;
     response.json({
-      version: Raiden.version,
+      contracts_version: Raiden.contractVersion,
+      token_network_registry_address: contracts.TokenNetworkRegistry.address,
+      secret_registry_address: contracts.SecretRegistry.address,
+      service_registry_address: contracts.ServiceRegistry.address,
+      user_deposit_address: contracts.UserDeposit.address,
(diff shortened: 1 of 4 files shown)


raiden-cli/src/routes/api.v1.ts
```diff
@@ -19,15 +19,24 @@ export function makeApiV1Router(this: Cli): Router {
...
 
   router.get('/version', (_request: Request, response: Response) => {
+    response.json({ version: Raiden.version });
+  });
+
+  router.get('/contracts', (_request: Request, response: Response) => {
+    const contracts = this.raiden.contractsInfo;
     response.json({
-      version: Raiden.version,
+      contracts_version: Raiden.contractVersion,
```
weilbith, line 28: Similar to an earlier issue. We should maybe consider to remove old interface definitions for API objects. Or if you still see an advantage of them add one for this route endpoint as well.

raiden-cli/src/routes/api.v1.ts
```diff
@@ -19,15 +19,24 @@ export function makeApiV1Router(this: Cli): Router {
...
+      monitoring_service_address: contracts.MonitoringService.address,
+      one_to_n_address: '',
     });
   });
 
   router.get('/address', (_request: Request, response: Response) => {
-    response.json({
-      our_address: this.raiden.address,
-    });
+    response.json({ our_address: this.raiden.address });
```
weilbith, line 39: Has formatting changed? I hate it when formatting is not deterministic. :expressionless:

andrevmatos: It didn't change, but prettier's recommended rules allow small objects to be either declared in one line or split lines. I just wanted to reduce the line counter for something so simple. If it becomes bigger, with more properties making it cross 99char line limit, then it must be broken.

raiden-cli/src/routes/api.v1.ts
andrevmatos: As I replied there, let's decide on this and do it on a separate PR.

Issue #237, opened 2019-08-05
Assess auto vs interactive request for delayed actions


andrevmatos:
## Description

In Raiden, some operations become only available after some (considerable amount of) time after the action which triggered them, currently:
- `channelSettle` (on-chain): it becomes available only `settleTimeout` (default=`500` ~= 2h) blocks after `channelClosed`, actually retrieve the left tokens back to user's account
- `transferExpire` (off-chain): a **lock**ed transfer may only be expired (sending a new set of messages starting with `LockExpired`),  effectivelly retrieves the tokens which were locked to first try the transfer, `expiration` (default = `DEFAULT_REVEAL_TIMEOUT * 2 = 100` ~= 22min) blocks after transfer creation, retrieve tokens locked in a failed transfer, returning them to capacity

These actions have a minimum point in time after which they can be requested, but they aren't required to be performed immediately (actually not even required by protocol to be performed at all), but they're usually interesting for the user to be performed (i.e. users have an incentive to do that).

Despite that, from a UX perspective, it may be quite confusing for the user to see unrequested prompts popping up in their screen minutes, hours or even days (e.g. when they come back online) after the state changes that first caused them.

### Current state is:
- settleChannel is done only on-demand (interactively), through [Raiden.settleChannel](https://github.com/raiden-network/light-client/blob/dfac0ca77f0e86a483a59e61547c89b0d53d34a5/raiden/src/raiden.ts#L592) public method. dApp is expected to notify users somehow when they have **settleable** channels, so they can perform the settling
- transferExpire (as per #232) is done automatically, as its timeout is shorter (possibly customizable in the future) and the transfer status overlay should stay on-screen while transfer is still pending. Notice this will only take that long on the full unhappy case, which requires signing the `LockExpired` message after the lock expires.

### Proposal is to make it configurable:
- `autoSettle`: automatically calls `channelSettle` when a channel becomes settleable, or else through `settleChannel` method (interactively).
- `autoExpire`: automatically performs `transferExpire` when a pending transfer lock expires, or else through a `expireTransfer` method (interactively).

### Where we can envision a few environments and its expected defaults/best values:
1. **m2m/nodeJS/local account/privatekey** (like scripts and non-interactive systems built on top of the SDK): both autoSettle and autoExpire set, auto-managed case
2. **browser/Metamask with subkey** (#172) for signing, ETH/tokens transferred to the subkey account: same as local account case
[...]

PR #1919, opened 2020-07-16
cli: add naive connections manager endpoints

andrevmatos:
Part of #1692 

**Short description**
- GET will return the connected TokenNetworks; as it's a dummy implementation, it has no notion of allocated funds, so the funds and deposits are just the sum of channel's deposits
- PUT works by connecting with a single hardcoded hub (currently, `hub.raiden.eth`, as suggested by dApp)
- DELETE does work as intended by closing all channels in a tokenNetwork, although auto-settling is still pending

**Definition of Done**

1.

### DIFF
--- a/raiden-cli/src/routes/connections.ts
+++ b/raiden-cli/src/routes/connections.ts
@@ -1,20 +1,133 @@
-import { Router, Request, Response } from 'express';
+import { Router, Request, Response, NextFunction } from 'express';
+import { first } from 'rxjs/operators';
+import { ChannelState, isntNil } from 'raiden-ts';
+
+import {
+  validateAddressParameter,
+  isConflictError,
+  isInsuficientFundsError,
+  isInvalidParameterError,
+} from '../utils/validation';
 import { Cli } from '../types';
+
+export interface ApiConnection {
+  channels: string;
+  sum_deposits: string;
+  funds: string;
+}
+
+export interface ApiTokenConnections {
+  [token: string]: ApiConnection;
+}
+
+async function getConnections(this: Cli, _request: Request, response: Response) {
+  const channelsDict = await this.raiden.channels$.pipe(first()).toPromise();
+  const connections = Object.entries(channelsDict).reduce<ApiTokenConnections>(
+    (connections, [token, partnerChannel]) => ({
+      ...connections,
+      [token]: Object.values(partnerChannel).reduce<ApiConnection>(
+        ({ channels, funds }, { ownDeposit, ownWithdraw }) => {
+          const deposits = ownDeposit.sub(ownWithdraw).add(funds).toString();
+          return {
+            channels: (+channels + 1).toString(),
+            sum_deposits: deposits,
+            funds: deposits,
+          };
+        },
+        { channels: '0', sum_deposits: '0', funds: '0' },
+      ),
+    }),
+    {},
+  );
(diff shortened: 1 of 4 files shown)


raiden-cli/src/routes/connections.ts
```diff
@@ -1,20 +1,133 @@
...
+async function getConnections(this: Cli, _request: Request, response: Response) {
+  const channelsDict = await this.raiden.channels$.pipe(first()).toPromise();
+  const connections = Object.entries(channelsDict).reduce<ApiTokenConnections>(
+    (connections, [token, partnerChannel]) => ({
+      ...connections,
+      [token]: Object.values(partnerChannel).reduce<ApiConnection>(
+        ({ channels, funds }, { ownDeposit, ownWithdraw }) => {
+          const deposits = ownDeposit.sub(ownWithdraw).add(funds).toString();
+          return {
+            channels: (+channels + 1).toString(),
```
weilbith, line 32: How is this syntax working?

raiden-cli/src/routes/connections.ts
```diff
@@ -1,20 +1,133 @@
...
+    if (isInvalidParameterError(error)) response.status(400).send(error.message);
+    else if (isInsuficientFundsError(error)) response.status(402).send(error.message);
+    else if (isConflictError(error)) response.status(409).send(error.message);
+    else next(error);
+  }
+}
+
+/**
+ * Closes all closeable channels in a token network
+ * TODO: implement auto-settle (see discussion at #237)
```
weilbith, line 90: Will this be actually a TODO of the CLI or wont it happen in the SDK exclusively?

raiden-cli/src/routes/connections.ts
```diff
@@ -1,20 +1,133 @@
...
+  );
+  const promises: Promise<string | undefined>[] = [];
+  for (let i = 0; i < closeableChannels.length; i++) {
+    // if more than one channel to close, wait 1s between each call to avoid nonce races
+    if (i > 0) await new Promise((resolve) => setTimeout(resolve, 1000));
+    const channel = closeableChannels[i];
+    promises.push(
+      this.raiden.closeChannel(token, channel.partner).then(
+        () => channel.partner,
+        () => undefined,
```
weilbith, line 107: So we simply ignore any closing errors? :thinking:

raiden-cli/src/routes/connections.ts
andrevmatos: The unary operator `+` is able to convert numeric strings into numbers, then it's incremented and converted back `toString` (since the interface expects a numeric string. Since we control `channels` from scratch, and knows it's small, it's safe to do this without checks.

raiden-cli/src/routes/connections.ts
andrevmatos: We could do it either implemented on the CLI or discuss #237 and possibly decide on doing it in the SDK through some `autoSettle` option. Until we decide, it remains a TODO and the users need to settle actively.

raiden-cli/src/routes/connections.ts
andrevmatos: They are logged, and the errored channels don't go to the output. The endpoint doesn't specify http errors for closing errors, so it remains on user's hand to be aware that the channel didn't close and possibly try later.

PR #1922, opened 2020-07-17
sdk: Add option to fetch all contracts addresses from UserDeposit address alone

### LINKED ISSUE #703: Fetch all contracts address from a single entry point
## Description

https://github.com/raiden-network/raiden/pull/5430#issuecomment-562858144 provided a very nice overview of how all contracts in Raiden Network are linked. 
Currently, we require 3 contracts addresses & deployment blocks to be passed as parameter to create in the form of `ContractsInfo` interface: `TokenNetworkRegistry`, `ServiceRegistry` & `UserDeposit`.
We should also implement it in the SDK, to reduce the number of parameters needed (specially in `Raiden.create`) and reduce the surface for errors, and fetch (and validate) the other addresses from the on-chain pointers.

## Acceptance criteria
- 

## Tasks
- [ ]

andrevmatos:
Fixes #703

**Short description**
Supports using `UserDeposit` contract's address as entrypoint to fetch all other addresses from it.
Deployment block for all contracts is set to oldest TokenNetwork registration event, as it's used only as lowerbound for events fetching.
Also, adds `OneToN` contract to `ContractsInfo` type, which completes it with all relevant deployed contracts.
The previous default way for feeding `contractsInfo` from the JSON generated deployment files is kept for backwards compatibility and efficiency, since it doesn't require the associated network requests.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/helpers.ts
+++ b/raiden-ts/src/helpers.ts
@@ -27,6 +27,9 @@ import ropstenServicesDeploy from './deployment/deployment_services_ropsten.json
 import rinkebyServicesDeploy from './deployment/deployment_services_rinkeby.json';
 import goerliServicesDeploy from './deployment/deployment_services_goerli.json';
 import mainnetServicesDeploy from './deployment/deployment_services_mainnet.json';
+import { UserDepositFactory } from './contracts/UserDepositFactory';
+import { MonitoringServiceFactory } from './contracts/MonitoringServiceFactory';
+import { TokenNetworkRegistryFactory } from './contracts/TokenNetworkRegistryFactory';
 
 /**
  * Returns contract information depending on the passed [[Network]]. Currently, only
@@ -341,3 +344,51 @@ export const isValidUrl = (url: string): boolean => {
       : /^(?:(http|https):\/\/)?([^\s\/$.?#&"']+\.)*[^\s\/$?#&"']+(?:(\d+))*$/;
   return regex.test(url);
 };
+
+/**
+ * Construct entire ContractsInfo using UserDeposit contract address as entrypoint
+ *
+ * @param provider - Ethers provider to use to fetch contracts data
+ * @param userDeposit - UserDeposit contract address as entrypoint
+ * @returns contracts info, with blockNumber as block of first registered tokenNetwork
+ */
+export async function fetchContractsInfo(
+  provider: JsonRpcProvider,
+  userDeposit: Address,
+): Promise<ContractsInfo> {
+  const userDepositContract = UserDepositFactory.connect(userDeposit, provider);
+
+  const oneToN = (await userDepositContract.one_to_n_address()) as Address;
+  const monitoringService = (await userDepositContract.msc_address()) as Address;
+  const monitoringServiceContract = MonitoringServiceFactory.connect(monitoringService, provider);
+
+  const tokenNetworkRegistry = (await monitoringServiceContract.token_network_registry()) as Address;
+  const tokenNetworkRegistryContract = TokenNetworkRegistryFactory.connect(
+    tokenNetworkRegistry,
+    provider,
+  );
+
+  const secretRegistry = (await tokenNetworkRegistryContract.secret_registry_address()) as Address;
+  const serviceRegistry = (await monitoringServiceContract.service_registry()) as Address;
+
+  const logs = await provider.getLogs({
+    ...tokenNetworkRegistryContract.filters.TokenNetworkCreated(null, null),
+    fromBlock: 1,
+    toBlock: 'latest',
+  });
+  let firstBlock = 0;
+  for (const { blockNumber } of logs) {
+    if (blockNumber) {
+      firstBlock = firstBlock ? Math.min(firstBlock, blockNumber) : blockNumber;
+    }
+  }
+
+  return {
+    TokenNetworkRegistry: { address: tokenNetworkRegistry, block_number: firstBlock },
+    ServiceRegistry: { address: serviceRegistry, block_number: firstBlock },
+    UserDeposit: { address: userDeposit, block_number: firstBlock },
+    SecretRegistry: { address: secretRegistry, block_number: firstBlock },
+    MonitoringService: { address: monitoringService, block_number: firstBlock },
+    OneToN: { address: oneToN, block_number: firstBlock },
+  };
+}
--- a/raiden-ts/src/helpers.ts
+++ b/raiden-ts/src/helpers.ts
@@ -358,7 +358,6 @@ export async function fetchContractsInfo(
 ): Promise<ContractsInfo> {
   const userDepositContract = UserDepositFactory.connect(userDeposit, provider);
 
-  const oneToN = (await userDepositContract.one_to_n_address()) as Address;
   const monitoringService = (await userDepositContract.msc_address()) as Address;
   const monitoringServiceContract = MonitoringServiceFactory.connect(monitoringService, provider);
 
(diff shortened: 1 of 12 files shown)


raiden-ts/src/helpers.ts
```diff
@@ -341,3 +344,51 @@ export const isValidUrl = (url: string): boolean => {
...
+ */
+export async function fetchContractsInfo(
+  provider: JsonRpcProvider,
+  userDeposit: Address,
+): Promise<ContractsInfo> {
+  const userDepositContract = UserDepositFactory.connect(userDeposit, provider);
+
+  const oneToN = (await userDepositContract.one_to_n_address()) as Address;
+  const monitoringService = (await userDepositContract.msc_address()) as Address;
+  const monitoringServiceContract = MonitoringServiceFactory.connect(monitoringService, provider);
```
weilbith, line 363: I have some issues with the details of naming here. First I thought it is okay to have the implication that all these variables will include the address. At least since the typehint is telling you that it is an address. But then there is the need to differ between the the address and the contract. The solution to only add an appendix to one of them is unsatisfying. Since I appreciate that you split this into two clear readable lines, even if the address retrieval is only necessary for the contract instantiation. But if you feel like this is absolutely horrible to append to all address variables `*Address` and want to stick to the implicit address property, then I would put this address retrieval inline and let the formatter split the code.

raiden-ts/src/helpers.ts
```diff
@@ -341,3 +344,51 @@ export const isValidUrl = (url: string): boolean => {
...
+ * @param userDeposit - UserDeposit contract address as entrypoint
+ * @returns contracts info, with blockNumber as block of first registered tokenNetwork
+ */
+export async function fetchContractsInfo(
+  provider: JsonRpcProvider,
+  userDeposit: Address,
+): Promise<ContractsInfo> {
+  const userDepositContract = UserDepositFactory.connect(userDeposit, provider);
+
+  const oneToN = (await userDepositContract.one_to_n_address()) as Address;
```
weilbith, line 361: You retrieve this contract pretty early in the function. It gets not used before inserted into the return value. I would think this can make sense if the "order" (also if there is no such forced order) of the return-object has this contrast as first. But since it is not, I would rather retrieve this information more close to where it gets used.

raiden-ts/src/helpers.ts
```diff
@@ -341,3 +344,51 @@ export const isValidUrl = (url: string): boolean => {
...
+  const tokenNetworkRegistryContract = TokenNetworkRegistryFactory.connect(
+    tokenNetworkRegistry,
+    provider,
+  );
+
+  const secretRegistry = (await tokenNetworkRegistryContract.secret_registry_address()) as Address;
+  const serviceRegistry = (await monitoringServiceContract.service_registry()) as Address;
+
+  const logs = await provider.getLogs({
+    ...tokenNetworkRegistryContract.filters.TokenNetworkCreated(null, null),
```
weilbith, line 375: Why does actually the token network contracts define the first block? Couldn't the user deposit contract be older and still includes interesting information to fetch even before the first network got deployed? :thinking: 
Sorry if this is a stupid question. But I know most-likely nothing about the contract architecture so far. :see_no_evil:

raiden-ts/src/helpers.ts
```diff
@@ -341,3 +344,51 @@ export const isValidUrl = (url: string): boolean => {
...
+    fromBlock: 1,
+    toBlock: 'latest',
+  });
+  let firstBlock = 0;
+  for (const { blockNumber } of logs) {
+    if (blockNumber) {
+      firstBlock = firstBlock ? Math.min(firstBlock, blockNumber) : blockNumber;
+    }
+  }
```
weilbith, line 384: The code looks more complicated than what it actually does. Looks pretty inefficient (code-wise). Couldn't you simply sort the logs by `blockNumber` and take the first/last list entry (depending on order) or `0` as default? This would be about 2 lines of code (before formatting) and much easier to read since it is more declarative (at least for me).

raiden-ts/src/raiden.ts
```diff
@@ -319,7 +320,7 @@ export class Raiden {
       | RaidenState
       | { storage: Storage; state?: RaidenState | unknown }
       | unknown,
-    contracts?: ContractsInfo,
+    contractsOrUserDepositAddress?: ContractsInfo | string,
```
weilbith, line 323: Oh I hate such code so damn much. Actually getting goose bumps while reading it. But since TypeScript doesn't support function overloading I guess we have to do it like this to maintain backwards compatibility.

raiden-ts/src/raiden.ts
```diff
@@ -338,8 +339,16 @@ export class Raiden {
...
 
     // if no ContractsInfo, try to populate from defaults
-    if (!contracts) {
+    let contracts;
+    if (!contractsOrUserDepositAddress) {
       contracts = getContracts(network);
+    } else if (typeof contractsOrUserDepositAddress === 'string') {
+      // if an Address is provided, use it as UserDeposit contract address entrypoint and fetch
+      // all contracts from there
+      assert(Address.is(contractsOrUserDepositAddress), ErrorCodes.DTA_INVALID_ADDRESS);
```
weilbith, line 348: I'm too lazy to check this right now: would this lead to a generic error message as from the `errors.json` only or does it get extended with the name of the variable? Else this functions takes quite a lot of parameter. Telling him which one was wrong would be quite helpful. I think we do this at many other places.

raiden-ts/src/helpers.ts
andrevmatos: This pattern is followed all through codebase, because we use the addresses much more often than the contracts (specially for `token` and `tokenNetwork`), which would make it very wasteful to rename everything, and also they `Address` is already implied by the variable type hint, but for the `Contract`, we need to deduplicate the name, and also the name of the contract is e.g. just `MonitoringService` (which inherits `Contract`, from `TypeChain`), which wouldn't tell you it's a contract.
Also, this isn't used only on contract instantiation, but also below, on th returned contractsInfo mapping, so I can't put it in the same line either.

raiden-ts/src/helpers.ts
andrevmatos: Sure, will do

raiden-ts/src/helpers.ts
andrevmatos: Because this contract is the most important one, and also one of the first. The `UserDeposit` and other services contracts are deployed only later in time (even because they must contain an address pointer to the TokenNetworkRegistry, therefore need to be known at instantiation time). The first deployed contract is actually the `SecretRegistry` (since `TokenNetworkRegistry` contains a pointer to it) and then the `TokenNetworkRegistry`, but only the later is guaranteed to have an event emitted early (the first registered `TokenNetwork`, through the `TokenNetworkCreated` event), and therefore it's the event we must scan for.

raiden-ts/src/helpers.ts
andrevmatos: I really don't understand how `sort` + `get first` is more inefficient than `Math.min`. I disagree with you on this, although I agree this can be made more explicit with `Math.min` as I did on the other PR. I'll rewrite this then.

raiden-ts/src/raiden.ts
andrevmatos: Yes. It actually do support function overloads, but only at interface level, so we could have the declared overloads with the shorter variables, but then we'd need an implementation with the full name like here. I'm not a big fan of it either, but since it's used only on this function and doesn't leak on the rest of the implementation, I'm fine with it also to maintain backwards compatibility.

raiden-ts/src/raiden.ts
andrevmatos: It's just a generic error. Sure, I'll include the details for this.

PR #1924, opened 2020-07-19
sdk: fix `fromEthersEvent` ranges fetching in case of temporary connectivity loss

### LINKED ISSUE #1923: `fromEthersEvent` utility can lose block ranges if more than `range` blocks are skipped, and errors if `getLogs` request throws
## Steps to Reproduce
This is an edge case which shouldn't occur on real life, and kind of theoric, although quite relevant for the event's reliability we require on the system.
#1832 introduced a new, more reliable algorithm for monitoring/filtering new contracts events which works by, in normal operation, ensuring the `range=5 (default)` last blocks are always scanned in each new block `getLogs` request for new events (instead of the default _last block only_ default heuristic from `ethers`). This works by both giving ~5 blocks of buffer for Infura nodes to pick up/sync a new block/event, and by giving each block/request ~5 opportunities to come in a request.
The above algorithm works fine if some requests may give inconsistent/slightly outdated results (like Infura), but less than 5 blocks are ever skipped on this polling. But, if for some reason (e.g. a temporary loss of connection), more than ~5 blocks are skipped, we would lose a sizeable portion of blocks in the range:
Example: we just received block=100 (which caused a `getLogs` in the range `[95, 100]`; then we lose connectivity, and only recover it in the block=120; the current algorithm will then scan range `[115, 120]`, and would have lost any event which happened between blocks `]100, 115[`.
Although losing connecion for more than 5 blocks could be considered undefined behavior, we want as much as possibly to not lose any range, and if possible, make sure every block in the range `[95, 120]` (in the above example) is given at least 
 `range` _opportunities_ of being scanned, regardless of connection loss.
Of course, if an event comes at any point, we still can be sure every event before that event's blockNumber was properly fetched, and can continue to optimize fetching from next block to avoid emit's duplication and optimize request.
Together with above, `getLogs` thrown exceptions (e.g. network request error) would be propagated to observable and errors code which doesn't expect it to ever error.

## Expected Result
- every block in range is scanned at least `range` times

## Actual Result
- some range inside connection loss time isn't fetched

andrevmatos:
Fixes #1923 
Follow-up of #1832 

**Short description**
This PR fixes the issue by using a `fromBlock` queue, instead of a single value, so we always keep track of "_[range]_ successful requests in the past", and can proceed with the correct ranges even in case of connectivity loss or exceptions thrown.
Also, the queue is only walked over (i.e. popping the front we just queried, pushing to the back the new block for future range request) if the request didn't throw, allowing us to safely ignore any network error/exception thrown while scanning ranges and retrying the previous `fromBlock` when a new `block` event is emitted, which indicates connectivity was restored.

**Definition of Done**

1. Disconnect the browser from internet while dApp is connected and monitoring some channel/tokenNetwork
2. Do some transaction which would show up on an event (e.g. a deposit on partner)
3. Wait more than 5 blocks after transaction, then connect back to internet
4. See event got picked up properly, network failures didn't error/shutdown the SDK

### DIFF
--- a/raiden-ts/src/utils/ethers.ts
+++ b/raiden-ts/src/utils/ethers.ts
@@ -4,7 +4,17 @@ import { JsonRpcProvider, Listener, EventType, Filter, Log } from 'ethers/provid
 import { Network } from 'ethers/utils';
 import { getNetwork as parseNetwork } from 'ethers/utils/networks';
 import { Observable, fromEventPattern, merge, from, of, EMPTY, combineLatest, defer } from 'rxjs';
-import { filter, first, map, mergeMap, share, toArray, debounceTime, tap } from 'rxjs/operators';
+import {
+  filter,
+  first,
+  map,
+  mergeMap,
+  share,
+  toArray,
+  debounceTime,
+  catchError,
+  exhaustMap,
+} from 'rxjs/operators';
 import sortBy from 'lodash/sortBy';
 
 import { isntNil } from './types';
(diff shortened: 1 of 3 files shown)


raiden-ts/src/utils/ethers.ts
```diff
@@ -43,29 +53,44 @@ export function fromEthersEvent<T>(
...
 
-  let first = 0;
-  let latestEventBlock = 0;
+  const blockQueue: number[] = []; // sorted 'fromBlock' queue, at most of [range] size
   return defer(() => {
-    // 'first' starts with subscription-time's private value set with provider.resetEventsBlock
-    first = (target as any)._lastBlockNumber || target.blockNumber || Number.POSITIVE_INFINITY;
+    // 'resetEventsBlock' is private, set at [[Raiden]] constructor, so we need 'any'
+    const resetBlock: number = (target as any)._lastBlockNumber;
+    const firstBlock = resetBlock && resetBlock > 0 ? resetBlock : target.blockNumber ?? 1;
```
weilbith, line 60: Which case would differ `resetBlock && resetBlock > 0` from `resetBlock > 0`? :thinking:

raiden-ts/src/utils/ethers.ts
```diff
@@ -43,29 +53,44 @@ export function fromEthersEvent<T>(
...
+    blockQueue.push(firstBlock); // starts 'blockQueue' with subscription-time's resetEventsBlock
```
weilbith, line 61: What is the purpose of first initiating an empty array and then pushing its first entry. Why not making it simple and say `const blockQueue = [firstBlock]`? :thinking:

raiden-ts/src/utils/ethers.ts
andrevmatos: The case in whic `resetBlock` is `undefined`. JS allows this comparison (which is always `false`), but TS complains if it discovers we're trying to compare `undefined` with `number`.

raiden-ts/src/utils/ethers.ts
andrevmatos: the `blockQueue.push` is called inside the `defer`, i.e. at **subscription** time, while the queue is defined outside of this context so it's available also for the following functions.

PR #1937, opened 2020-07-22
sdk: emit logs upon confirmation from fromEthersEvent

### LINKED ISSUE #1936: sdk: contracts events payload values can change in case of reorgs
Our events confirmation logic assumes (mistakenly) that events payload data would remain unchanged in case of reorgs. It expected that reorgs could at most remove the transaction, not mine it with completely different values. Unfortunately, that's the case. Specially with the `channelId` value, this can mess with the events monitoring logic.
This is quite critical, and breaks important assumptions which we made for a few places in the architecture which deal with contracts events and results, but it's mostly only triggerable in stress conditions such as SP's BF6 scenario.

## Steps to Reproduce
1. Run a scenario on SP where several nodes open channels on the same network concurrently (e.g. BF6)
2. If a reorg happen, the confirmed tx can have the same txHash of the first emit, but with different channelId, and deposit will fail and state will be inconsistent.

## Expected Result
- Confirmation logic and events monitoring can cope with reorgs

## Actual Result
- Undefined behavior, inconsistent state

```
- Browser & version
- Browser console errors
- Device / OS
```

andrevmatos:
Fixes #1936 

**Short description**
The problem with our previous confirmation logic was that it only took into account the `txHash` and at most updated a reorged `txBlock`: any pending action payload which depended on values which could change on reorgs (main example is `channel_identifier`, which currently depends strictly on tx order, since it's a nonce number) would get out-of-date and both the unconfirmed and its confirmed reemission would be inconsistent with the blockchain. This was actually witnessed upon stress testing on BF6 scenario on SP, which opens several channels on different clients simultaneously, and a reorg changed the order of the `openChannel` transactions on block and made a totally inconsistent state.

The fix keeps the previous `confirmationEpic`, but which should be used only for ConfirmableActions which have values/payloads which don't depend on order or reorgs of blocks/events on-chain. But now, the main way to confirm a log/event/action is through the same `fromEthersEvent` + `map(logToContractEvent)` utilities, which knows how to re-decode reorged events
- `fromEthersEvent` will watch double the range as before (`2 * confirmationBlocks`)
  - the higher half is for unconfirmed logs/events, as before, and the lower half is dedicated for confirmed logs;
  - unconfirmed logs can be emitted multiple times, in order to update their representation on `pendingTxs` state
  - confirmed actions are emitted as confirmed only once, and may replace the values with the confirmed counterparts; the `afterLogBlock` cleanup only takes place on this 'confirmed' part, to allow unconfirmed actions to be re-fetched
  - it can be used to query from a specific block in the past, taking the role previously with `getEventsStream`, in a sorted and standardized way
- `pendingTxs` reducer now deduplicate ConfirmableActions per action type and `txHash`, ensuring re-emitted unconfirmed actions are updated on state, which may make any depending unconfirmed action also be updated (e.g. deposits over unconfirmed channel)
- `makeDeposit$` function now needs to wait upon `openChannel` tx is confirmed, since it requires the `channelId` which only "stabilizes" after confirmation; `approve` side can still be done in parallel

**Definition of Done**

1. Check bug doesn't happen
2. Possible only once we can run BF6 scenario

### DIFF
--- a/raiden-ts/src/channels/reducer.ts
+++ b/raiden-ts/src/channels/reducer.ts
@@ -34,19 +34,23 @@ const tokens = createReducer(initialState.tokens).handle(
     state[token] === tokenNetwork ? state : { ...state, [token]: tokenNetwork },
 );
 
+function removeAction(pendingTxs: readonly ConfirmableAction[], action: ConfirmableAction) {
+  return pendingTxs.filter(
+    (a) => a.type !== action.type || action.payload.txHash !== a.payload.txHash,
+  );
+}
+
 const pendingTxs: Reducer<RaidenState['pendingTxs'], RaidenAction> = (
   state = initialState.pendingTxs,
   action: RaidenAction,
 ): RaidenState['pendingTxs'] => {
   // filter out non-ConfirmableActions's
   if (!ConfirmableAction.is(action)) return state;
-  // if confirmed==undefined, add action to state
-  else if (action.payload.confirmed === undefined) return [...state, action];
+  // if confirmed==undefined, deduplicate and add action to state
+  else if (action.payload.confirmed === undefined) return [...removeAction(state, action), action];
   // else (either confirmed or removed), remove from state
   else {
-    const newState = state.filter(
-      (a) => a.type !== action.type || action.payload.txHash !== a.payload.txHash,
-    );
+    const newState = removeAction(state, action);
     if (newState.length !== state.length) return newState;
     return state;
   }
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -30,6 +30,8 @@ import {
   first,
   delayWhen,
   finalize,
+  concatMap,
+  takeUntil,
 } from 'rxjs/operators';
 import findKey from 'lodash/findKey';
 import sortBy from 'lodash/sortBy';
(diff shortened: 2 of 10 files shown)


raiden-ts/src/channels/reducer.ts
```diff
@@ -34,19 +34,24 @@ const tokens = createReducer(initialState.tokens).handle(
...
   state = initialState.pendingTxs,
   action: RaidenAction,
 ): RaidenState['pendingTxs'] => {
   // filter out non-ConfirmableActions's
   if (!ConfirmableAction.is(action)) return state;
-  // if confirmed==undefined, add action to state
-  else if (action.payload.confirmed === undefined) return [...state, action];
+  // if confirmed==undefined, deduplicate and add action to state
+  else if (action.payload.confirmed === undefined)
+    return [...withoutAction(state, action), action];
```
weilbith, line 51: Nitpick: this function is **doing** something. So I would use a verb for it like `removeAction()`. The first time I read this (during the meeting) I didn't understood what this should mean. So I actually had to read the function itself to understand the idea. Now it makes sense. Anyways I think this could be improved. There is enough logic you need to understand. Little supporting hints like this are appreciated.

christianbrb: @weilbith This reminds of one of the most important books about software programming I have ever read (Code Complete by Steve McConnell - Chapter: Good Routine Names: https://flylib.com/books/en/2.823.1.63/1/).

christianbrb:
@andrevmatos Great PR description and process :) 👍

raiden-ts/src/channels/reducer.ts
andrevmatos: Done

PR #1959, opened 2020-07-25
sdk: don't request PFS if there's no suitable channel

### LINKED ISSUE #1948: Run BF1 scenario
## Description
Run the [BF1 scenario](https://github.com/raiden-network/raiden/blob/develop/raiden/tests/scenarios/bf1_basic_functionality.yaml) and document issues. Fix SP and/or the Light Client, if necessary.

Follow-up from #1691 

## Acceptance criteria
- Scenario is running on the developer machine
- Issues are identified
- Issues are fixed

## Tasks
- [ ]

### LINKED ISSUE #1958: sdk: pathFindingServiceEpic should fail early, before paying a PFS request, if no channel could route payment
Part of #1948 . 
BF1 currently fails because it expected that node 03 would only have done 10 PFS requests, but it did 11, because on the first payment try (which fails, as it should), a PFS request could be avoided and was done anyway. This is an optimization we didn't do before, and instead relyed on PFS to tell us no route existed, even though we could know that already.

## Steps to Reproduce
1. Run BF1
2. See it fail [on this line](https://github.com/raiden-network/raiden/blob/9bd24b1ca72555e7594540a16438534a00f19620/raiden/tests/scenarios/bf1_basic_functionality.yaml#L120)

## Expected Result
- That test pass

## Actual Result
- It fails because it expected 10 PFS requests, but 11 were made

```
- Browser & version
- Browser console errors
- Device / OS
```

andrevmatos:
Fixes #1958 
Fixes #1948 

**Short description**
Adds a simple `assert` to fail early if there's no channel which could handle the transfer.

**BF1** passes after this fix:

Screenshot_20200724_223951

**Definition of Done**

1. BF1 passes

### DIFF
--- a/raiden-ts/src/services/epics.ts
+++ b/raiden-ts/src/services/epics.ts
@@ -897,17 +897,29 @@ function validateRoute$(
 }
 
 function validateRouteTargetAndEventuallyThrow(action: pathFind.request, latest: Latest): void {
-  const { tokenNetwork, target } = action.meta;
+  const { tokenNetwork, target, value } = action.meta;
   const { state, presences } = latest;
 
-  if (!Object.values(state.tokens).includes(tokenNetwork))
-    throw new RaidenError(ErrorCodes.PFS_UNKNOWN_TOKEN_NETWORK, { tokenNetwork });
+  assert(Object.values(state.tokens).includes(tokenNetwork), [
+    ErrorCodes.PFS_UNKNOWN_TOKEN_NETWORK,
+    { tokenNetwork },
+  ]);
 
-  if (!presences[target].payload.available)
-    throw new RaidenError(ErrorCodes.PFS_TARGET_OFFLINE, { target });
+  assert(presences[target].payload.available, [ErrorCodes.PFS_TARGET_OFFLINE, { target }]);
 
-  if (presences[target].payload.caps?.[Capabilities.NO_RECEIVE])
-    throw new RaidenError(ErrorCodes.PFS_TARGET_NO_RECEIVE, { target });
+  assert(!presences[target].payload.caps?.[Capabilities.NO_RECEIVE], [
+    ErrorCodes.PFS_TARGET_NO_RECEIVE,
+    { target },
+  ]);
+
+  assert(
+    Object.values(state.channels).some(
+      (channel) =>
+        channelCanRoute(state, presences, tokenNetwork, channel.partner.address, target, value) ===
+        true,
+    ),
+    ErrorCodes.PFS_NO_ROUTES_BETWEEN_NODES,
+  );
 }
 
 function pfsIsDisabled(action: pathFind.request, latest: Latest): boolean {
--- a/raiden-ts/tests/e2e/raiden.spec.ts
+++ b/raiden-ts/tests/e2e/raiden.spec.ts
@@ -62,6 +62,7 @@ describe('Raiden', () => {
     revealTimeout: 5,
     confirmationBlocks: 2,
   };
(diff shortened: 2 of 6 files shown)


raiden-ts/src/services/epics.ts
```diff
@@ -897,17 +897,29 @@ function validateRoute$(
...
+    ErrorCodes.PFS_TARGET_NO_RECEIVE,
+    { target },
+  ]);
 
-  if (presences[target].payload.caps?.[Capabilities.NO_RECEIVE])
-    throw new RaidenError(ErrorCodes.PFS_TARGET_NO_RECEIVE, { target });
+  assert(
+    Object.values(state.channels).some(
+      (channel) =>
+        channelCanRoute(state, presences, tokenNetwork, channel.partner.address, target, value) ===
```
weilbith, line 918: I think we should rethink the `target` parameter for this function in general. I forgot the details of this function and while reading this I was slightly confused. So actually the target is only used to check if the partner is the target or not and if not check if he can mediate. Rather I would use a boolean flag that defines if the partner needs to mediate.
I would eventually open a tiny issue for that if you agree with me that this could be improved.

raiden-ts/tests/e2e/raiden.spec.ts
```diff
@@ -837,11 +835,21 @@ describe('Raiden', () => {
         await new Promise((resolve) => setTimeout(resolve, 100));
       });
 
-      afterEach(() => raiden1.stop());
-
       test('fail: direct channel without enough capacity, pfs disabled', async () => {
-        expect.assertions(2);
+        expect.assertions(3);
```
weilbith, line 839: What is the background of increasing the node count here and there?

raiden-ts/tests/unit/epics/path.spec.ts
```diff
@@ -958,11 +935,10 @@ describe('PFS: pathFindServiceEpic', () => {
     setTimeout(() => action$.complete(), 10);
 
     await expect(promise).resolves.toMatchObject(
-      pathFind.failure(expect.objectContaining({ message: ErrorCodes.PFS_NO_ROUTES_FOUND }), {
-        tokenNetwork,
-        target,
-        value,
-      }),
+      pathFind.failure(
+        { message: ErrorCodes.PFS_NO_ROUTES_BETWEEN_NODES },
```
weilbith, line 939: Is this actually a PFS error? :D
I know this related to the commit before which is throwing this error. But if the nitpick-policy recognized that this is not error of the path finding **service**. :stuck_out_tongue:

raiden-ts/tests/e2e/raiden.spec.ts
andrevmatos: If you look closely, the new opened channel has an `expect` on it

raiden-ts/tests/unit/epics/path.spec.ts
andrevmatos: Not all PFS errors come from the remote service: PFS errors are errors on the pathFinding epics and actions functionality. So this is it.

raiden-ts/src/services/epics.ts
weilbith: Done with #1968

raiden-ts/tests/e2e/raiden.spec.ts
weilbith: Ah f*** I can't read and misunderstood that. Super dump. I thought you increased the number of nodes.

PR #1971, opened 2020-07-27
sdk: ensure failing channelOpen with deposit doesn't deposit

### LINKED ISSUE #1969: channelOpen.request with deposit on an already opened channel deposits
Triggered by [this line of BF2](https://github.com/raiden-network/raiden/blob/9bd24b1ca72555e7594540a16438534a00f19620/raiden/tests/scenarios/bf2_long_running.yaml#L89). This was caused by #1957 , didn't get released yet, and shouldn't be easily triggerable outside of scenarios.
Issue detected as part of #1949 

## Steps to Reproduce
1. Somehow force the SDK to try to open a channel in a tokenNetwork+partner already with an open channel, with a deposit set

## Expected Result
- The request fails and no deposit is made

## Actual Result
- Deposit is performed

```
- Browser & version
- Browser console errors
- Device / OS
```

andrevmatos:

Fixes #1969 

**Short description**
Fix issue.

**Definition of Done**

1. Test issue doesn't happen

### DIFF
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -36,6 +36,7 @@ import {
 } from 'rxjs/operators';
 import sortBy from 'lodash/sortBy';
 import isEmpty from 'lodash/isEmpty';
+import findKey from 'lodash/findKey';
 
 import { BigNumber, concat, defaultAbiCoder } from 'ethers/utils';
 import { Event } from 'ethers/contract';
@@ -645,15 +646,21 @@ export const channelOpenEpic = (
   action$: Observable<RaidenAction>,
   state$: Observable<RaidenState>,
   { log, signer, address, main, provider, getTokenNetworkContract, config$ }: RaidenEpicDeps,
-): Observable<channelOpen.failure> =>
+): Observable<channelOpen.failure | channelDeposit.request> =>
   action$.pipe(
     filter(isActionOf(channelOpen.request)),
     withLatestFrom(state$, config$),
     mergeMap(([action, state, { settleTimeout, subkey: configSubkey }]) => {
       const { tokenNetwork, partner } = action.meta;
       const channelState = state.channels[channelKey(action.meta)]?.state;
+      const isPending = state.pendingTxs.some(
+        (a) =>
+          channelOpen.success.is(a) &&
+          a.meta.tokenNetwork === tokenNetwork &&
+          a.meta.partner == partner,
+      );
       // fails if channel already exist
-      if (channelState)
+      if (channelState || isPending)
         return of(
           channelOpen.failure(
             new RaidenError(ErrorCodes.CNL_INVALID_STATE, { state: channelState }),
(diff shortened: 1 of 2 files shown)


raiden-ts/src/channels/epics.ts
```diff
@@ -669,29 +676,45 @@ export const channelOpenEpic = (
...
+          takeUntil(
+            action$.pipe(
+              filter(channelOpen.success.is),
+              filter(
+                (action_) =>
+                  action_.meta.tokenNetwork === tokenNetwork && action_.meta.partner === partner,
+              ),
             ),
           ),
+          // ignore success so it's picked by channelEventsEpic
```
weilbith, line 714: Why do we not simply defer the deposit request here? Why do the complicated waiting stuff? Is there an  advantage? It feels like tightening the two epics too much together.

raiden-ts/src/channels/epics.ts
```diff
@@ -798,71 +815,72 @@ export const channelDepositEpic = (
...
-                  failure.meta.tokenNetwork === action.meta.tokenNetwork &&
-                  failure.meta.partner === action.meta.partner,
-              ),
-              mapTo(notOpenAction),
-            ),
-          ).pipe(take(1)); // for each request, take only first re-emited request or failure
-        }),
-        // concatMap to deposit on channels serially per tokenNetwork, in the order they're
-        // detected as 'open' and confirmed
+        // groupBy + concatMap ensure actions handling is serialized in a given tokenNetwork
```
weilbith, line 822: Shouldn't we group by channel only? Could be that we open a channel with an initial deposit. While we wait for the on-chain transactions to open before we can deposit, we could not deposit to an already open different channel in parallel. At least how I understand it.

raiden-ts/src/channels/epics.ts
```diff
@@ -798,71 +815,72 @@ export const channelDepositEpic = (
...
           combineLatest([latest$, config$]).pipe(
             first(),
             mergeMap(([{ state }, { subkey: configSubkey }]) => {
-              // if channelDeposit.failure, it means channel was not 'open', passthrough failure
-              if (channelDeposit.failure.is(action)) return of(action);
-
               const { tokenNetwork, partner } = action.meta;
-              // channel is guaranteed to be present and 'open', by mergeMap above
+
+              const token = findKey(state.tokens, (tn) => tn === tokenNetwork)! as Address;
```
weilbith, line 829: What is our guarantee that the token is already monitored? :thinking: Do we rely on the caller to check that first?

raiden-ts/src/channels/epics.ts
andrevmatos: Sorry, I don't get what you mean: this is exactly the same code as before (for open), the only addition is `concat`ing a `channelDeposit.request` before actually doing the channel open, so the `approve` can be sent/mined in parallel

raiden-ts/src/channels/epics.ts
andrevmatos: No, that wouldn't work, because both requests would fight for the `allowance` resource: `approve` needs to be called before `setTotalDeposit`, and `setTotalDeposit` also decreases allowance, and as they go on-chain, races could make us waste gas on unneeded transactions and require us to `retryTx` something which should work, so it needs to be serialized per token. Fortunatelly, this only happens for parallel/concurrent deposits on the same tokenNetwork, which usually doesn't happen outside of scenarios, and for them, it's ok to serialize and make it wait

raiden-ts/src/channels/epics.ts
andrevmatos: `Raiden.openChannel` & `Raiden.depositChannel` do that, it's guaranteed, but it's also checked and will fail below if not, so it's safe

raiden-ts/src/channels/epics.ts
weilbith: But isn't the approve done by the `makeDeposit` function which is called **after** the channel was opened? :thinking:

raiden-ts/src/channels/epics.ts
weilbith: Hmm okay.

raiden-ts/src/channels/epics.ts
andrevmatos: No! This was previously like that, but this fixes that by ensuring we only wait before `setTotalDeposit` (notice we pass the `channelId$`, which is `mergeMapTo` after `approve`), and that's precisely the optimization here

weilbith: Okay something I don't understand. We have [this line](https://github.com/raiden-network/light-client/blob/aa7edf8560de85101707899e67d38fcb9990c553/raiden-ts/src/channels/epics.ts#L867). The `channel$` observable will emit only when the channel has become available (assuming the `waitOpen` was set). Only then the pipe can trigger the `makeDeposit` call. Not correct?
I'll give you the approval since everything else is fine, but I'm only too stupid to get this details here.

andrevmatos: No, `publishReplay` (fortunately, and that's why it was chosen here) is able to create the "inner" observable as soon as it's subscribed, and merge it to the output, which means `makeDeposit$` will get called right away, but the `channelId$` will only get a value once something goes through the filters, and if it happens before the inner subscription or is an error, the implicit `ReplaySubject(1)` will ensure this is also _cached_ and forwarded to the output.

PR #1978, opened 2020-07-27
sdk,cli: expose and use transfer's lockTimeout

### LINKED ISSUE #1710: Raiden API compatibility discussion - transfer lock timeout
## Description

The Raiden API specifies the optional request parameter `lock_timeout` when initiating a transfer (careful, because API calls transfer payments). This is not available by the SDK and appears to follow a different strategy. This seem to be a similar compatibility discussion as for the reveal timeout (#1656).

## Acceptance criteria
- decision how to handle this parameter to be comptible
- cli endpoints got adjusted to decision
- sdk may provide new feature based on decision

## Tasks
- [ ]

andrevmatos:
Fixes #1710 

**Short description**
Title

**Definition of Done**

1. Test CLI honors `lock_timeout` param

### DIFF
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -810,6 +810,7 @@ export class Raiden {
    *     Is ignored if paths were already provided. If neither are set and config.pfs is not
    *     disabled (null), use it if set or if undefined (auto mode), fetches the best
    *     PFS from ServiceRegistry and automatically fetch routes from it.</li>
+   * @param options.lockTimeout - Specify a lock timeout for transfer; default is 2 * revealTimeout
    * @returns A promise to transfer's unique key (id) when it's accepted
    */
   public async transfer(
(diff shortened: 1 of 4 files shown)


raiden-ts/src/raiden.ts
```diff
@@ -840,6 +842,9 @@ export class Raiden {
     const pfs = !options.pfs
       ? undefined
       : decode(PFS, options.pfs, ErrorCodes.DTA_INVALID_PFS, this.log.info);
+    const expiration = !options.lockTimeout
+      ? undefined
+      : this.state.blockNumber + options.lockTimeout;
```
weilbith, line 847: You state in the header what the default of this value is. But you don't set it here, rather somewhere else. It took me a couple of minutes to search through the code base where this default is set. It is within the `makeAndSignTransfer$()` observable if I'm correct. Is there something that prevents us from defining this here and do not split it? :thinking:

andrevmatos: The reasoning for this is that the automatic value (when undefined) may be a little different at the time the `transfer.request` is processed (due to `concatMap`), and since it's automatic, I'd rather have it calculated at handling time, instead of at dispatch time like here (which, of course, isn't a concern if the `lockTimeout` is explicitly passed.
I'll add a comment to inform where the default is defined.

PR #1999, opened 2020-07-29
sdk: Fix concurrent login and confirmation of reverted by reorg events

### LINKED ISSUE #1963: Run BF7 scenario
## Description
Run the [BF7 scenario](https://github.com/raiden-network/raiden/blob/develop/raiden/tests/scenarios/bf7_long_path.yaml) and document issues. Fix SP and/or the Light Client, if necessary.

Follow-up from #1691 

## Acceptance criteria
- 

## Tasks
- [ ]

### LINKED ISSUE #1997: Starting many nodes at the same time can make transport login fail
On BF7 (#1963), the SP starts 15 instances in parallel. The LightClient initializes too fast, and all clients try to login in the server at the same time, triggering a `429 Limit Exceeded` error in some of the clients, which then proceed to try to register and fail (since registration is disabled due to the eth auth plugin).
This leads to the node shutting down with the following traceback:
```
2020-07-27T19:21:51.367Z [info] %c action     color: #03A9F4; font-weight: bold {
  type: 'raiden/shutdown',
  payload: {
    reason: M_UNKNOWN: Registration has been disabled
        at parseErrorResponse (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/matrix-js-sdk@7.1.0/node_modules/matrix-js-sdk/lib/http-api.js:849:13)
        at /mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/matrix-js-sdk@7.1.0/node_modules/matrix-js-sdk/lib/http-api.js:804:17
        at Request._callback (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/matrix-js-sdk@7.1.0/node_modules/matrix-js-sdk/lib/http-api.js:738:9)
        at Request.self.callback (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/request@2.88.2/node_modules/request/request.js:185:22)
        at Request.emit (events.js:315:20)
        at Request.<anonymous> (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/request@2.88.2/node_modules/request/request.js:1154:10)
        at Request.emit (events.js:315:20)
        at IncomingMessage.<anonymous> (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/request@2.88.2/node_modules/request/request.js:1076:12)
        at Object.onceWrapper (events.js:421:28)
        at IncomingMessage.emit (events.js:327:22) {
      errcode: 'M_UNKNOWN',
      data: [Object],
      httpStatus: 403
    }
  }
}
```

Fix is to retry 429 errors on login

## Steps to Reproduce
1. Run BF7 scenario

## Expected Result
- All nodes can get up and login

## Actual Result
- Some nodes are rate-limited and don't start

```
- Browser & version
- Browser console errors
- Device / OS
```

### LINKED ISSUE #1998: Reorgs which make previous tx of interest revert can get confirmed by confirmationsEpic
This is a very edgy case, where reorgs can make transactions previously accepted get reverted, and while the `fromEthersEvent + map(logToContractEvent)` can detect the new confirmed transaction, the previously detected event which now got reverted may remain in `pendingTxs` and get wrongly confirmed by `confirmationsEpic`.
The fix is just to check the `status` of the transaction on `confirmationsEpic`.

Additionally, now that we can be sure confirmed `ChannelOpened` events will go through before at least one additional pass of `getLogs` for other events, we should also avoid emitting unconfirmed events for channels not yet confirmed, to avoid them hanging on `pendingTxs` and getting confirmed with wrong values. It's safe to emit them only once we're sure about the (confirmed) `channelId`, and still know that any unconfirmed or confirmed event after it will still be checked again.

Detected as part of #1963, where  `channelOpen.request`s are sent in parallel in both directions, and it expects the concurrent channel detection to take place and deposit to be done on such channel.

## Steps to Reproduce
1. Run BF7, see it fail sometimes in case of reorgs

## Expected Result
- Reorgs which make previously detected unconfirmed transactions revert are properly detected and cleaned

## Actual Result
- See the reverted event be confirmed by `confirmationsEpic`

andrevmatos:
Fixes #1963
Fixes #1997
Fixes #1998

**Short description**
Fixes some edge cases on high concurrency. This kind of errors aren't expected on normal operations, and are triggered by BF7 level of concurrency.

The scenario passes with these fixes:

Screenshot_20200729_161358

**Definition of Done**

1. Run BF7


raiden-ts/src/transport/epics/helpers.ts
```diff
@@ -156,10 +156,10 @@ export function waitMemberAndSend$(
           // up to RETRY_COUNT times; if it continues to error, throws down
           err$.pipe(
             withLatestFrom(config$),
-            mergeMap(([err, { httpTimeout }], i) => {
+            mergeMap(([err, { pollingInterval }], i) => {
```
weilbith, line 159: Isn't the polling interval meant for the eth provider? :thinking:

raiden-ts/src/transport/epics/init.ts
```diff
@@ -222,76 +222,100 @@ function fetchSortedMatrixServers$(matrixServerLookup: string, httpTimeout: numb
...
+              () =>
+                matrix.login('m.login.password', {
+                  user: userName,
+                  password,
+                  device_id: DEVICE_ID,
+                }),
+              pollingInterval,
+              (err) => err?.httpStatus !== 429, // retry only if rate-limit error
+            ).pipe(
+              catchError((err) =>
```
weilbith, line 268: I know this is not part of this PR, but couldn't we check for a specific error that we are required to register first? Similar like we do above for the rate limit error.

andrevmatos: This doesn't happen on real life. The `register` is just a fallback for when using vanilla matrix servers, while all Raiden matrix servers have the ETH auth plugin, which only requires login and disables register. So, we first try to login with the deterministic password, and try to register if it fails, but if it fails once, then we should be good to throw and move to the next server, as the server is probably not working properly as we expect, and we want to fail early and optimize for the common case.

weilbith: Hmm okay.

PR #2009, opened 2020-07-30
Rename VUE_APP_TRANSPORT and support VUE_APP_MATRIX_LIST_URL envvar

### LINKED ISSUE #2002: CORS error for light client on all transport servers on mainnet
## Description
The light client encounters CORS exceptions while calling the mainnet matrix server API's at startup
This is not the problem of the light client but its just that the matrix servers or for that matter any other servers if they are called from a different domain and return the reply without the `Access-Control-Allow-Origin` headers set then the browser blocks the response. 

light_client_cors_error

## Acceptance criteria
- The only thing we can do in this case is to request RSB providers to allow certain domains like 
1. https://lightclient.raiden.network/
2. http://localhost:8080/
may be some more commonly used domains used by light-client users as allowed on the CORS headers 
## Tasks
-  [ ]  RSB providers to Send CORS headers (Access-Control-Allow-Origin)

[raiden_0x14efA2b271969A46A094A23e0d49E32C1B617f89_2020-07-30T07_39_20.313Z.log](https://github.com/raiden-network/light-client/files/5000313/raiden_0x14efA2b271969A46A094A23e0d49E32C1B617f89_2020-07-30T07_39_20.313Z.log)

@christianbrb please check whether the content of this issue is correct.

andrevmatos:
Fixes #2002 

**Short description**
- Rename `VUE_APP_TRANSPORT` to `VUE_APP_MATRIX_SERVER` to avoid conflict with previous var
- Add `VUE_APP_MATRIX_LIST_URL`:
  - passed to SDK's `matrixServerLookup` config
  - Has less priority than `VUE_APP_MATRIX_SERVER`, so make sure that's unset if one wants to iterate over the list in case of the first server isn't responding properly
  - Should be set to a URL ofa file containing a yaml-like list of servers, e.g.: `https://raw.githubusercontent.com/raiden-network/raiden-service-bundle/master/known_servers.main.yaml`

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-dapp/.env.development
+++ b/raiden-dapp/.env.development
@@ -1,4 +1,4 @@
 VUE_APP_PFS=https://pfs.demo001.env.raiden.network
-VUE_APP_TRANSPORT=https://transport.demo001.env.raiden.network
+VUE_APP_MATRIX_SERVER=https://transport.demo001.env.raiden.network
 VUE_APP_HUB=hub.raiden.eth
 VUE_APP_ALLOW_MAINNET=true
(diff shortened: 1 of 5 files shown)


andrevmatos:
@agatsoh this should allow you to set the mainnet's known_server's list URL, and the client will sort them by response times for you, and also try next if one server is failing or not CORS-enabled.
Make sure to **unset** the previous transport var (not only set it to empty)

PR #2021, opened 2020-08-01
Exponential back-off strategy for delays between protocol messages retries

### LINKED ISSUE #2019: Implement exponential back-off strategy for protocol messages retry
## Description
It's more efficient than a constant value, as we use currently, since it can retry over shorter delays in the beginning, when there's a higher probability of the partner catching up and responding to our message.
This is the same strategy as used in raiden-py

## Acceptance criteria
- Message are retried using an exponential back-off spacing of delays

## Tasks
- [ ]

andrevmatos:
Fixes #2019 

**Short description**
retry utilities also accept an `Iterator<number>` for the delay, and will wait each yielded value between retries; a `exponentialBackoff` generator utility is provided, and used for the protocol messages retries, starting with `config.pollingInterval = 5s`, and increasing this delay by 40% on each iteration up to the maximum value, which we set by default as `2 * httpTimeout = 60s`
This is an optimization, since first retries can get a message accepted (if partner synced to accept the message) sooner than a constant value, like before

**Definition of Done**

1. MS4 can trigger reject a message sometimes due to partner not yet having confirmed the deposit
2. Check the transfer message is retried in 5s instead of 30s, and accepted

### DIFF
--- a/raiden-ts/src/transfers/epics/retry.ts
+++ b/raiden-ts/src/transfers/epics/retry.ts
@@ -17,7 +17,7 @@ import {
   transferSecretReveal,
 } from '../actions';
 import { Direction } from '../state';
-import { retrySendUntil$ } from './utils';
+import { retrySendUntil$, exponentialBackoff } from './utils';
 
 /**
  * Handles a transferSigned action and retry messageSend.request until transfer is gone (completed
@@ -40,7 +40,7 @@ const signedRetryMessage$ = (
   if (action.meta.direction !== Direction.SENT) return EMPTY;
   return config$.pipe(
     first(),
-    switchMap(({ httpTimeout }) => {
+    switchMap(({ pollingInterval, httpTimeout }) => {
       const secrethash = action.meta.secrethash;
       const signed = action.payload.message;
       const send = messageSend.request(
@@ -60,7 +60,12 @@ const signedRetryMessage$ = (
         ),
       );
       // emit request once immediatelly, then wait until success, then retry every 30s
-      return retrySendUntil$(send, action$, notifier, httpTimeout);
+      return retrySendUntil$(
+        send,
+        action$,
+        notifier,
+        exponentialBackoff(pollingInterval, httpTimeout * 2),
+      );
     }),
   );
 };
@@ -86,7 +91,7 @@ const unlockedRetryMessage$ = (
   return state$.pipe(
     first(),
     withLatestFrom(config$),
-    switchMap(([state, { httpTimeout }]) => {
+    switchMap(([state, { pollingInterval, httpTimeout }]) => {
       const secrethash = action.meta.secrethash;
       const unlock = action.payload.message;
       const locked = state.sent[secrethash].transfer[1];
@@ -101,7 +106,12 @@ const unlockedRetryMessage$ = (
       );
       // emit request once immediatelly, then wait until respective success,
       // then repeats until confirmed
-      return retrySendUntil$(send, action$, notifier, httpTimeout);
+      return retrySendUntil$(
+        send,
+        action$,
+        notifier,
+        exponentialBackoff(pollingInterval, httpTimeout * 2),
+      );
     }),
   );
 };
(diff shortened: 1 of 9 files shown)


raiden-ts/src/transfers/epics/retry.ts
```diff
@@ -60,7 +60,12 @@ const signedRetryMessage$ = (
         ),
       );
       // emit request once immediatelly, then wait until success, then retry every 30s
-      return retrySendUntil$(send, action$, notifier, httpTimeout);
+      return retrySendUntil$(
+        send,
+        action$,
+        notifier,
+        exponentialBackoff(pollingInterval, httpTimeout * 2),
```
weilbith, line 67: I'm a bit confused about this mix of the polling interval and the timeout for HTTP. It feels a bit randomly selected. Could we please document what each of the values is meant for in detail? Or point me to the existing one. Because right now it feels more like just using some values we already have disregarding their purpose.

raiden-ts/src/utils/rx.ts
```diff
@@ -66,23 +68,32 @@ export function distinctRecordValues<R>(
 /**
  * Operator to repeat-subscribe an input observable until a notifier emits
  *
- * @param notifier - Notifier observable
- * @param delayMs - Delay between retries
- * @returns Monotype operator
+ * @param notifier - Notifier observable to stop repeating
+ * @param delayMs - Delay between retries; in milliseconds, or an Iterator of delays
```
weilbith, line 72: Both are in ms right?
```suggestion
 * @param delayMs - Delay between retries or an Iterator of delays; in milliseconds
```

raiden-ts/src/utils/rx.ts
```diff
@@ -66,23 +68,32 @@ export function distinctRecordValues<R>(
...
     input$.pipe(
-      repeatWhen((completed$) => completed$.pipe(delay(delayMs))),
+      repeatWhen((completed$) =>
+        completed$.pipe(
+          map(() => {
+            if (typeof delayMs === 'number') return delayMs;
+            const next = delayMs.next();
+            return !next.done ? next.value : -1;
+          }),
+          takeWhile((value) => value >= 0), // stop repeatWhen when done
```
weilbith, line 93: When will this become true for the plain number? :thinking:

raiden-ts/src/transfers/epics/retry.ts
andrevmatos: These values are just "constant" values which happen to be configurable, which enables us to set custom ones in tests to have these generic/general "waitings" & "timeouts" as sensible values which won't make our tests hang forever.
As I explained in the transport call, these are just config values, and we should keep in mind their defaults when using them:
- `pollingInterval=5s`, which is a nice initial retry delay: notice this is also the value used by `JsonRpcProvider` to poll for new blocks, so this means there's a good chance for, after this interval, the partner have retried, seen a new block/event, which makes them now accept a previously ignored message from us
- `httpTimeout=30s`, which is still a little bit spammy if the partner hasn't accepted our messages so far, probably they aren't going to, but we need to keep trying until something else happens (e.g. channel gets closed and disputed on-chain), so 60s is just a sensible value

raiden-ts/src/utils/rx.ts
weilbith: So in this case it would depend on the `notifier`?

raiden-ts/src/transfers/epics/retry.ts
weilbith: Hmm. I just want to increase the awareness how we use these values. In the end this are configurable values. If the user/developer can't know anymore (and **we too**) what is actually all affected by this, then we have a big problem. This example might sound awkward because we are not there yet with these values, but it shows the issue perfectly. So we have the option to configure the `settleTimeout`. The user of the SDK expects this value to be used for things that affect the settlement of channels. If we would then think, oh the default is actually a nice value we could use it also for something else and it is actually somehow related with some phantasy, then the user affects behavior with changes of this setting he does not intended. And this could lead to behavior also we don't understand anymore.

So as long as it makes sense to maintain such time values for a more general purpose I'm fine with it. Just make sure they are explained like that to the user. It must be possible to describe them. And when you use them you should check if you are still following this definition. Else this will be the start of a messy codebase.

raiden-ts/src/utils/rx.ts
andrevmatos: It already does most of the time, even if exponential backoff. This `takeWhile` is just a way to be more consistent with the introduced generator, but usually the one to tell to stop is `notifier`, e.g.: a `Processed` message came for that LockedTransfer, give up retrying because we're good already.

PR #2048, opened 2020-08-06
Re-Writing the monitorRequestEpic test

agatsoh:

Fixes # 

**Short description**

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/tests/unit/epics/monitor.spec.ts
+++ b/raiden-ts/tests/unit/epics/monitor.spec.ts
@@ -1,6 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import {
-  raidenEpicDeps,
   makeLog,
   makeAddress,
   makeHash,
@@ -12,7 +11,6 @@ import {
   makeRaiden,
 } from '../mocks';
 import {
-  epicFixtures,
   tokenNetwork,
   token,
   deposit,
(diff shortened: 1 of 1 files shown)


raiden-ts/tests/unit/epics/monitor.spec.ts
```diff
@@ -357,6 +369,268 @@ describe('monitorRequestEpic', () => {
   });
 });
 
+describe('monitorRequestEpicNewPattern', () => {
```
andrevmatos, line 372:
```suggestion
describe('monitorRequestEpic', () => {
```

raiden-ts/tests/unit/epics/monitor.spec.ts
```diff
@@ -357,6 +369,268 @@ describe('monitorRequestEpic', () => {
...
+    raiden.store.dispatch(raidenConfigUpdate({ monitoringReward: null }));
+
+    await ensureChannelIsDeposited([raiden, partner], deposit);
+    await ensureChannelIsDeposited([partner, raiden], deposit);
+    await ensureTransferUnlocked([partner, raiden], amount);
+    await waitBlock();
+
+    expect(raiden.output).not.toContainEqual(
+      messageGlobalSend(
+        { message: expect.objectContaining({ type: MessageType.MONITOR_REQUEST }) },
```
andrevmatos, line 519: Very nice! Broad `.not.toContainEqual`, and strict whole message checks when it does

raiden-ts/tests/unit/epics/monitor.spec.ts
```diff
@@ -357,6 +369,268 @@ describe('monitorRequestEpic', () => {
...
+      .spyOn(raiden.deps.signer, 'signMessage')
+      .mockImplementation(async (message) => {
+        if (signerSpy.mock.calls.length > 1) throw new Error('Signature rejected');
+        return originalSign.call(raiden.deps.signer, message);
+      });
+
+    await ensureTransferUnlocked([partner, raiden], amount);
+    await waitBlock();
+
+    expect(raiden.output).not.toContainEqual(
```
andrevmatos, line 558: Please, since you wanna also validate it's `not fatal`, check output doesn't contain any `raidenShutdown` after the rejections

raiden-ts/tests/unit/epics/monitor.spec.ts
```diff
@@ -357,6 +369,268 @@ describe('monitorRequestEpic', () => {
...
+      messageGlobalSend(
+        { message: expect.objectContaining({ type: MessageType.MONITOR_REQUEST }) },
+        expect.anything(),
+      ),
+    );
+    // expect(signerSpy).toHaveBeenCalledTimes(2);
```
andrevmatos, line 564: And also ensure that `signerSpy` got called enough to reach the throw part, and didn't shutdown, or else it could shutdown because of an unhandled error and the test wouldn't notice.

agatsoh:
I took care of your comments in a fixup commit please take a look

raiden-ts/tests/unit/epics/monitor.spec.ts
agatsoh: Took care in next fixup commit

raiden-ts/tests/unit/epics/monitor.spec.ts
agatsoh: taken care in next commit please have a look

raiden-ts/tests/unit/epics/monitor.spec.ts
agatsoh: Got confused so changed it from my side

raiden-ts/tests/unit/epics/monitor.spec.ts
```diff
@@ -110,124 +81,58 @@ describe('monitorRequestEpic', () => {
...
-        },
-        { tokenNetwork, partner },
-      ),
-      channelDeposit.success(
+    );
+    await ensureChannelIsDeposited([raiden, partner], deposit);
+    await ensureChannelIsDeposited([partner, raiden], deposit);
+    await ensureTransferUnlocked([partner, raiden], amount);
+
+    await waitBlock(2);
```
andrevmatos, line 89: `ẁaitBlock(2)` doesn't wait 2 blocks, but emits the `blockNumber=2`, to all clients, which is wrong since it's smaller than the default starting block of 100. This just works here because it's not being used by the specific block, and instead only for waiting a block and the time for its event to be properly handled.
```suggestion
    await waitBlock();
```

raiden-ts/tests/unit/epics/monitor.spec.ts
```diff
@@ -241,119 +146,186 @@ describe('monitorRequestEpic', () => {
...
 
-    expect(signerSpy).toHaveBeenCalledTimes(2);
-    signerSpy.mockRestore();
+    expect(raiden.output).not.toContainEqual(
+      messageGlobalSend(
+        { message: expect.objectContaining({ type: MessageType.MONITOR_REQUEST }) },
+        expect.anything(),
+      ),
+    );
+    expect(last(raiden.output)).not.toEqual(raidenShutdown({ reason: error }));
```
andrevmatos, line 259: You want to be broad. Don't check only `last` (although it'd be safe to expect the shutdown would be the last action), but be broader that it output does `.not.toContainEqual(raidenShutdown(expect.anything()))`

agatsoh:
Thanks for fast approval Andre

agatsoh:
Would you also merge PR or i can do it ? @andrevmatos

Issue #2038, opened 2020-08-04
Improve SDK code navigation by avoiding import all statements


weilbith:
## Description

Augusto just came to me asking where all the exported epic functions get used. If there is any automated magic. His IDE could not see any usages/references of these functions in the code base.
This is because we use in the root epic import statements like `import * as ChannelsEpics from './channels/epics';`. The big advantage of this is the single responsibility. We can add new epics without changing the root epic module. The disadvantage is the code navigation and the resulting readability.

So what does this import do. It collects all export statements of the module to import and put them into an object we name ourselves. Later we then disassemble them again to mix all epics together. In attempt to get the best of both worlds, I would recommend to rather do an `export default ChannelEpics = { confirmationEpic, ... }` in `channels/epics.ts` and then do `import ChannelEpics from './channels/epics';`. So the root epic module would remain unaffected of code changes when adding/removing/changing the epics in all the smaller modules. But we would get code navigation.

## Acceptance criteria
- 

## Tasks
- [ ]

weilbith:
@andrevmatos your thoughts

andrevmatos:
Please, not. This would still require changing 2 places to add a new epic: the implementation and the default exported mapping. It doesn't matter that much that these 2 places would be in the same file, and that's not even true for split epics modules.
I do get proper code navigation here, with `coc.nvim` and `coc-tsserver`. VSCode also gets proper code navigation with the `*` imports, no questions asked. We may instead have a guide for properly configuring editor's intellisense. Also, these imports are not exported, they're local to the `src/epics.ts` file, it doesn't make sense to go over all this hassle just to fix some local imports/variables.

PR #2050, opened 2020-08-06
Using epics by name.

hackaugusto:
I'm opening the PR even though it is broken because I want to see if you guys want to merge something like this or not.

It is only a syntatic change, the only benefit is that one can run `grep CloseEpics` or so and see were the epic is being used. Not terribly useful but IMO it makes a bit easier to navigate through the codebase. Anyways, not merging this is also fine.


weilbith:
See #2038

hackaugusto:
I did not know this was discussed in an issue.

PR #2073, opened 2020-08-11
Make it explicit the requirement for ES2019/Node12 build

### LINKED ISSUE #2049: Bump ECMAScript compatibility to ES6
## Description
ES6 is already [standard on browsers and nodejs](https://www.w3schools.com/js/js_es6.asp) for a few years for now, and its module system provides a huge benefit in facilitating tree shaking.
Initially, we went with ES5 due to the need  to run the light-client bundle on small runtimes which only supported this old version (#551), but since then, we managed to run the full `NodeJS` on these devices without problem, so we don't have this constraint anymore, and ES6 is already the new default.
Any project which strictly require ES5 should just drop babel and polyfill it on their own, there's absolutely no need for this to be provided from the library side, and it's already not used in neither dApp nor CLI, which always transpile from the more modern version.

## Acceptance criteria
- 

## Tasks
- [ ]

andrevmatos:
Fixes #2049 (for now, until we decide to revisit the ESM subject)

**Short description**
We already use a few features from up to ES2018, but we weren't explicit about them. Node12 (our minimum required version) supports the full set of ES2019. Therefore, we bump the build requirement to this version, in order to produce more optimized and smaller code. Babel must be used anyway if someone really need to run the SDK on environments which support ES5- only.
We need to keep the CommonJS build, as it's required on Node for now.

**Definition of Done**

1. Build the SDK
2. Build the CLI
3. Check CLI still runs on bare Node12 (without `--experimental-*` flags)

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## [Unreleased]
+### Changed
+- [#2049] Target ES2019 (NodeJS 12+) on SDK builds
 
 ## [0.11.0] - 2020-08-04
 ### Fixed
@@ -40,6 +42,7 @@
 [#2010]: https://github.com/raiden-network/light-client/issues/2010
 [#2019]: https://github.com/raiden-network/light-client/issues/2019
 [#2022]: https://github.com/raiden-network/light-client/pull/2022
+[#2049]: https://github.com/raiden-network/light-client/issues/2049
 
 ## [0.10.0] - 2020-07-13
 ### Fixed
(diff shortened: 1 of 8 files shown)


raiden-ts/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
 ## [Unreleased]
+### Changed
+- [#2049] Target ES2019 (NodeJS 12+) on SDK builds
```
weilbith, line 5: Do we have some signal/flag/hint/... that this is a breaking change? :thinking:

andrevmatos: I don't think we have. At least, never went through it. In any case, this shouldn't be "breaking" right now, in the sense the the SDK and CLI already implicitly required Node12, so it shouldn't break any current usecase.

Issue #2044, opened 2020-08-04
Move state storage to IndexedDB/LevelDB


andrevmatos:
## Description
Fixes #1933 #1831
The above issues seems to be caused by client losing state, as we've feared for long, or client crashing before persisting state after a critical state change took place (like new BalanceProof created). This is a tricky problem, due to the ephemeral nature of in-browser storage keeping.

Current `RaidenState` is stored by default like this:
- An optional `localStorage` instance is received by `Raiden.create`
- The static factory then registers a debounced callback to `state$` observable, which by default **serialize** and store the state after `1s` without a new state being generated, and at most every `5s` if new states keep going through.
  - This is done because the `serialize` part can be quite CPU intensive, and both the serialize and state saving on `LocalStorage` currently are synchronous operations which can take time and block other more important events processing.
  - Upon shutdown (for any reason), the state is also persisted, but of course this won't protect from browser suddenly crashing or other catastrophic event like this.
  - Unfortuntately, this is quite a big time window for state to be lost, and can lead to the mentioned bugs.
  - LocalStorage is also quite limited in size, often being capped at 5MB in browsers, which may not be enough for longer-lived clients.

The proposed solution is to move the state to the `IndexedDB` or even directly a `LevelDB`:
- Available on all modern browsers
- Have a way higher size cap, often going up to 50% of user's available disk space
- Quite easily extendable to use external/dedicated keystore/database systems, like Mongo, Redis or even Postgres, for even stricter threat models.
- `put` operations are asynchronous
- IndexedDB/LevelDB is able to receive the state _as is_, without requiring serializing them ourselves
  - the only reason to serialize the state currently would be `BigNumber`, but IndexedDB is able to encode them as `{"_hex": "0x1234"}` objects which, as long as we're able to decode (and [we already are](https://github.com/raiden-network/light-client/blob/1aaf46fbde748a8058c401ab168874d61857ada2/raiden-ts/src/utils/types.ts#L77)), should allow us to persist the state directly, without an expensive encoding step
  - Our immutable state policy also allows us to do this right away, without worrying about locking or concurrent changes to the storing state
- We could keep certain parts of the state which are prone to growing indefinitely, like `sent` & `received` transfers, on a separate object store, or key prefix, for higher efficiency on storing of the general state, although requiring async access (shouldn't be a problem for the async epics, may require some thinking for the reducers)
  - Fixes #1618
[...]

PR #2075, opened 2020-08-11
Improve types and convert timed tuples to extended objects

andrevmatos:
Part of #2044

**Short description**
This change is purely internal, and shouldn't affect the user at all.
It improves some refined tipes inheritance (in order to properly identify them later) and convert the `timed` tuples (`[<timestamp>, <object>]`) to an extended object (`{ ...<object>, ts: number }`), which effectively extends the object (i.e. the resulting object is assignable to a variable of type of the original object) and can be properly described by a `json-schema` (for RxDB), while tuples can't on the json-schema subset supported by rxdb.

**Definition of Done**

1. Tests cover the whole extend of these changes; migration and migration tests are also provided

### DIFF
--- a/raiden-ts/src/utils/types.ts
+++ b/raiden-ts/src/utils/types.ts
@@ -221,26 +221,42 @@ export const Address: AddressC = new t.RefinementType<HexStringC<20>, Address, s
 );
 
 /**
- * Helper function to create codecs to validate [timestamp, value] tuples
+ * Helper type to extend a given type T to contain a timestamp ts member
+ */
+export type Timed<T> = T & { readonly ts: number };
+export interface TimedC<T extends t.Mixed>
+  extends t.IntersectionC<[T, t.ReadonlyC<t.TypeC<{ ts: t.NumberC }>>]> {}
+/**
+ * Helper function to create codecs to validate derived types containing a timestamp ts
  *
- * @param codec - Codec to compose with a timestamp in a tuple
- * @returns Codec of a tuple of timestamp and codec type
+ * @param codec - Codec to compose with a ts timestamp property
+ * @returns Codec validating such subtype
  */
 export const Timed: <T extends t.Mixed>(
   codec: T,
-) => t.TupleC<[t.NumberC, T]> = memoize(<T extends t.Mixed>(codec: T) =>
-  t.tuple([t.number, codec]),
+) => TimedC<T> = memoize(<T extends t.Mixed>(codec: T) =>
+  t.intersection([codec, t.readonly(t.type({ ts: t.number }))]),
 );
-export type Timed<T> = [number, T];
 
 /**
- * Given a value of type T, returns a Timed<T> tuple with current time as first value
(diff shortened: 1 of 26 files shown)


raiden-ts/src/utils/types.ts
```diff
@@ -221,34 +221,50 @@ export const Address: AddressC = new t.RefinementType<HexStringC<20>, Address, s
 );
 
 /**
- * Helper function to create codecs to validate [timestamp, value] tuples
+ * Helper type to extend a given type T to contain a timestamp ts member
+ */
+export type Timed<T> = T & { readonly ts: number };
```
weilbith, line 226: Do you think `ts` is sufficient enough? :thinking: 
In future comment you always state `a ts timestamp property`. So does `ts` mean something else than **t**ime**s**tamp?

raiden-ts/src/utils/types.ts
```diff
@@ -221,34 +221,50 @@ export const Address: AddressC = new t.RefinementType<HexStringC<20>, Address, s
...
+  return { ...v, ts: Date.now() };
+}
+
+/**
+ * Remove ts timestamp field (from timed) from object passed as parameter (immutably)
+ *
+ * @param v - Timed object
+ * @returns return a copy of v without ts property
+ */
+export function untime<T extends { readonly ts: number }>(v: T): Omit<T, 'ts'> {
```
weilbith, line 257: With my bad knowledge of this: why is `T extends { readonly ts: number}` not `Timed<T>`?

raiden-ts/src/utils/types.ts
andrevmatos: `ts` is the abbreviation for timestamp here, and the name of the property. `timestamp` on this comment is used as a _[semantic] type_ for the property: i.e. a number which contains unix epoch value in milliseconds. But yes, they're the same thing, just wanted to be more descriptive on the type docstring.

raiden-ts/src/utils/types.ts
andrevmatos: `A extends B` on generics isn't a statement of extension, but a type **constraint**. It means TypeScript won't accept anything for A which doesn't somehow extends the type of B. In this case, I want to constrain it to a type which can be assigned to a variable of type `{ readonly ts: number }`. Instead of having the literal type there, I could use the equivalent `Timed<{}> = {} & { readonly ts: number } = { readonly ts: number };`, but TypeScript/ESLint doesn't like empty object types, so I thought it wouldn't matter much to have the type again since it's close enough.

raiden-ts/src/utils/types.ts
weilbith: Hmm. I would loved to have `timestamp` as the property name. :sweat:

andrevmatos: `ts` is very commonly used as abbreviation of `timestamp`, including on the linux kernel. don't suffer for it xD

weilbith: You definitely got me with the argument that the Linux kernel does the same. :grimacing:

PR #2076, opened 2020-08-12
Simplified transferSecretReducer

hackaugusto:
The main problem I had with this function was actually the comment:

https://github.com/raiden-network/light-client/blob/0abfd9ee9d4359b150048f7018dd13fe07175086/raiden-ts/src/transfers/reducer.ts#L41

Which led me to think the code was never overwriting the `registerBlock`. My interpretation of the comment was wrong, and the code is overwriting the `registerBlock` as long as the value changes.

This PR just moved a few bits around which IMO makes things a bit easier to read. Nothing crucial though, just a suggesiton.

### DIFF
--- a/raiden-ts/src/transfers/reducer.ts
+++ b/raiden-ts/src/transfers/reducer.ts
@@ -1,4 +1,5 @@
 import pick from 'lodash/fp/pick';
+import { assert } from '../utils/error';
 import { channelKey } from '../channels/utils';
 import { RaidenState, initialState } from '../state';
 import { RaidenAction } from '../actions';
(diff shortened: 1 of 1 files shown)


raiden-ts/src/transfers/reducer.ts
```diff
@@ -32,27 +32,25 @@ function transferSecretReducer(
...
-  )
-    state = {
-      ...state,
-      [action.meta.direction]: {
-        ...state[action.meta.direction],
-        [secrethash]: {
-          ...state[action.meta.direction][secrethash],
-          secret: timed({ value: action.payload.secret, registerBlock }),
-        },
+      : 0;
```
andrevmatos, line 42: This will reset the `registerBlock` if a `transferSecret` action (which doesn't contain a block) goes through, even if the block had been registered before. We don't expect many of these actions to come after a register, but it should be supported by the reducer and it should not lose this information.

raiden-ts/src/transfers/reducer.ts
```diff
@@ -32,27 +32,25 @@ function transferSecretReducer(
...
+
+  return {
+    ...state,
+    [action.meta.direction]: {
+      ...state[action.meta.direction],
+      [secrethash]: {
+        ...transferState,
+        secret: timed({ value: action.payload.secret, registerBlock }),
```
andrevmatos, line 50: This will reset the `ts` timestamp every time an action for this is handled. The `timed` version should only be constructed when the value changes, as it represents when it got to that state. Currently, this timestamp is only used for informative purposes, but we must keep it correct.

hackaugusto:
@andrevmatos comments should have been addressed

raiden-ts/src/transfers/reducer.ts
```diff
@@ -33,27 +34,37 @@ function transferSecretReducer(
...
+  const transferState = state[action.meta.direction][secrethash];
+  if (!transferState) return state;
+
+  // - Preserve the timestamp of the secret registration.
+  // - Preserve block number if a secret reveal message is received *after* the
+  //   secret is registered.
+  const previousRegisterBlock = transferState.secret?.registerBlock;
+  if (previousRegisterBlock) {
+    if (transferSecretRegister.success.is(action)) {
+      assert(action.payload.txBlock == previousRegisterBlock);
```
andrevmatos, line 47: Reducers never, ever throw. If some expectation isn't fulfilled, they should simply return the previous `state`.

andrevmatos: Also, we want to still update the `registerBlock` if it's different than the previous one. This shouldn't happen in practice, since we already check it's a `confirmed` action, but it's still valid from reducer's perspective.

PR #2103, opened 2020-08-21
dApp: Transfer screen redirect when no channels exist

### LINKED ISSUE #2039: Transfer screen is shown after the last channel has been settled
## Steps to Reproduce
1. Settle the last channel
2. Close the browser window
3. Connect to the dApp

## Expected Result
- Observe the transfer screen
- Dialog "You don't have channels any more" -> "Continue" button
- Afterward the "Start" screen is shown

    grafik

## Actual Result
- The transfer screen is shown
- If you click on "Channel" there is no channel shown
- If you click on "Change Token" there is no connected token shown

    grafik

```
- Browser & version
- Browser console errors
- Device / OS
```

taleldayekh:
Fixes #2039

**Short description**
Whenever all channels for all connected tokens are closed and settled a dialog appears warning the user about not having any connected open channels or connected tokens. The dialog has a button that redirects the user to the token select screen.

**Definition of Done**

1. Connect to a token and open a channel.
2. Close and settle the channel.

### DIFF
--- a/raiden-dapp/src/components/dialogs/NoChannelsDialog.vue
+++ b/raiden-dapp/src/components/dialogs/NoChannelsDialog.vue
@@ -1,0 +1,45 @@
+<template>
+  <raiden-dialog class="no-channels-dialog" :visible="visible" hide-close>
+    <v-card-title>{{ $t('transfer.no-channels-dialog.title') }}</v-card-title>
+    <v-card-action>
+      <v-img
+        class="no-channels-dialog__warning"
+        :src="require('@/assets/warning.svg')"
+        height="132px"
+        width="150px"
+      />
+      <action-button
+        :text="$t('transfer.no-channels-dialog.button')"
+        enabled
+        @click="navigateToTokenSelect()"
+      />
+    </v-card-action>
+  </raiden-dialog>
+</template>
+
+<script lang="ts">
+import { Component, Prop, Mixins } from 'vue-property-decorator';
+import RaidenDialog from '@/components/dialogs/RaidenDialog.vue';
+import ActionButton from '@/components/ActionButton.vue';
+import NavigationMixin from '../../mixins/navigation-mixin';
+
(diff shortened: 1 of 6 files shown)


raiden-dapp/src/components/dialogs/NoChannelsDialog.vue
```diff
@@ -0,0 +1,45 @@
...
+import RaidenDialog from '@/components/dialogs/RaidenDialog.vue';
+import ActionButton from '@/components/ActionButton.vue';
+import NavigationMixin from '../../mixins/navigation-mixin';
+
+@Component({
+  components: { RaidenDialog, ActionButton },
+})
+export default class NoChannelsDialog extends Mixins(NavigationMixin) {
+  @Prop({ required: true })
+  visible!: boolean;
```
weilbith, line 31: Maybe I lack in knowledge about the dialog, but I would rather put the render condition where used in the parent component and not here with a prop. Else I think we could end up in an awkward  mix of different pattern.

weilbith (approved):
Really good!
I like that we discussed together first to find the best solution. Well done.

Unfortunately I can't test it right now. And I also miss details on Vuetify. So can the dialog be closed else than clicking on the action button? Because that should not be possible. Optimistic approval that this works. :)

raiden-dapp/src/components/dialogs/NoChannelsDialog.vue
weilbith: Ignore this comment. I think this is perfectly fine.

taleldayekh:
> So can the dialog be closed else than clicking on the action button?

Usually, you can also close the dialog by clicking outside it but when the `hide-close` prop is set you can only close it by manually emitting a close event.

taleldayekh:
Thanks for taking the time to review @weilbith 👍

PR #2105, opened 2020-08-24
dApp: UDC deposit validation logic bug fix

### LINKED ISSUE #2099: Test UDC deposit and RDN on mainnet
Test UDC deposit and RDN on mainnet

taleldayekh:
Fixes #2099 

**Short description**
This fixes a bug in the amount input validation regex. The regex used to only check for integers which made the validation fail on values like `5.0` but not `5`.

The PR also adapts the validation logic for mainnet. When mainnet is used it checks whether the utility token balance that the user types is less or equal to the total available amount, whether the inputted amount is more than zero and whether it matches above mentioned regex.

**Definition of Done**

1. Works as expected on testnet, mainnet needs to be tested @agatsoh

### DIFF
--- a/raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
+++ b/raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
@@ -149,7 +149,7 @@ export default class UdcDepositDialog extends Vue {
 
     if (this.mainnet) {
       return (
-        utilityTokenAmount.lte(this.udcToken.balance) &&
+        utilityTokenAmount.lte(this.udcToken.balance as BigNumber) &&
         utilityTokenAmount.gt(Zero)
       );
     } else {
--- a/raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
+++ b/raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
@@ -144,19 +144,17 @@ export default class UdcDepositDialog extends Vue {
         this.udcToken.decimals
       );
     } catch (err) {
-      utilityTokenAmount = Zero;
+      return false;
     }
 
-    if (!this.mainnet) return utilityTokenAmount.gt(Zero);
-
-    const utilityTokenBalance = parseUnits(
-      this.utilityTokenBalance,
-      this.udcToken.decimals
-    );
-
-    return (
-      utilityTokenAmount.lte(utilityTokenBalance) && utilityTokenAmount.gt(Zero)
-    );
+    if (this.mainnet) {
+      return (
+        utilityTokenAmount.lte(this.udcToken.balance) &&
+        utilityTokenAmount.gt(Zero)
+      );
+    } else {
+      return utilityTokenAmount.gt(Zero);
+    }
   }
 
   async mounted() {
--- a/raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
+++ b/raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
@@ -135,7 +135,17 @@ export default class UdcDepositDialog extends Vue {
   }
 
   get valid(): boolean {
-    return /^[1-9]\d*$/.test(this.defaultUtilityTokenAmount);
+    const utilityTokenAmount = Number(this.defaultUtilityTokenAmount);
+    const utilityTokenBalance = Number(this.utilityTokenBalance);
+    const amountRegEx = /^\d+\.?\d*$/;
+
+    if (!this.mainnet) return amountRegEx.test(this.defaultUtilityTokenAmount);
+
+    return (
+      utilityTokenAmount <= utilityTokenBalance &&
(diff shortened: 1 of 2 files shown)


raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
```diff
@@ -135,7 +135,17 @@ export default class UdcDepositDialog extends Vue {
   }
 
   get valid(): boolean {
-    return /^[1-9]\d*$/.test(this.defaultUtilityTokenAmount);
+    const utilityTokenAmount = Number(this.defaultUtilityTokenAmount);
```
andrevmatos, line 138: It's not safe to convert these BigNumbers to Numbers directly. They can go up to 256bits in length, and JS vms only support up to 53 bits. Please, use `parseUnits` from `ethers/utils`, with the respective decimals, and compare it using BigNumber functions.

raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
```diff
@@ -135,7 +135,17 @@ export default class UdcDepositDialog extends Vue {
...
+    const utilityTokenBalance = Number(this.utilityTokenBalance);
+    const amountRegEx = /^\d+\.?\d*$/;
```
andrevmatos, line 140: You don't need regex if you're using the utility, it'll throw if it fails. You may need to handle though.

taleldayekh: I think I need the RegEx for the testnet since a user can enter any value they want?

andrevmatos: Let them, `parseUnits` should throw if it's invalid, you then just handle it.

raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
```diff
@@ -135,7 +136,27 @@ export default class UdcDepositDialog extends Vue {
...
+        this.udcToken.decimals
+      );
+    } catch (err) {
+      utilityTokenAmount = Zero;
+    }
+
+    if (!this.mainnet) return utilityTokenAmount.gt(Zero);
+
+    const utilityTokenBalance = parseUnits(
+      this.utilityTokenBalance,
```
andrevmatos, line 153: Notice `this.utilityTokenBalance` is a getter which formats the already BigNumber `this.udcToken.balance`. It doesn't make sense to parse it back to BigNumber here. Use `this.udcToken.balance` directly.

raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
```diff
@@ -135,7 +136,27 @@ export default class UdcDepositDialog extends Vue {
...
-    return /^[1-9]\d*$/.test(this.defaultUtilityTokenAmount);
+    let utilityTokenAmount: BigNumber;
+
+    try {
+      utilityTokenAmount = parseUnits(
+        this.defaultUtilityTokenAmount,
+        this.udcToken.decimals
+      );
+    } catch (err) {
+      utilityTokenAmount = Zero;
```
andrevmatos, line 147: Maybe if it fails to parse, you should directly `return false;`?

raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
```diff
@@ -135,7 +136,27 @@ export default class UdcDepositDialog extends Vue {
...
+    }
+
+    if (!this.mainnet) return utilityTokenAmount.gt(Zero);
```
andrevmatos, line 150: Nitpick: with the comment below (about taking out the balance constant), it'd be clearer here that the two conditions (on mainnet and not) would be in an `if/else` block. The way it is now, looks like the amount is only required to be positive if on testnet, and then the person needs to go to the other return to see the same condition is also checked there with something else (`amout <= balance`, since testnet can mint)

raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
taleldayekh: Ah, good point.

raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
taleldayekh: Yes, I'll do so.

raiden-dapp/src/components/dialogs/UdcDepositDialog.vue
taleldayekh: I see, yes, added an `if/else` block.

taleldayekh:
Thanks for the review @andrevmatos and for clarifying BigNumbers, implemented your suggestions.

PR #2109, opened 2020-08-25
Path new tests

### LINKED ISSUE #1616: Convert services unit tests to the new pattern
## Description
With #1377 / #1577 , there's a new standardized way to do epics unit async testing.
We should convert the services (path, monitoring) tests to use this new pattern. Check #1614 for the new flow. We should also convert the generic unit `raiden.spec.ts` on this issue.
New `ensure*` fixtures functions may be needed to ensure some common state.

## Acceptance criteria
- [ ] [path.spec.ts](https://github.com/raiden-network/light-client/blob/master/raiden-ts/tests/unit/epics/path.spec.ts) is converted to the new tests fixtures
- [x] [monitor.spec.ts](https://github.com/raiden-network/light-client/blob/master/raiden-ts/tests/unit/epics/monitor.spec.ts) is converted to the new tests fixtures
- [x] [raiden.spec.ts](https://github.com/raiden-network/light-client/blob/master/raiden-ts/tests/unit/epics/raiden.spec.ts) is converted to the new tests fixtures
- [ ] Tests don't use neither old `raidenEpicDeps` nor `epicFixtures` utilities, and instead rely on the new fixtures: `makeRaiden`, `makeRaidens`, `ensure*` and constants

agatsoh:

Fixes #1616

**Short description**
Please check these 3 tests 
1. `fail on failing matrix presence request` from the `PFS: pathFindServiceEpic1`  tests which is **failing**, I am looking into it but if you can give me an idea it would be great
2. `skip: channel is not open` from the `PFS: pfsFeeUpdateEpic1` which is passing but i am not sure whether I am testing the right thing
3. `skip: NO_MEDIATE` from the `PFS: pfsFeeUpdateEpic1` which is passing but I am not sure whether i am testing the correct thing.

I will create fixup commits for the review comments and changes that may come and i am sure there would be many.
Thanks a lot.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/tests/unit/epics/path.spec.ts
+++ b/raiden-ts/tests/unit/epics/path.spec.ts
@@ -6,8 +6,21 @@ import {
   providersEmit,
   makeAddress,
   waitBlock,
+  sleep,
+  MockedRaiden,
 } from '../mocks';
-import { epicFixtures, ensureChannelIsOpen, ensureChannelIsDeposited, deposit } from '../fixtures';
+import {
+  token,
+  tokenNetwork,
+  amount,
+  openBlock,
+  epicFixtures,
+  ensureChannelIsOpen,
+  ensureChannelIsDeposited,
+  deposit,
+  ensureTransferUnlocked,
+  ensureTransferPending,
+} from '../fixtures';
 
 import { Observable } from 'rxjs';
 import { first, toArray, pluck } from 'rxjs/operators';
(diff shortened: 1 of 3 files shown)


andrevmatos (approved):
Managed to cut in half the added lines, even before removing the outdated tests. Gave some cleanups on fixtures/functions used only on these suites. lgtm now. Please, merge if you agree.

agatsoh:
Looks good to me Andre i will merge this now

Issue #1618, opened 2020-06-02
Evaluate risks of keeping all the state in memory


kelsos:
## Description

In our current approach, the whole SDK state is kept in memory while it currently works fine and it is not a big problem, we keep adding things. As we keep adding more state that is kept in memory and persisted in local storage.

In https://github.com/raiden-network/light-client/issues/1607 the balance proofs received and send for each channel will be also added in the state. We have seen in the python client that extended usage keeps increasing the database size. For the light client, all this data is also active in memory.

We should evaluate any possible risks or problems that might arise from this approach.

## Acceptance criteria
- 

## Tasks
- [ ]

andrevmatos:
Thank you very much for creating this issue. First, I think it isn't going to be a problem for the foreseable future, as you said, since we only store very simple objects, and the on-chain operations gas and off-chain fees requirements would be limitants for attackers to abuse state space constraints. Proactively, we could and should put safeguards in place to prevent some specific peer/partner/sender/channel state becoming unreasonably big.
Ideally, the whole state would be stored out-of-memory, but that would not be practical/possible with redux. Also, some state is kind of required to be readily available, since they're synchronously needed for specific validations, like latest channel state. Beyond that, I think we must first identify members which have potential to grow limitlessly, and find ways to store them somewhere else. They're still part of state, and would need to be put together and re-hydrated on state download/upload, but at least most of it wouldn't be needed to be available in memory all the time and could be requested by some kind of async service.
 
Some of the best candidates could be:
- older (non-latest) balanceProof from us and partners (possibly needed for settleChannel)
- old (settled) channels (not currently used for anything, hard to be abused since requires onchain txs)
- completed sent and received transfers (unlocked off-chain, already unlocked on-chain, expired; needed only to avoid secret[hash] reuse and startup's emit for history)
- matrix rooms with non-partners (e.g. initiators; we could have a policy in place to un-monitor users we don't care for anymore, and possibly leave rooms with them and clean state; would help both with redux and matrix's state sizes)

[...]

PR #2110, opened 2020-08-26
Database

### LINKED ISSUE #1615: Convert transfers unit tests to the new pattern
## Description
With #1377 / #1577 , there's a new standardized way to do epics unit async testing.
We should convert the transfers tests to use this new pattern. Check #1614 for the new flow.
New `ensure*` fixtures functions may be needed to ensure some common state.

## Acceptance criteria
- [send.spec.ts](https://github.com/raiden-network/light-client/blob/master/raiden-ts/tests/unit/epics/send.spec.ts) is converted to the new tests fixtures
- [receive.spec.ts](https://github.com/raiden-network/light-client/blob/master/raiden-ts/tests/unit/epics/receive.spec.ts) is converted to the new tests fixtures
- [mediate.spec.ts](https://github.com/raiden-network/light-client/blob/master/raiden-ts/tests/unit/epics/mediate.spec.ts) is converted to the new tests fixtures

## Tasks
- [ ]

### LINKED ISSUE #1831: Staging Mediating Transfers Expires in the mediating node
## Steps to Reproduce
1. Perform a one hop mediated transfer
2. 

I have 3 nodes 
light client
`0xAa1Ab1fDA93e14c2d564B25Ba08c9d374F5F9541` Main Account
`0x9bf12572aBd794b031313fE286040nce727eC6191` Raiden Account
                                   .|
                                  V
`0x949e47d1E0C5351eCBb7C50bB5Dcf3fE915468e9` RC
                                   .|
                                  V
`0xdD29eA8d403365BcbFFd659a13d4A24370e2d940` RC

Mediated transfers remains Pending in the mediating node and then expires

## Expected Result
- Mediated Transfer should be successful

## Actual Result
- 
light_mediated_staging_failed_spinner
light_mediated_staging_failed

Raiden Python Client version:
`1.1.0 Sun Prince`
```
- Browser & version
- Browser console errors
- Device / OS
```

### LINKED ISSUE #1933: RaidenError: nonce mismatch
## Steps to Reproduce
1. Should have withdrawn previously 
2. transfer to the light client or simply connecting to the light client

## Expected Result
- The light client should be synched with the correct nonce

## Actual Result
- The light client gets 
```
{
  "type": "transfer/failure",
  "payload": {
    "details": {
      "condition": false
    },
    "name": "RaidenError"
  },
  "meta": {
    "secrethash": "0xff588f59cf4e76fce7e56936ce2fc00b92c922d3886a490d0a8c1a480bac19fb",
    "direction": "received"
  },
  "error": true
}
```
light_client_nonce_error

[raiden_2020-07-21T12_29_18.130Z.log](https://github.com/raiden-network/light-client/files/4953463/raiden_2020-07-21T12_29_18.130Z.log)

Initiator node log

[node-F128A4-mainnet.21thJuly2020.log.tar.gz](https://github.com/raiden-network/light-client/files/4953529/node-F128A4-mainnet.21thJuly2020.log.tar.gz)

Mediating node logs 

[node-1Fd883-mainnet.21thJuly2020.log.tar.gz](https://github.com/raiden-network/light-client/files/4953534/node-1Fd883-mainnet.21thJuly2020.log.tar.gz)

```
- Browser & version
- Browser console errors
- Device / OS
```

### LINKED ISSUE #2025: Consider increasing the time it takes to send a withdraw expired
https://github.com/raiden-network/light-client/blob/bc066c08a0adc02346f062fdc0d00c394448fe31/raiden-ts/src/transfers/epics/withdraw.ts#L304-L314

The receiver must only accept a withdraw expired message if the expiration block has been confirmed, otherwise it is possible for the withdraw to occur in a reorg which has not been seen by the receiver yet.

Because of the above, the sender has to wait at *least* confirmation blocks before sending the withdraw expired. However, sending withdraw expired at the exact moment a new block is seen by the sender is susceptible to race conditions, where the sender sees block `x` before the receiver. In this case the receiver will reject the withdraw expired message until it sees block `x`.

The problem above will eventually fix itself, since the sender will keep retrying the message until it it processed by the receiver. The problem here is not that the system won't work, but that this will artificially increase latency of the system. This is specially problematic because messages have to be processed in the order of the `nonce`, which means that any mediated transfer sent after the withdraw expired message will have to wait too.

The fix is simple, just introduce a bit of latency before sending the withdraw expired message from the sender, this way transfers won't suffer a hiccup.

### LINKED ISSUE #2044: Move state storage to IndexedDB/LevelDB

### LINKED ISSUE #2056: Consider increasing the time it takes to send a lock expired message
Just like with the withdraw #2025 there are race conditions in respect to the perceive latest block among both ends of a channel. Just like with the withdraw, the lock expired message can stall the transfers in a channel. Relevant code:

https://github.com/raiden-network/light-client/blob/27d308d6279ec7f8d797ea8fc5829738992d905c/raiden-ts/src/transfers/epics/expire.ts#L38-L40

### LINKED ISSUE #2077: registerBlock has to be verified
A lot of the code assumes that `TransferState`s with a non-zero `registerBlock` had a valid on-chain registration.

Examples:

https://github.com/raiden-network/light-client/blob/0abfd9ee9d4359b150048f7018dd13fe07175086/raiden-ts/src/transfers/epics/close.ts#L49-L54

https://github.com/raiden-network/light-client/blob/0abfd9ee9d4359b150048f7018dd13fe07175086/raiden-ts/src/transfers/epics/locked.ts#L614

https://github.com/raiden-network/light-client/blob/0abfd9ee9d4359b150048f7018dd13fe07175086/raiden-ts/src/transfers/epics/expire.ts#L40-L41

https://github.com/raiden-network/light-client/blob/59e7662e643e5b0a17db8b46a44d29b471b0263d/raiden-ts/src/transfers/state.ts#L36-L40

However, that is not sufficient. Not only the secret has to be registered on-chain, but the block at which the secret was registered, must be smaller than the `lock.expiration`. All of the above sites and any other places that use `registerBlock` without comparing to the `lock.expiration` have to be fixed.

### LINKED ISSUE #2085: off-by-one on secret epic
Follow up to https://github.com/raiden-network/light-client/issues/2077 .

The code below is accepting the secret registration were `lock.expiration == registration_block` as valid.

https://github.com/raiden-network/light-client/blob/0abfd9ee9d4359b150048f7018dd13fe07175086/raiden-ts/src/transfers/epics/secret.ts#L326-L343

However the smart contract does not:

https://github.com/raiden-network/raiden-contracts/blob/93230caa554f6f29e55b4521aafd0af20b710b1f/raiden_contracts/data/source/raiden/TokenNetwork.sol#L1679-L1686

The code should be changed from `gte` to `gt`.

andrevmatos:
Fixes #2044 
Fixes #1933 
Fixes #1831 
Fixes #2025 
Fixes #2056 
Fixes #2077 
Fixes #2085 
Fixes #1615 

**Short description**
Use PouchDB (backed by default by IndexedDB on browsers, LevelDown on NodeJS) as a persistent storage to store state.
Only pending transfers are kept in Redux (in memory) state, and synced to the database as soon as they change, by a persister middleware. This gives us the speed of in-memory access, the backends/adapters/sync power of Pouch, and memory-safety of not keeping everything in memory all the time and instead accessing historic data only if needed.
RaidenState/Channels is kept in memory (Redux) and just synced asynchronously to the `db` (per individual top level keys, and individual channels).
Transfers state, as they can grow boundless (#1618), is kept on database and retrieved on demand if needed using secondary indexes. This allows us to keep sub-0.2s transfers (cli-to-cli, nodejs, webrtc) while saving everything at real time to avoid losing state.
Legacy (once) and new async migration schemas are also implemented.

**Definition of Done**

1.
2.


andrevmatos:
This is too big to get reviewed. I'll merge as is, and we can test and narrow down any issues on our _development branch_ (`master`) ;)

PR #2112, opened 2020-08-27
dApp: Channel settlement notifications improvement

### LINKED ISSUE #1929: Formatting of channel settlement notification and message
## Description
The channel notification and message text need to get formatted nicely. We should also check, if the text is understandable. There are a lot of addresses within the notification.

grafik

## Acceptance criteria
- Addresses are truncated
- Icon is shown

## Tasks
- [ ]

taleldayekh:
Fixes #1929 

**Short description**
This is an updated `Channel Settlements` notification. In the snackbar we only display the title of the notification and the details on the notification card are limited to one sentence containing only the address of the partner whom the channel was settled (if a user wants any additional information they can always visit etherscan).

**Definition of Done**

1. Close and settle a channel.


raiden-dapp/src/store/notifications/notification-context.ts
```diff
@@ -3,4 +3,5 @@ export enum NotificationContext {
   INFO = 'info',
   WARNING = 'warning',
   ERROR = 'error',
+  SETTLE = 'notification_settle',
```
andrevmatos, line 6: It's weird to see `SETTLE` as a context/level besides `INFO`, `WARNING`, etc... Here, it seems to be needed only to change the icon. May I suggest to change slightly the logic here, add an optional `icon` property to the Notification payload, and have the levels as the standardized keywords, with respective color codes (e.g. blue for info, yellow for warning, red for error, etc), and use the default icons only if the payload one is missing? This way, we won't be required to implement a new 'context' every time we want a custom icon on a notification, and can just commit the action with the property in the payload.

taleldayekh:
Thanks for taking the time to review @andrevmatos , I will implement your suggestions regarding the context later today before merging.

Issue #2149, opened 2020-09-22
Add unit tests to the store notifications module


taleldayekh:
## Description
Currently the notifications module of the store completely lacks tests. We need to add test coverage for that part of the code base.

## Acceptance criteria
- Cover the notification store code with unit tests.

## Tasks
- [ ]

PR #2164, opened 2020-09-24
dApp: Channel open notifications adjustments

### LINKED ISSUE #2154: Feedback on Open Channel Progress
## Description

@taleldayekh Thanks a lot for the implementation of #1941. It improved a lot the usability.

Some in-iteration feedback on the current implementation: 

- [x] 1. The grey text on the notifciation is hard to read. What do you think about switching to white? The date/time could stay grey. CC @sashseurat 

- [x] 2. Sorting of the notifications should be descending (latest on top). It is the other way around.

    grafik

3. It says "Channel successfully opened" before the deposit has taken place. Is this by design? It might be better to create a new issue with the deposit progress. Would that work and make sense from your perspective?

- Created an issue: https://github.com/raiden-network/light-client/issues/2156

grafik

## Acceptance criteria
- 

## Tasks
- [ ]

taleldayekh:
Fixes #2154 

**Short description**
Small adjustment that addresses feedback for the newly introduces channel open notifications. The adjustments include:
- White text color.
- Sorting notifications newest first in the notification panel.

**Definition of Done**

1. Trigger at least two notifications.

### DIFF
--- a/raiden-dapp/src/components/notification-panel/NotificationCard.vue
+++ b/raiden-dapp/src/components/notification-panel/NotificationCard.vue
@@ -96,14 +96,21 @@ export default class NotificationCard extends Vue {
       flex-direction: column;
       margin-left: 16px;
 
+      &__block-count,
+      &__received {
+        font-size: 12px;
+      }
+
       &__title {
         color: $primary-color;
       }
 
-      &__received,
       &__block-count {
+        color: $color-white;
+      }
+
+      &__received {
         color: $secondary-text-color;
-        font-size: 12px;
       }
     }
   }
--- a/raiden-dapp/src/store/notifications/getters.ts
+++ b/raiden-dapp/src/store/notifications/getters.ts
@@ -8,9 +8,7 @@ import { NotificationImportance } from '@/store/notifications/notification-impor
 
 export const getters: GetterTree<NotificationsState, RootState> = {
   notifications: (state: NotificationsState): NotificationPayload[] => {
-    return Object.values(state.notifications)
-      .sort((a, b) => a.id - b.id)
-      .reverse();
+    return Object.values(state.notifications).sort((a, b) => b.id - a.id);
   },
   nextNotificationId: ({ notifications }: NotificationsState): number => {
     const id = Object.values(notifications).map(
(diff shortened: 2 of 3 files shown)


taleldayekh:
The store for notifications lacks test coverage. We have a separate issue for that #2149 . So if this gets approved I will force merge it.

raiden-dapp/src/components/notification-panel/NotificationCard.vue
```diff
@@ -96,14 +96,21 @@ export default class NotificationCard extends Vue {
...
+        font-size: 12px;
+      }
+
       &__title {
         color: $primary-color;
       }
 
-      &__received,
       &__block-count {
+        color: $color-white;
```
weilbith, line 109: Please consider to take advantage of Vueitfy and apply `white--text` to the template directly. This prevents us from doing this manually over and over again and we will still not produce a mass of classes nobody how to use them or when to delete what.
But this is just a general recommendation to use the styling here only for more special stuff. Vuetify comes with a ton of convenient helpers that leave our code more clean.

raiden-dapp/src/store/notifications/getters.ts
```diff
@@ -8,7 +8,9 @@ import { NotificationImportance } from '@/store/notifications/notification-impor
 
 export const getters: GetterTree<NotificationsState, RootState> = {
   notifications: (state: NotificationsState): NotificationPayload[] => {
-    return Object.values(state.notifications).sort((a, b) => a.id - b.id);
+    return Object.values(state.notifications)
+      .sort((a, b) => a.id - b.id)
```
weilbith, line 12: Not sure what is more complicated to understand:
- `sort((a, b) => a.id - b.id).reverse()`
- `sort((a, b) => b.id - a.id)`

It doesn't really matter. Just feels like unnecessary work for the CPU and one more line of code. I mean why not sort it directly correct? :thinking:

weilbith (approved):
I love such small and efficient PRs. :hugs: 

Just tiny ideas/inspirations. No need for further reviews.

raiden-dapp/src/store/notifications/getters.ts
taleldayekh: Ah, of course. That slipped my mind. Yes, let's get rid of the reverse function. :)

raiden-dapp/src/components/notification-panel/NotificationCard.vue
taleldayekh: Yeah, I know. But since it is already defined and we are mostly using the colors in the styles I'll just leave it as is. Also, I think it is a bit nice to know that I only need to look in the styles to find styles related code.

PR #2168, opened 2020-09-24
Transport tests new

### LINKED ISSUE #1614: Convert transport unit tests to the new pattern
## Description
With #1577 , there's a new pattern for simplified unit async tests:
- create 1-3 clients as needed by the tests with `await makeRaidens` from `mocks.ts`
- ensure they're in a given state (e.g. there's a funded channel between some of them)
- trigger some action or event
- optionally, await a little with `await waitBlock`
- `expect` some thing happened: an action was emitted, state was changed, a function was called
- use as much as possible constant fixtures or constant/awaited values from clients themselves instead of some 'global'/`describe`-level instances.

We need to convert tests in [transport.spec.ts](https://github.com/raiden-network/light-client/blob/master/raiden-ts/tests/unit/epics/transport.spec.ts) to use this new pattern. New `ensure*` fixtures utils functions may be created as needed.
Notice the new pattern also improves on mocked `MatrixClient`, by having it being created/logged in by transport's init themselves instead of a global mock, so we need to remove those `createClient` test-level mocks.

## Acceptance criteria
- `transport.spec.ts` is refactored according to the new unit async testing pattern
- any old unused mocks/fixtures is removed from tests

## Tasks
- [ ]

agatsoh:

Fixes #1614 

**Short description**
This pull request converts `transport.spec.ts` tests to new test pattern

I would request reviewer to check these 4 tests specifically 

1. **matrix throws if can not fetch servers list** 
I feel this test should not be there.

2. `matrixUpdateCapsEpic`

3. `success callee, receive message & channel error`
this test is flaky

4. `success caller & candidates`
this test is also flaky

Edit: 
5 **invite if there is room for user** it was hard for me to get this code running which calls the `matrix.invite` 
https://github.com/raiden-network/light-client/blob/master/raiden-ts/src/transport/epics/rooms.ts#L78-L82 

**Definition of Done**

1.
2.


agatsoh:
@andrevmatos when you have time you can let me know what do you think about the above mentioned tests. As well as overall tests in general. The tests might not pass mostly because of the tests highlighted in bold. So please let me know.

agatsoh:
Update as off 12Oct 2020
@andrevmatos 
Some tests i would like to specially bring your notice
1. I have not included the 
`matrix server config set same as stored setup`
tests in the new pattern because it would be similar to 
`matrix server config set without stored setup`
in the new pattern

2. **matrix throws if can not fetch servers list**
In this test it will always fail because the `if` condition is always satisfied and it never goes into the else part
https://github.com/raiden-network/light-client/blob/master/raiden-ts/src/transport/epics/init.ts#L356-L370 

3. accept & join from previous presence
```
expect(matrix.joinRoom).toHaveBeenCalledTimes(4);
```

raiden-ts/tests/unit/epics/transport.spec.ts
```diff
@@ -1828,3 +1836,1716 @@ describe('transport epic', () => {
...
+    ok: true,
+    status: 200,
+    text: jest.fn(async () => `- ${matrixServer}`),
+  }));
+  Object.assign(global, { fetch });
+
+  describe('initMatrixEpic', () => {
+    test('matrix stored setup', async () => {
+      expect.assertions(2);
+      const [raiden] = await makeRaidens(1);
```
andrevmatos, line 1857: You should use `makeRaiden()` when creating a single instance. `makeRaidens` is more to make a batch of mocked instances.

raiden-ts/tests/unit/epics/transport.spec.ts
```diff
@@ -1828,3 +1836,1716 @@ describe('transport epic', () => {
...
+      const matrix = (await raiden.deps.matrix$.toPromise()) as jest.Mocked<MatrixClient>;
+
+      const userId = matrix.getUserId()!;
+      expect(raiden.output).toContainEqual(
```
andrevmatos, line 1861: This test was supposed to test how the `initMatrixEpic` behaves when a matrix setup is already present in state (it means the client was already initialized before). To test/simulate that, the previous version `next`ed the setup directly before subscribing to the observable, and you missed it, so this test is running a first call, instead of this intended stored setup test.

raiden-ts/tests/unit/epics/transport.spec.ts
```diff
@@ -1828,3 +1836,1716 @@ describe('transport epic', () => {
...
 });
+
+describe('transport epic', () => {
+  const matrixServer = 'matrix.raiden.test'; // define later in fixture constant
+  const partnerRoomId = `!partnerRoomId:${matrixServer}`;
+  const processed: Processed = {
+    type: MessageType.PROCESSED,
+    message_identifier: makeMessageId(),
+  };
+  const fetch = jest.fn(async () => ({
```
andrevmatos, line 1847: Since some time ago, `fetch` is provided as part of the `mocks.ts`, you don't need/shouldn't re-create and re-assign it here. If you don't need the `json` method, you can make it optional in the type declaration there.

raiden-ts/tests/unit/epics/transport.spec.ts
```diff
@@ -1828,3 +1836,1716 @@ describe('transport epic', () => {
...
+          },
+        }),
+      );
+      expect(fetch).not.toHaveBeenCalled();
+    });
+
+    test('matrix fetch servers list', async () => {
+      expect.assertions(2);
+
+      jest.setTimeout(30000);
```
andrevmatos, line 1900: Don't set timeout inside a test. If you need to set it there, you must pass it as 3rd param to `test` declaration. But 30s is too much for a test anyway, you probably should decrease `httpTimeout`/`pollingInterval` config params if you have some timeout taking longer than it should inside your test, instead of increasing test timeout.

andrevmatos (changes requested):
There're 28 tests failing on this suite. I need it closer to passing before I can continue looking into it. Thanks

raiden-ts/tests/unit/epics/transport.spec.ts
agatsoh: I thought the `next`ed `matrixSetup` was the initialisation in  the previous test pattern. 
In the present pattern the
`const raiden = await makeRaiden(undefined);` 
already makes a `matrixSetup` action and after that there can no further  `matrixSetup` actions.

andrevmatos: `makeRaiden` will output a `matrixSetup` anyway, as it's part of its initialization, but this test wants to test how it behaves if the state already contains a populated `state.transport` setup, and this is set by that early next'd `matrixSetup`, which is missing on the new test.

raiden-ts/tests/unit/epics/transport.spec.ts
agatsoh: Just made required changes

raiden-ts/tests/unit/epics/transport.spec.ts
agatsoh: Also made the changes

raiden-ts/tests/unit/epics/transport.spec.ts
agatsoh: Also made the required changes but would wait for your review if anything else is required

agatsoh:
@andrevmatos I have rebased and adapted the new tests for nonce mismatch PR, please have a look

andrevmatos (approved):
LGTM! I moved some things around, fixed the erroring test and improved some which were not testing the right thing, but most of it was correct, thank you for implementing this! Nice job.

PR #2170, opened 2020-09-25
dApp: redirect on state upload when connected

taleldayekh:
**Short description**
When state has been uploaded the dApp directly connects the user.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-dapp/src/components/account/backup-state/UploadStateDialog.vue
+++ b/raiden-dapp/src/components/account/backup-state/UploadStateDialog.vue
@@ -140,6 +140,7 @@ export default class UploadStateDialog extends Mixins(NavigationMixin) {
   async connectAndRedirect(retrievedState: string) {
     let { useRaidenAccount } = this.settings;
 
+    /* istanbul ignore next */
     await this.$raiden.connect(
       retrievedState,
       useRaidenAccount ? true : undefined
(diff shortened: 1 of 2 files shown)


raiden-dapp/src/components/account/backup-state/UploadStateDialog.vue
```diff
@@ -136,14 +137,24 @@ export default class UploadStateDialog extends Vue {
     }, 2000);
   }
 
-  uploadStateAndConnect(uploadedFile: FileList) {
+  async connectAndRedirect(retrievedState: string) {
```
weilbith, line 140: Nitpick: 
Could you eventually move this function down? This will only be called by the `uploadState()` function which gets defined after, if I'm not wrong. At least I prefer to have the functions in a module a bit more sorted in the order they are used (some linter even force that).

weilbith (approved):
It works in practice. :+1: 
Just a nitpick, else very good. I like the function split especially. :grimacing:

raiden-dapp/src/components/account/backup-state/UploadStateDialog.vue
taleldayekh: Cool, yes I will do that.

PR #2187, opened 2020-09-30
dapp: add new feature - optional identifier query param in route

### LINKED ISSUE #1255: I want to add a custom payment identifier to my transfer
## User Story
As Chris, I want to add a custom payment identifier to my transfer so that the target can identify the payment.

## Acceptance criteria
- 

## Tasks

ezdac:
Fixes #1255

**Short description**

Adds an optional `identifier` query parameter to the `TransferStepsRoute` url,
that overwrites the automatically generated `paymentId` when a value is provided.

**Definition of Done**

1. Initate a transfer via the transfer route with an additional `identifier` query param, e.g. `#/transfer/0xC563388e2e2fdD422166eD5E76971D11eD37A466/0x11faFfcc7Aa31d731a8Cf6e998216731ed78D7AF?identifier=640467093185029602&amount=0.001` - this has to be a `uint64`
2. Verify on the receiver's end that the transfer has the desired identifier, e.g. `640467093185029602`

### DIFF
--- a/raiden-dapp/src/utils/query-params.ts
+++ b/raiden-dapp/src/utils/query-params.ts
@@ -1,4 +1,13 @@
 import AddressUtils from '@/utils/address-utils';
+import { BigNumber, bigNumberify } from 'ethers/utils';
+
+export function getPaymentId(queryParam: any): BigNumber | undefined {
+  try {
+    return bigNumberify(queryParam);
+  } catch (_error) {
+    return undefined;
+  }
+}
 
 export function getAmount(queryParam: any): string {
   let amount = '';
--- a/raiden-dapp/tests/unit/query-params.spec.ts
+++ b/raiden-dapp/tests/unit/query-params.spec.ts
@@ -1,4 +1,5 @@
-import { getAddress, getAmount } from '@/utils/query-params';
+import { getAddress, getAmount, getPaymentId } from '@/utils/query-params';
+import { bigNumberify } from 'ethers/utils';
 
 describe('query params', () => {
   describe('amount', () => {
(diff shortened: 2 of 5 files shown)


raiden-dapp/src/utils/query-params.ts
```diff
@@ -1,4 +1,18 @@
...
+  if (queryParam && typeof queryParam === 'string') {
+    try {
+      paymentId = bigNumberify(queryParam);
+    } catch (error) {
+      return undefined;
+    }
+  }
+
+  return paymentId;
+}
```
weilbith, line 15: Nitpick:
This function isn't doing much. Consider to simplify it.

```suggestion
export function getPaymentId(queryParam: any): BigNumber | undefined {
  try {
    return bigNumberify(queryParam);
  } catch (_error) {
    return undefined;
  }
}
```

raiden-dapp/tests/unit/query-params.spec.ts
```diff
@@ -34,4 +35,19 @@ describe('query params', () => {
       expect(getAddress(undefined)).toBe('');
     });
   });
+
+  describe('paymentId', () => {
+    test('invalid param returns undefined', () => {
+      expect(getPaymentId('onehundred')).toBe(undefined);
+    });
+
+    test('valid string returns BigNumber', () => {
```
weilbith, line 44: Just to be correct:
If you agree to simply the `getPaymentId` function as suggested above you can make this more generic because it accepts everything that `bigNumberify` can handle, not only strings. But ofc we don't want to test the function of our dependencies, just make sure that formats we want to work are actually working. This could be 1-2 different string formats.

weilbith (approved):
Thanks for your contribution! :upside_down_face: 
Just a small suggestion how to improve the code slightly. :man_shrugging:

raiden-dapp/tests/unit/query-params.spec.ts
ezdac: Added hex encoded string test

PR #2192, opened 2020-10-01
sdk: allow to replace db equal to current state

weilbith:
It is confusing if the user attempts to backup his state and uploads it right after. This gets denied so far.


andrevmatos (approved):
lgtm, ty

Issue #1584, opened 2020-05-28
Implement a first basic scenario with Cypress for the dApp


kelsos:
## Description

This issue is a followup to #1404. 
In #1580 the basics are done and Cypress works with the integration image.
The next step for us is to cover the basic dApp functionality with tests.

## Acceptance criteria
- Have tests for basic actions in the dApp
- Runs in the CI
- open channels
- deposit to UDC
- make a direct transfer
- request a route
- make a mediated transfer

## Tasks
- [x] Write down the scenario @christianbrb 
- [x] Cypress implementation @taleldayekh 
- [ ] Cleaning up the cypress code @taleldayekh 
- [x] Integrating in the CI @weilbith 

## Additional information
- Good architecture overview: https://www.toolsqa.com/cypress/what-is-cypress/
- Cypress best practises: https://docs.cypress.io/guides/references/best-practices.html

christianbrb:
Start with the basic main scenarios

taleldayekh:
I guess we need to discuss what the basic scenarios are and how to organise the tests.

weilbith:
@taleldayekh you assigned me as well. What tasks do you expect I should start to work on?

taleldayekh:
@weilbith I assigned you because I guess we will be having some discussions on how to best implement this stuff. :) But if you want you could resarch the CI integration.

weilbith:
Sounds good. Ping me when you have something ready I could try to make work with CI.

christianbrb:
I was thinking about implementing the following scenario, but we would need to discuss how and if we break it down.

The scenario has been build under the assumption, that those are high level integration tests and not pure UI tests.

1. Open the dApp with the address, token and token amount as a paramter: https://lightclient.raiden.network/#/connect/0xE2b702eD684bEb02850ac604278f078A4ce8b6E6/0x1F916ab5cf1B30B22f24Ebf435f53Ee665344Acf?deposit=0.01
2. Accept the disclaimer
3. Connect
4. Open Channel
5. Navigate to the notification panel
6. Delete the notification
7. Make a direct transfer using the route with parameters (eg. https://lightclient.raiden.network/#/transfer/0x59105441977ecD9d805A4f5b060E34676F50F806/0x1F916ab5cf1B30B22f24Ebf435f53Ee665344Acf?amount=0.001)
8. Go to the account menu
10. Backup the state
11. Disconnect
12. Delete the browser store
13. Upload the state
14. Connect
[...]

PR #2201, opened 2020-10-06
Bugfix/sdk switch integration and e2e tests

weilbith:
Related to #1584

**Short description**
How to explain this shortly... As I talked with Andre is has turned out that the integration and end-to-end tests of the SDK have evolved awkwardly over the time. So are today the integration tests actually the end-to-end tests and vice-versa. This leads to confusion when we now start to add end-to-end tests for the dApp. Also is the general handling like in the CI configuration confusing. The unit and e2e tests run in the default workflow. But the integration tests nightly. Read more in my commit message(s) for further details or examples.
After all Andre agreed that is makes sense to just switch both names. So the what is now called integration tests for the SDK become the e2e tests and the over way around. After all the tests and everything functional wise hasn't changed by this naming. But it allowed to clean up this awkward naming conflict that has evolved with the dApp tests now.
This PR includes mainly renaming. Adjusting docs and references according to it. But also small improvement so the bash script (using a linter) have been done, as well as a convenient npm script for the developer. **Please review this PR commit by commit**. Due to the limitations of how Git works this is pretty confusing else (moving, renaming files and change their content does not get presented well).

_MISSING_ is the renaming of the DockerHub repository. This is still called `raidennetwork/lightclient-integration`. According to the new defined naming it should be a `raidennetwork/lightclient-e2e-environment`. Also do we need to add a tagging system for this image, since CircleCI is about to restrict `latest` pulls.

**Definition of Done**

1.
2.


raiden-ts/tests/e2e/e2e.spec.ts
```diff
@@ -55,24 +55,24 @@ async function createRaiden(account: number | string | Signer): Promise<Raiden>
 }
 
 const wait = async (timeInMs: number): Promise<void> =>
-  new Promise((resolve) => setTimeout(() => resolve(), timeInMs));
```
andrevmatos: Nitpick: these parenthesis changes are against our current linter rules. Do you have up-to-date packages?

raiden-ts/tests/e2e/e2e.spec.ts
```diff
@@ -55,7 +55,7 @@ async function createRaiden(account: number | string | Signer): Promise<Raiden>
 }
 
 const wait = async (timeInMs: number): Promise<void> =>
-  new Promise(resolve => setTimeout(() => resolve(), timeInMs));
```
andrevmatos: Oh, it's done already. Thanks

andrevmatos (approved):
lgtm, thank you

weilbith:
I'm so ignorant and disregard what CodeClimate is reporting. It is quite picky and not related to this PR. Apparently it was once accepted like this. If it pops-up again we might handle it.

PR #2202, opened 2020-10-06
Feature/dapp cypress jest plugin

weilbith:
Related to #1584 

**Short description**
We use _Jest_ for all tests so far. Cypress is based on Mocha with Chai as assertion library. This can lead to confusion when switching between the test. It also feels like a bad practice to mix those frameworks. Unfortunately is the issue on Cypress to support Jest native still open. I'm following.  The plugin added here brings at least the Jest kind of `expect` assertions back to Cypress. Unfortunately does Cypress has the concept of asynchronous `Chainer`s. They are probably a great design approach, but make the usage of element retrieved from the DOM cumbersome, using this plugin. 
@taleldayekh Please evaluate if you are fine with this syntax with the arrow functions. For all other assertions where we have an actual object we can use `expect()` straight forward. Not sure how often this will be the case. If you dislike it and say it leads to an awkward mix, just close this PR. I'm split in my opinion because of the reasons in the above paragraph.

Eventually can [this plugins command](https://github.com/Lakitna/cypress-commands/blob/develop/docs/to.md) make things easier to use...

**Definition of Done**

1.
2.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -127,6 +127,7 @@ importers:
       canvas: 2.6.1
       copy-webpack-plugin: 6.1.1_webpack@4.44.2
       cypress: 3.8.3
+      cypress-jest-adapter: 0.1.1_jest@26.4.2
       eslint: 7.10.0
       eslint-import-resolver-alias: 1.1.2_eslint-plugin-import@2.22.1
       eslint-plugin-import: 2.22.1_eslint@7.10.0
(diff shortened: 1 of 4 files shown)


taleldayekh (approved):
I don't mind the arrow functions. It is nice to use the same as the other tests. I guess it could get a bit confusing when one is looking for references in the Cypress docs but I'm sure that is pretty easily sorted out. Let's go with this, I agree with you that it is nice to stick to one framework. Thanks for doing the research.

PR #2206, opened 2020-10-08
Adapt capabilities to query string format

### LINKED ISSUE #1975: Capability parsing is trying to handle quote marks using regex
https://github.com/raiden-network/light-client/blob/126f898409667c437c952a01a40713aad735ceb6/raiden-ts/src/transport/utils.ts#L58-L59

regular expressions can not properly handle balanced characters. This should be replaced with a parser that properly handles balanced quote marks.

Two other alternatives are just removing the feature altogether or using a negative lookbehind to ignore escaped commas `(?<!\\),`

Edit: The value parsing is also not handling edge cases properly:

- white space is not handled consistently:
  - All the prefix white space is ignored for the keys `\s*`, however no post-fix white space is ignored because `([^=]+)` is greedy
  - Values don't ignore either prefix nor post-fix white spaces inside quotes
  - Values allow for exactly one-or-zero white space as a prefix before the quotes, while keys allow for zero-or-more, and values ignore all white space after the quotes
- the value part of the regex is using the non-greedy operator `(.*?)` but at the same time it anchors the match at the end of the string `"?\s*)?$`, this just makes the regex more complicates the regex.

My suggestion here is to simply split by the first escaped equal sign which is not escaped.

Perhaps it would make sense to just use json for this data?

Edit2: The regex above is not handling escaping of the `\` character properly, it was just a suggestion, if this is implemented that should be taken care of.

### LINKED ISSUE #2205: Adapt capabilities to new format
## Description
On https://github.com/raiden-network/raiden/issues/6503, capabilities format was discussed in order to provide better backward and forward compatibility, by using query string encoded URL.
It also replaces the negative capabilities with their positive form and explicit defaults for when missing.
We must adapt our format in order to be compatible with python's implementation.
Supersedes #1975 
This breaks compatibility with previous LC's capabilities, but since capabilities fallback should always be backwards compatible, Raiden-protocol-wise, it's a non-breaking change.

## Acceptance criteria
- Capabilities are adapted to new format.

## Tasks
- [ ]

andrevmatos:
Based on #2174 
Fixes #2205 
Fixes #1975 

**Short description**
Implement the new capabilities format and sane fallbacks.

**Definition of Done**

1.
2.


raiden-ts/src/transport/utils.ts
```diff
@@ -33,33 +34,52 @@ export const getPresences$: (action$: Observable<RaidenAction>) => Observable<Pr
...
  * @returns stringified version of caps
  */
 export function stringifyCaps(caps: Caps): string {
-  return Object.entries(caps)
-    .filter(([, v]) => typeof v !== 'boolean' || v)
-    .map(([k, v]) => (typeof v === 'boolean' ? k : `${k}="${v}"`))
-    .join(',');
+  const url = new URL('mxc://raiden.network/cap');
+  // URLSearchParams.append can handle all primitives
+  const appendParam = (key: string, value: string | number | boolean | null) =>
```
weilbith, line 47: Isn't the value type of `CapsPrimitive`?

raiden-ts/src/transport/utils.ts
```diff
@@ -33,33 +34,52 @@ export const getPresences$: (action$: Observable<RaidenAction>) => Observable<Pr
...
   if (!caps) return;
-  const result: { [k: string]: string | boolean } = {};
+  const result: Mutable<Caps> = {};
   try {
-    // this regex splits by comma, but respecting strings inside double-quotes
-    for (const cap of caps.split(/,(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)/g)) {
-      const match = cap.match(/^\s*([^=]+)(?: ?= ?"?(.*?)"?\s*)?$/);
-      if (match) result[match[1]] = match[2] ?? true;
-    }
+    const url = new URL(caps!);
```
weilbith, line 67: Why is the force here necessary? Shouldn't TypeScript now that there is the `if` condition before? At least when I remove the `!` locally I don't get any error. (I verified that when I remove the if-line I get the error)

raiden-ts/src/transport/utils.ts
```diff
@@ -33,33 +34,52 @@ export const getPresences$: (action$: Observable<RaidenAction>) => Observable<Pr
...
+        const lowValue = value.toLowerCase();
+        if (lowValue === 'none' || lowValue === 'null') resValue = null;
+        else if (lowValue === 'false') resValue = false;
+        else if (lowValue === 'true') resValue = true;
+      }
+      if (key in result) {
+        let prevValues = result[key];
+        if (!Array.isArray(prevValues)) result[key] = prevValues = [prevValues];
+        prevValues.push(resValue);
+      } else result[key] = resValue;
```
weilbith, line 82: Reads confusing. First there will be always the else case. Only from the second iteration on there might be the chance to have the if case. Could you switch that. I would read better and the logic express itself better as well. My opinion. Nitpick category.

raiden-ts/src/transfers/epics/secret.ts
```diff
@@ -418,7 +419,12 @@ export const transferAutoRegisterEpic = (
         takeUntil(grouped$.pipe(ignoreElements(), endWith(1))),
       ),
     ),
-    takeIf(config$.pipe(pluck('caps', Capabilities.NO_RECEIVE)), true),
+    takeIf(
+      config$.pipe(
+        pluck('caps'),
+        map((caps) => getCap(caps, Capabilities.RECEIVE)),
+      ),
+    ),
```
weilbith, line 427: Is there a negation missing?

weilbith (approved):
Looks pretty good. I like the positive capabilities way more. :+1:

raiden-ts/src/transfers/epics/secret.ts
andrevmatos: The 2nd param for this `takeIf` operator is `negate`, and is missing now.

raiden-ts/src/transport/utils.ts
andrevmatos: Nice catch, yes, didn't notice it. Thanks!

raiden-ts/src/transport/utils.ts
andrevmatos: yes, but `CapsPrimitive` is a codec, not a type.. I wanted to avoid exporting it just because of this, but thinking better, I'll do it. Thanks

PR #2212, opened 2020-10-08
Check for overflow when creating transfer

### LINKED ISSUE #2078: Missing overflow check for lock.amount
Similar to #2058, #2059, and #2063 

https://github.com/raiden-network/light-client/blob/0abfd9ee9d4359b150048f7018dd13fe07175086/raiden-ts/src/transfers/epics/locked.ts#L137

Is not checking for overflow.

palango:

Fixes #2078

**Short description**

Add am overflow check for the amount of the locked transfer before sending it.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/transfers/epics/locked.ts
+++ b/raiden-ts/src/transfers/epics/locked.ts
@@ -44,7 +44,7 @@ import { RaidenEpicDeps } from '../../types';
 import { isActionOf } from '../../utils/actions';
 import { LruCache } from '../../utils/lru';
 import { pluckDistinct } from '../../utils/rx';
-import { Hash, Signed, UInt, Int, Address, untime } from '../../utils/types';
+import { Hash, Signed, UInt, Int, Address, untime, decode } from '../../utils/types';
 import { RaidenError, ErrorCodes } from '../../utils/error';
 import { Capabilities } from '../../constants';
 import {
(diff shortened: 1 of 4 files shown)


raiden-ts/src/transfers/epics/locked.ts
```diff
@@ -131,8 +131,14 @@ function makeAndSignTransfer$(
     'expiration too soon',
   );
 
+  // fee is added to the lock amount
+  const amount_with_fee = decode(
```
andrevmatos, line 135: Please, use `camelCase` in our codebase. `snake_case` is allowed only because and on external APIs and messages which already use it.

raiden-ts/tests/unit/epics/send.spec.ts
```diff
@@ -115,6 +115,37 @@ describe('send transfer', () => {
     expectChannelsAreInSync([raiden, partner]);
   });
 
+  test('overflow in transfer value is handled', async () => {
+    // expect.assertions(6);
+
+    const max = Two.pow(256) as UInt<32>;
```
andrevmatos, line 121: Nitpick: `2^256` is already outside the `UInt<32>` range. Could you use something smaller (maybe `2^256-1`), but which would fail only after the sum?

andrevmatos (approved):
Two smallish comments, so I'm preemptively approving, you can merge after fixing them.
The linked issue mentions only this line (which I'd say is the main case), but did you check there's no other instances where this kind of cast could cause security issues?
Thanks

palango:
Thanks for the review! I'll fix the points. As for the other locations: I saw there were more issues, so I just wanted to this to get up to speed.

Do you have any idea why the tests would fail? I can't really interpret the errors.

andrevmatos:
@palango to ease reading the logs (specially when it was a single failure, like your error), I suggest piping the url from `Open Raw Step Output` CircleCI action with `curl` to `less`, and backwards-searching (`?`) for `FAIL`, it should take you directly to the failed test logs.
In this case, it seems the expected action simply wasn't there, and if you look into the end of the `output` array, there's no `transfer.success` or `transfer.failure` for your request. This is probably caused by you violating some expectation with the type assertion, by telling the compiler to trust you that BigNumber is a `UInt<32>`, while it's just outside of the range. Use a value in the range, and it should pass, the rest of the test seem ok.

PR #2215, opened 2020-10-09
Hotfix deployment of latest v0.11.1 without mainnet

weilbith:
Unfortunately our CI pipeline is broken and has built the latest release
in development mode. This for examples allows the user to connect to the
main network which is not allowed yet.


weilbith:
@christianbrb @taleldayekh I tested it locally, but let's be super secure: please test this locally. Serve pr does not work. Instructions follow.

taleldayekh (approved):
Tested locally and works as expected. 👍

PR #2225, opened 2020-10-09
sdk: improve default capability composition

andrevmatos:
Fixes #

**Short description**
Just a small improvement to capabilities: composes the default and user defined ones, so the user defined caps object is merged with the default, instead of replacing it. This allows us to only set caps which are different from default. Only applies to advanced, direct call of the config API, since these aren't exposed in the dApp.

**Definition of Done**

1.
2.


agatsoh (approved):
The added code looks more correct than previous one so LGTM

PR #2226, opened 2020-10-10
sdk: apply optional chaining operator only on optional properties

andrevmatos:
Fixes # 

**Short description**
Refactor a few instances of the optional chaining operator to make it more strictly typed:
- If you have `foo?.bar?.baz`, where `foo` can be `undefined`, but if defined, `bar` is always a defined property of it, containing an always defined `baz` property, you don't need the second opt-chain op: the whole `foo?.bar.baz` will type-safely evaluate to `undefined` if `foo` is not defined, and to `foo.bar.baz` if it is defined.
  - Previously, there were several instances where we applied `?.` to all further properties on a chain, if a previous one could be undefined; this PR fixes this by applying the optional chain operator only to path members which can be undefined, then more strictly checking the ones expected to be defined
  - This also applies to functions/methods which are always defined if their object is (i.e. fixing unneeded `method?.(...)` usages)
- With the above also applied to the optional `deposit` param for the `Raiden.openChannel` method, this fixes an issue on BF2, where a `deposit=0` was passed to `openChannel`, which made it wait forever for a `channelDeposit.success` call which never happened because it was filtered out due to deposit=0 being noop.

**Definition of Done**

1. Tests pass
2. BF2 pass

### DIFF
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -50,7 +50,7 @@ import { isActionOf } from '../utils/actions';
 import { pluckDistinct, distinctRecordValues, retryAsync$, takeIf } from '../utils/rx';
 import { fromEthersEvent, getNetwork, logToContractEvent } from '../utils/ethers';
 import { encode } from '../utils/data';
-import { RaidenError, ErrorCodes } from '../utils/error';
+import { RaidenError, ErrorCodes, assert } from '../utils/error';
 import { createBalanceHash, MessageTypeId } from '../messages/utils';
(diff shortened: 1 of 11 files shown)


agatsoh (approved):
The `?.` check has been removed from all the property chains for both the properties as well as functions. So LGTM

Only one comment are these the only file where such changes are necessary or there are more files where such changes might be required.?

andrevmatos:
I searched for every instance of more than one `?.` operators in the same line and inspected each if it was really necessary. Did the same for the function calls. Should be all of it, although something multiline could have slipped.

PR #2232, opened 2020-10-13
dApp: Cypress test scenarios

taleldayekh:

Fixes

**Short description**

**Definition of Done**

1.
2.


raiden-dapp/tests/e2e/specs/scenario.spec.ts
```diff
@@ -1,73 +1,18 @@
...
-  navigateToUDC();
-  withdrawUDCTokens();
-  navigateToDownloadLogs();
+describe('dApp e2e tests', () => {
+  before(async () => {
+    const allIndexedDBs = await window.indexedDB.databases();
+    allIndexedDBs.forEach((indexedDB) => {
+      window.indexedDB.deleteDatabase(indexedDB.name);
+    });
+  });
```
weilbith, line 54: I thought it is recommended to do such things in `support/index.ts`? :thinking: 
There is works for every test suite automatically. Does that not work? Or do you prefer this solution because of other reasons? :smiley:

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 logs
+.DS_Store
```
weilbith, line 3: Out of curiosity: what is this?

weilbith (approved):
Thanks for investing all the time to make this possible. I think these tests are super important for a secure continuous development and will eventually safe our ass a couple of times. I think you did a great job in terms of writing informative commit messages. They really helped me while reading the code.

I removed a couple of comments I made on all of the navigation and user interaction functions. I think you did a good job in splitting them all up into meaningful small chunks. These comments where like: consider to add a function parameter for the amount or rename the function to make clear its requirements (being on which screen or having no tokens or..). But I think that is not how those functions are meant. They are basically not meant to get re-used, but rather just collect a bunch of instructions that belong together and make the huge flow better to handle. So for the moment I agree that this is probably the best solution. Eventually it just needs some time to get used to how this super big singe test case works.

PS: Did you opened an issue for the adoptions for the Jest plugin? Or do we abandon it completely?

PPS: I'll temporally push a commit to enable the nighly workflow on CI, actually testing the end-to-end tests. Therefore I'll also upload the new Docker image to the hub.

weilbith:
@taleldayekh you can remove the commit causing the merge conflict. This is already solved on `master.

I'm just about to fix local issues to re-build the Docker image and run the tests locally before I push to DockerHub and then let the nighlies run.

weilbith:
The test was successful! :+1: 
Don't wonder that the SDK E2E fail. Their fix is in master already.

@taleldayekh when you eventually react to my comments at the review and then clean up the history, please make the nightlies functional again.

taleldayekh:
> PS: Did you opened an issue for the adoptions for the Jest plugin? Or do we abandon it completely?

I'll open an issue for that, together with issues for closing channel test and mediated transfers test.

raiden-dapp/tests/e2e/support/index.ts
```diff
@@ -1,22 +1,7 @@
...
-// Import commands.js using ES2015 syntax:
 import '@cypress/code-coverage/support';
 import './commands';
 import 'cypress-jest-adapter';
 
-// Alternatively you can use CommonJS syntax:
-// require('./commands')
+Cypress.on('uncaught:exception', () => {
+  return false;
+});
```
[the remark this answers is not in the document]
taleldayekh: Good point, I'll do that!

raiden-dapp/tests/e2e/specs/scenario.spec.ts
taleldayekh: I did not try adding it in the `support/index.ts`. I read in the docs and went with the above solution. Also, it is kinda nice to clear the databases in the before block because then I only have to open the test file to understand what is happening instead of poking around in the support dir. :)

.gitignore
taleldayekh: Haha, it's a Mac-specific file that stores custom configurations for directories.

taleldayekh: It's hidden and exists on all Macs.

weilbith: Please consider to put such stuff into your local `~/.config/git/gitignore` file. But since we have multiple Mac users lets keep it here.

taleldayekh: Ah, ok 👍

Issue #1790, opened 2020-06-24
Investigate the resilience of epic on web3 provider errors


kelsos:
## Description

While there should not be many places where that happens, we have to ensure that there is no possibility of specific epics completing when a web3 error happens during contract interaction.

There is at least one instance of the UDC withdrawal code where this might happen.

If the issue should be retried we should ensure that we do that, and recover so that the epics keep working normally.

If the issue is not recoverable a shutdown should happen.

## Acceptance criteria
- All inner and outer observables handle the web3 smart contract errors 
- The user is getting informed if it is not recoverable in a dialogue

## Tasks
- [ ] Write down the tasks (every epic or topic - eg. channels)

christianbrb:
Proposal: Disconnect from Wifi and check when the SDK is crashing

Issue #1256, opened 2020-03-30
Disable receiving if blocks don't arrive in a timely manner


christianbrb:
## Description
Disable receiving if blocks don't arrive in a timely manner

Related https://github.com/raiden-network/raiden/issues/5955

Thanks @Dominik1999 

## Acceptance criteria
- It is exposed as an interface
- It is not needed for sending
- Receiving is disabled, if the node is out of sync
- If the node is back in sync enable receiving

## Tasks
- [ ]

PR #2236, opened 2020-10-13
Convert epics to non-arrow functions and fix some network errors

### LINKED ISSUE #2227: cli: fix wrtc bugs biting us
[wrtc](https://github.com/node-webrtc/node-webrtc) is a NodeJS native module providing WebRTC on this environment. It's very fast (written in C++) and provides the same interface as browser's WebRTC objects, allowing us have a single codebase supporting both environments. There're though some bugs on this module which can become a bigger issue if `raiden-cli` is intended to be a major client on servers, and we need them fixed.

Most of the times, WebRTC data channels can be established and remain like that for a long time, but on cli + wrtc, sometimes it disconnects and can't get connected again. Also, this seems to take down all other WebRTC connections on that node as well, and the process starts consuming 100% of one CPU core.

Fortunately, [we managed to detect this state](https://github.com/raiden-network/light-client/blob/48da3d7b6f946942f29bd7d5c7f3252847a2e459/raiden-ts/src/transport/epics/webrtc.ts#L358-L365) and discard the channel (`connectionState` becomes `failed`), falling back to matrix messages (i.e. the client and protocol continue usable, although wasting one core of CPU/power), so the client remains responsive, but it's a pity to lose WebRTC randomly for the rest of a session.

The entry-point (and possibly a cause or lead to the cause) can be seen as a segfault [if we try to close the connection](https://github.com/raiden-network/light-client/blob/48da3d7b6f946942f29bd7d5c7f3252847a2e459/raiden-ts/src/transport/epics/webrtc.ts#L406-L410). Maybe the 100% CPU issue is because the datachannel gets closed but the peer connection doesn't.

This seems to be more commonly triggered on high-concurrency scenario, like BF6, or a longer-running one, like BF1. When you see a `node` process starting consuming 100% of one core, you can check the logs and see a `RTC: connection failed` log and the retry loop for the webRTC kicking in (and looping each 5-60s, failing and retrying).

We should try to fix this bug ourselves upstream, pay for someone to fix it, or put a bounty out to get it fixed.

Some possibly related upstream bugs:
- https://github.com/node-webrtc/node-webrtc/issues/576
- https://github.com/node-webrtc/node-webrtc/issues/458
- https://github.com/node-webrtc/node-webrtc/issues/636
- https://github.com/node-webrtc/node-webrtc/issues/645

## Steps to Reproduce
1. Run BF1 and/or BF6 scenarios, check logs when/if one `node` process starts consuming 100% CPU.
2. Possibly related to the same bug, or a different one: uncomment that `connection.close()` line, get a client connected via webRTC with another (e.g. by having 2 raiden-cli clients having a Raiden channel between them and being online at the same time), terminates one of them, and see the other segfaults when closing the connection. Usually the terminated one too segfaults on `connection.close()` while shutting down.

## Expected Result
- webRTC channels can get reliably connected, closed, retried and reconnected during a session

## Actual Result
- `connection failed` errors happen and make webRTC unusable for the rest of that session's lifecycle.

## Remark
- After this has been fixed revert the ForceExit in the E2E nightly tests (see https://github.com/raiden-network/light-client/issues/2262)

andrevmatos:
Part of #1790 

**Short description**
So far, when an error failed with an error that should/could be handled inside the epic, the error logs on CLI were not very informative, and didn't contain the name of the function, because the epics functions were usually arrow functions. By using normal functions for the exported top level functions of the epics, and also including `epic.name` explicitly on the epics-specific error log, we get a better view of which epic failed.
This is just a stylistic change, and doesn't change anything in the epics logic.

An example of this error is the network error below, which made it explicit `monitorUdcBalanceEpic` was not properly handling failures there, and then a new commit fixed it.

While debugging it on CLI, it also was discovered that the force-exit after 10s was not getting triggered on self-shutdown of the SDK (only by the TERM/INT signal handlers), so this also got fixed. Hopefully, this won't be needed after we fix #2227 , so it's just a fallback right now.

```
2020-10-13T13:36:13.049Z [error] Epic error monitorUdcBalanceEpic [Function: monitorUdcBalanceEpic] Error: invalid response - 0
    at exports.XMLHttpRequest.request.onreadystatechange (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/ethers@4.0.48/node_modules/ethers/utils/web.js:84:29)
    at exports.XMLHttpRequest.dispatchEvent (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/ethers@4.0.48/node_modules/ethers/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:591:25)
    at setState (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/ethers@4.0.48/node_modules/ethers/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:610:14)
    at exports.XMLHttpRequest.handleError (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/ethers@4.0.48/node_modules/ethers/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:532:5)
    at ClientRequest.errorHandler (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/ethers@4.0.48/node_modules/ethers/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:459:14)
    at ClientRequest.emit (events.js:314:20)
    at Socket.socketErrorListener (_http_client.js:428:9)
    at Socket.emit (events.js:314:20)
    at emitErrorNT (internal/streams/destroy.js:92:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:60:3) {
  statusCode: 0,
  responseText: 'Error: getaddrinfo EAI_AGAIN parity.goerli.ethnodes.brainbot.com\n' +
    '    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:66:26)'
}
2020-10-13T13:36:13.061Z [info] %c action     color: #03A9F4; font-weight: bold {
  type: 'raiden/shutdown',
  payload: {
    reason: Error: invalid response - 0
        at exports.XMLHttpRequest.request.onreadystatechange (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/ethers@4.0.48/node_modules/ethers/utils/web.js:84:29)
        at exports.XMLHttpRequest.dispatchEvent (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/ethers@4.0.48/node_modules/ethers/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:591:25)
        at setState (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/ethers@4.0.48/node_modules/ethers/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:610:14)
        at exports.XMLHttpRequest.handleError (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/ethers@4.0.48/node_modules/ethers/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:532:5)
        at ClientRequest.errorHandler (/mnt/sec/Dev/brainbot/light-client/node_modules/.pnpm/ethers@4.0.48/node_modules/ethers/node_modules/xmlhttprequest/lib/XMLHttpRequest.js:459:14)
        at ClientRequest.emit (events.js:314:20)
        at Socket.socketErrorListener (_http_client.js:428:9)
        at Socket.emit (events.js:314:20)
        at emitErrorNT (internal/streams/destroy.js:92:8)
        at emitErrorAndCloseNT (internal/streams/destroy.js:60:3) {
      statusCode: 0,
      responseText: 'Error: getaddrinfo EAI_AGAIN parity.goerli.ethnodes.brainbot.com\n' +
        '    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:66:26)'
    }
  }
}
2020-10-13T13:36:13.063Z [info] Closing server...
```

**Definition of Done**

1. Try to trigger a network error on the CLI
2. Check the logged error is informative of which epic raised it
3. Check it's not `monitorUdcBalanceEpic`


andrevmatos:
Changelog isn't needed, since this shouldn't affect normal usage behavior, and is purely an internal change.

raiden-ts/src/services/epics.ts
```diff
@@ -442,15 +445,20 @@ export function pfsServiceRegistryMonitorEpic(
...
+  return action$.pipe(
+    filter(newBlock.is),
+    startWith(null),
     // it's seems ugly to call on each block, but UserDepositContract doesn't expose deposits as
     // events, and ethers actually do that to monitor token balances, so it's equivalent
-    exhaustMap(() => userDepositContract.functions.effectiveBalance(address) as Promise<UInt<32>>),
+    exhaustMap(() =>
+      defer(
+        async () => userDepositContract.functions.effectiveBalance(address) as Promise<UInt<32>>,
+      ).pipe(catchError(constant(EMPTY))),
```
weilbith, line 460: Does that mean we simply ignore errors? :scream:

weilbith (approved):
Could you eventually describe the fix for the UDC epic on a higher level? Not 100% sure I got the solution approach. Else the changes look good. I like to use "actual" functions and the CLI shutdown is also fine. :+1: 
I'm optimistically approving. It is just about knowledge sharing of the solution what is left for a complete review.

raiden-ts/src/services/epics.ts
andrevmatos: Yes, we do, since these errors should be just network-related (not something triggered by the user, or so), and it's a retriable error by nature, therefore we just omit any error here, and on next block it'll get requested again: notice that, if this was caused by connectivity loss (as I've witnessed), no new blocks notification will go through, ethers loop omit them and will re-emit when connectivity is restored, therefore these errors should happen only in case of a race (`newBlock` was emitted, but at the time `effectiveBalance` was requested, it got disconnected).
I'll write a comment to describe it better. Thanks

weilbith: Could you eventually do a small new PR and add some logging to it? I think it is fine to handle it like this. But for the (uncertain?) case that there could be a problem at a later point, it would be cool to at least see this in the logs while investigating on errors.

weilbith:
> It is just about knowledge sharing of the solution 

^-- @andrevmatos ?

raiden-ts/src/services/epics.ts
andrevmatos: This is a retried error, which is not user-triggered. I don't think logs are useful here, specially because it triggers only on race-conditions (current blockNumber call succeeded, function call failed). A more useful logging would be detecting we are missing blocks for some time, possibly even halting receiving transfers => #1256

PR #2260, opened 2020-10-15
Bugfix/circlci builds release in development mode

### LINKED ISSUE #2222: Fix the CI in GitHub release build to build the for the correct environment
## Description
Fix the CI in GitHub release build

The release should be built for the staging ghpages environemnt
It was working for the staging environment.

## Acceptance criteria
- 

## Tasks
- [ ]

weilbith:
Fixes #2222

**Short description**
Included in theses changes I reworked the triggers for the CI workflows. I wanted to be sure that I did not messed up anything that would then get triggered by opening a PR. Therefore I first pushed the changes to my fork only and let the CI run there. @palango took a look at the fork and thinks the changes are not breaking. Therefore I'm opening the actual PR now.

As always: please read the commits one by one and follow the argumentation/reasoning in the commit messages. Thanks! :pray: 

**Definition of Done**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,12 +1,5 @@
 version: 2.1
 
-templates:
-  filter-tag-templates: &filter-version-regex
-    tags:
-      only: /^v\d+\.\d+\.\d+$/
-    branches:
-      ignore: /.*/
-
 var_1: &docker_image raidennetwork/lightclient-node-pnpm:latest
 var_2: &cache_key raiden-{{ .Branch }}-{{ checksum "pnpm-lock.yaml" }}
 var_3: &working_dir ~/src
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,6 +231,11 @@ jobs:
             - ./raiden-dapp/dist/docs
 
   deploy_gh_pages:
+    parameters:
+      deploy_base:
+        description: The (sub-)directory where to put the build (empty for root)
+        type: string
+        default: ""
     executor: base-executor
     working_directory: *dapp_working_dir
     steps:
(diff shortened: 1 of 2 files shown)


weilbith:
I added multiple reviewers. Not because I don't trust a single one, but I want to share the knowledge of our CI pipe more. After all I want to have all "security" I can to make sure issues like the (hopefully) solved one do not occur again.

palango (approved):
As discussed privately I think this is a good change that makes stuff easier to understand and reason about:
- The three workflows make clearer what's supposed to happen when
- The `deploy_gh_pages` parameter makes the deployment target more explicit
- The `build_dapp` parameter makes the mode more explicit.

weilbith:
Note to myself: After this got merged I can remove the not used context on CircleCI.

weilbith:
I'll amend the latest commit and force-push. Apparently does the CI not get triggered since it run already on my fork. I think there was a configuration option at CircleCI for that. Anyways because of my paranoia I want to have CI run here as well. Because of contexts, configurations, ... :shrug:

.circleci/config.yml
```diff
@@ -303,52 +311,57 @@ workflows:
...
-  publish:
+  publish_release:
     jobs:
       - checkout:
           filters:
-            <<: *filter-version-regex
+            tags:
+              only: /^v\d+\.\d+\.\d+$/
+            branches:
+              ignore: /.*/
```
andrevmatos, line 344: You sure `branches` & `tags` conditions is ORed? We need to try a release to check if this works.

andrevmatos (approved):
lgtm, thanks

.circleci/config.yml
weilbith: Well this is the old filter I just copy-pasted. So I haven't changed the behavior in that detail. But after your comment I was reading the docs again in more detail. I fail to find an exact statement. Just the proposal that in future they should act as AND.

Anyway you pointed me to the fact that for some reason I did not remove the original anchor/template (even I thought I had).

weilbith:
Just as note: after mergin this to master the "new" `publish_staging` workflow was successfully triggered. Let's check the build result at the end. :blush:

PR #2264, opened 2020-10-15
Update to JSON based matrix server list

### LINKED ISSUE #1976: fetchSortedMatrixServers$ has a hand rolled parser for YAML
Just like the python client (https://github.com/raiden-network/raiden/issues/6424) the light client has a hand rolled parser:

https://github.com/raiden-network/light-client/blob/5b3bfb9b26f9f090f787d321b4b1214dfaa2dde7/raiden-ts/src/transport/epics/init.ts#L154-L164

The fix should be synchronized among both implementations.

palango:

Fixes #1976

**Short description**

Updates to the newer JSON_based matrix server list. This let's us remove our hand-rolled yaml parser.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/transport/epics/init.ts
+++ b/raiden-ts/src/transport/epics/init.ts
@@ -1,3 +1,4 @@
+import * as t from 'io-ts';
 import {
   Observable,
   combineLatest,
@@ -44,6 +45,7 @@ import { RaidenAction } from '../../actions';
 import { RaidenConfig } from '../../config';
 import { RaidenState } from '../../state';
 import { getServerName } from '../../utils/matrix';
+import { decode } from '../../utils/types';
 import { pluckDistinct, retryAsync$, retryWaitWhile } from '../../utils/rx';
 import { exponentialBackoff } from '../../transfers/epics/utils';
 import { matrixSetup } from '../actions';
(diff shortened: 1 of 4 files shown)


raiden-ts/src/transport/epics/init.ts
```diff
@@ -206,10 +189,10 @@ function fetchSortedMatrixServers$(matrixServerLookup: string, httpTimeout: numb
         response.ok,
         `Could not fetch server list from "${matrixServerLookup}" => ${response.status}`,
       );
-      return response.text();
+      return decode(MatrixServerInfo, response.json());
```
andrevmatos, line 192: `response.json()` returns a `Promise<any>` (to the json object), so you need to `return response.json();` (in order to get it merged to the stream and unrolled), and then, on a `map`, `decode` it.

andrevmatos: TS didn't complain because `decode` signature is very forgiving, since it's expected to receive anything and decode if possible or throw if it can't.

andrevmatos (approved):
There's an issue only, since it's simple, I'm optimistically approving. Otherwise, everything is perfect. Thanks.

raiden-ts/src/transport/epics/init.ts
andrevmatos: You can also `decode + mergeMap` the array on a single step (since decode is synchronous).

raiden-ts/src/transport/epics/init.ts
```diff
@@ -209,7 +192,8 @@ function fetchSortedMatrixServers$(matrixServerLookup: string, httpTimeout: numb
       return response.text();
     }),
     timeout(httpTimeout),
-    mergeMap((text) => yamlListToArray(text)),
+    map((text) => decode(MatrixServerInfo, JSON.parse(text), new RaidenError(ErrorCodes.TRNS_NO_MATRIX_SERVERS))),
```
palango, line 195: @andrevmatos Any hints on how to handle exceptions in `JSON.parse` here? And should I add a new error message for this?

andrevmatos: I'd suggest not calling `JSON.parse` directly. Instead, `return response.json()` above, and decode the resolved `json` object. I'd suggest also not passing an explicit error here, let the decode error flow (it can be more informative), and then catch it later, I think it's already handled on the function which calls `fetchSortedMatrixServers$`

andrevmatos: If you need to parse the text by yourself, use `jsonParse` utility

PR #2277, opened 2020-10-20
Fix mismatch between UDC totalDeposit and effectiveBalance

### LINKED ISSUE #2275: MetaMask - RPC Error: execution reverted: deposit not increasing
## Steps to Reproduce
1. Deposit tokens to the UDC 
udc_deposit_screen

2. Withdraw tokens from the UDC. They would get withdrawn to the raiden account. These tokens are kept as it is in the raiden account.

3. Again deposit to the UDC. The UDC has 0 SVT token balance(because we just withdrew in the previous step).
 
execution_reverted_deposit_not_increasing

## Expected Result
- Ideally the user should be able to deposit and withdraw tokens as he wishes.

## Actual Result
 ```
inpage.js:1 MetaMask - RPC Error: execution reverted: deposit not increasing {code: -32603, message: "execution reverted: deposit not increasing", data: {…}}
```

```
- Browser & version
- Browser console errors
- Device / OS
```

### PR DESCRIPTION
**Pull request by andrevmatos**

Fixes #2275

**Short description**
`udcDeposit` actions contained `totalDeposit` as sole `meta` property, but it actually was the `effectiveBalance` value, which raised issues when calculating the new totalDeposit when making a deposit after a withdraw was performed.
This PR fixes it by using actual `totalDeposit` as meta value (for linking `udcDeposit` requests & success/failure responses), and added `balance` to payload, to be used as UDC balance, instead of the deposit.
Integration tests were extended to assert behavior when depositing after withdraw, which works as regression test for this action.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/services/epics.ts
+++ b/raiden-ts/src/services/epics.ts
@@ -459,13 +459,16 @@ export function monitorUdcBalanceEpic(
        * etc), but merged on the top-level observable, therefore connectivity issues can cause
        * exceptions which would shutdown the SDK. Let's swallow the error here, since this will be
        * retried on next block, which should only be emitted after connectivity is reestablished */
-      defer(
-        async () => userDepositContract.functions.effectiveBalance(address) as Promise<UInt<32>>,
+      defer(async () =>
+        Promise.all([
+          userDepositContract.functions.effectiveBalance(address) as Promise<UInt<32>>,
+          userDepositContract.functions.total_deposit(address) as Promise<UInt<32>>,
+        ]),
       ).pipe(catchError(constant(EMPTY))),
     ),
     withLatestFrom(latest$),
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: raiden-ts/src/services/epics.ts [hunk 1]
```diff
@@ -459,13 +459,16 @@ export function monitorUdcBalanceEpic(
        * etc), but merged on the top-level observable, therefore connectivity issues can cause
        * exceptions which would shutdown the SDK. Let's swallow the error here, since this will be
        * retried on next block, which should only be emitted after connectivity is reestablished */
-      defer(
```
palango, line 462: Regarding line 455: We should open an issue for the contracts to add events.

palango (approved):
Looks good.

PR #2282, opened 2020-10-21
dApp: Fixes issues where disclaimer content appears behind button on mobile

### LINKED ISSUE #2276: Not all elements reachable on mobile Firefox
When using https://lightclient.raiden.network with Android Firefox 81.1.2, I can't reach all elements of the UI (e.g. the checkboxes and "Accept" button on the first page, see screenshot).

Screenshot_20201020-152612

Selecting the "Desktop site" option in Firefox allows me to reach all UI elements.

taleldayekh:
Fixes #2276 

**Short description**
Apparently, on mobile, the content was still displayed behind the `Accept` button on the disclaimer screen. This was not the case when viewing the dApp in the dev tools but it was the case when opening the dApp on an actual mobile. I managed to reproduce the problems by viewing the dApp in the Safari dev tools. These fixes make the disclaimer screen look good in the Safari dev tools so hopefully, that will also be the case when viewing it on an actual mobile device.

**Definition of Done**

1. View disclaimer on actual mobile, not sure how you would go about doing this without merging this PR to master...

### DIFF
--- a/raiden-dapp/src/views/DisclaimerRoute.vue
+++ b/raiden-dapp/src/views/DisclaimerRoute.vue
@@ -1,7 +1,7 @@
 <template>
-  <v-container fluid>
-    <div class="disclaimer">
-      <div class="disclaimer__paragraphs font-weight-light">
+  <div class="disclaimer">
+    <div class="disclaimer__content">
+      <div class="disclaimer__content__paragraphs font-weight-light">
         <p
           v-for="(paragraph, index) in $t('disclaimer.paragraphs')"
           :key="index"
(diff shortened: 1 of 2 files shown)


weilbith (commented):
So what is the idea of the fix in detail? I fail to identify it in the changes.
Did you consider to use `flex-grow` for the paragraphs div?
Btw you should be able to serve the dApp, get the IP address of your PC in the local network (don't ask a website or so, since they report your external IP) and access the dApp from your mobile phone (maybe ask Karl to test as well). Note that your phone must be in the same local network. Thereby we can be sure it actually works.

taleldayekh:
Doesn't `flex-grow` and `flex` do the same thing? Currently, the div wrapping all paragraphs is set to `flex: 1` which expands it to all available space.

weilbith:
> Doesn't `flex-grow` and `flex` do the same thing? Currently, the div wrapping all paragraphs is set to `flex: 1` which expands it to all available space.

Sorry, my mistake. I dislike the pure `flex` property because I always don't understand what it sets and what is automated. But that is my own incompetence.

Could you still try to do an actual test?

taleldayekh:
Yes, looking at how to test it the way you suggested now. :)

weilbith:
Great. Thanks for doing that. I think it is the best way to be sure it is actually working. :+1:

taleldayekh:
I tested it using the my internal IP, thanks for the tips! It looks great now on Chrome, Firefox and Safari mobile.

weilbith (approved):

PR #2291, opened 2020-10-26
Update e2e environment

palango:
**Thank you for submitting this pull request :)**

Fixes #2223 

**Short description**

Update the e2e environment to the latest version of dependencies.

**Definition of Done**

- [ ] Steps to manually test the change have been documented
- [x] Acceptance criteria are met
- [ ] Change has been manually tested by the reviewer (dApp) 

**Steps to manually test the change (dApp)**

1.
2.

PR #2292, opened 2020-10-26
Feature/ci improvements and fixes

weilbith:
Fixes _everything_. :stuck_out_tongue_winking_eye:  

**Short description**

Our CircleCI configuration has become quite big. I guess that is the case for any mono-repo at some point in time. Though we experienced some issues recently, getting confronted with a little messy setup of our workflows and jobs.
While fixing our recent issue I read a lot of CircleCI documentation and learned a couple of new features they support. I think they can help to make our configuration more clean again and help to avoid future issues. Although this is not the end of the road (list of ideas follows), this should be enough to get its own PR. Future improvements I eventually continue to do voluntary in my free time. So no worries I would waste time.
I tested all of my changes for each step over and over again with a CircleCI pipeline setup for my fork of the project. Thereby I'm for example that the workflows are still triggered correctly and the fixes solve what they promise. As always should each commit message include enough information to understand all changes and their reasoning.

Thinks I would like to fix:

There is the feature of layered Docker images for an executor in CircleCI. This allows to use a base image where all jobs get executed inside and all others just offer services to the first one. This pretty much exactly what we need. The base image provides everything to get all projects installed, build, tested, ... The second image provides all services for a full features end-to-end test environment. At the moment this image must also make sure the projects can be build and run inside. And even worse: the jobs need to be parametrized to differ the used executor. This leads to some overhead in the configuration of CircleCI, the end-to-end image and also ugly (and not meaningful) auto-generated job names (thanks CircleCI). Unfortunately it does not work for the moment for multiple reasons. First do the SDK e2e tests need the environment file of the deployment infos, which residence in the e2e image. But eventually CircleCI injects the Docker socket into the base layer image and allows us to copy it from there inside a job. Anyways the second layer image always stops for some reason. According to CircleCI there is a non-zero exit code. From the logs the container works fine (on a first glance) and no errors are reported.

I still think that separating the logic into different workflows makes sense. I prefer this solution over a single workflow that has a lot of if-else-filter-... It makes it super hard to get what happens when under which condition. Workflows now have a single condition/trigger that makes it super easy to read. Unfortunately does this lead to a lot of copy-pasted code, since all of these workflows have a lot in common. I would love to make this re-usable and more pleasant to read/understand. YAML anchors are not solution. They would eventually make it worse and after all they could not handle the job parameters. It would be great to export a few of the current jobs as commands and then have jobs which orchestrate multiple of them with shared parameters. But this would remove the ability to parallelize many of these jobs (then commands). There are options to split a jobs its steps, but this appears quite complex and I'm not even sure it works for this case (meant to split tests over multiple machines to speed them up). So I don't know how to continue here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,13 +79,16 @@ jobs:
           command: npm install license-checker@25.0.1
       - run:
           name: Check SDK Licenses
-          command: cd ~/src/raiden-ts && license-checker --production --onlyAllow 'MIT;BSD;Apache;MPL;LGPL;ZPL;ISC;WTFPL;Unlicense' --excludePackages 'uuid@2.0.1' --excludePrivatePackages # uuid@2.0.1 has an issue with the license information but it is licensed under MIT
+          working_directory: raiden-ts
+          command: license-checker --production --onlyAllow 'MIT;BSD;Apache;MPL;LGPL;ZPL;ISC;WTFPL;Unlicense' --excludePackages 'uuid@2.0.1' --excludePrivatePackages # uuid@2.0.1 has an issue with the license information but it is licensed under MIT
       - run:
           name: Check dApp Licenses
-          command: cd ~/src/raiden-dapp && license-checker --production --onlyAllow 'MIT;BSD;Apache;MPL;LGPL;ZPL;ISC;WTFPL;Unlicense' --excludePackages 'uuid@2.0.1' --excludePrivatePackages # uuid@2.0.1 has an issue with the license information but it is licensed under MIT
+          working_directory: raiden-dapp
+          command: license-checker --production --onlyAllow 'MIT;BSD;Apache;MPL;LGPL;ZPL;ISC;WTFPL;Unlicense' --excludePackages 'uuid@2.0.1' --excludePrivatePackages # uuid@2.0.1 has an issue with the license information but it is licensed under MIT
       - run:
           name: Check CLI Licenses
-          command: cd ~/src/raiden-cli && license-checker --production --onlyAllow 'MIT;BSD;Apache;MPL;LGPL;ZPL;ISC;WTFPL;Unlicense' --excludePackages 'uuid@2.0.1' --excludePrivatePackages # uuid@2.0.1 has an issue with the license information but it is licensed under MIT
+          working_directory: raiden-cli
+          command: license-checker --production --onlyAllow 'MIT;BSD;Apache;MPL;LGPL;ZPL;ISC;WTFPL;Unlicense' --excludePackages 'uuid@2.0.1' --excludePrivatePackages # uuid@2.0.1 has an issue with the license information but it is licensed under MIT
 
   lint:
     executor: base-executor
@@ -190,7 +193,8 @@ jobs:
           command: yarn workspace raiden-dapp build --mode << parameters.mode >>
       - run:
           name: Compress dapp
-          command: cd raiden-dapp && tar -czvf dapp.tar.gz dist
+          working_directory: raiden-dapp
+          command: tar -czvf dapp.tar.gz dist
       - store_artifacts:
           path: raiden-dapp/dapp.tar.gz
           destination: dapp.tar.gz
@@ -209,8 +213,8 @@ jobs:
           command: yarn workspace raiden-cli run build && yarn workspace raiden-cli run build:bundle
       - run:
           name: Compress CLI bundle
+          working_directory: raiden-cli
           command: >
-            cd raiden-cli ;
             cp -av ./raiden bundle/ ;
             tar -czvf raiden-cli.tar.gz bundle --transform='s/^bundle/raiden-cli/'
       - store_artifacts:
@@ -222,7 +226,9 @@ jobs:
     working_directory: << pipeline.parameters.working_directory >>
     steps:
       - attach_workspace: *attach_options
-      - run: cd docs && ./generate_documentation.sh
+      - run:
+          working_directory: docs
+          command: ./generate_documentation.sh
       - persist_to_workspace:
           root: << pipeline.parameters.working_directory >>
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ executors:
   base-executor:
     <<: *executor_shared_options
     docker:
-      - image: circleci/node:<< pipeline.parameters.base_executor_image_tag >>
+      - image: cimg/node:<< pipeline.parameters.base_executor_image_tag >>
 
   e2e-environment-executor:
     <<: *executor_shared_options
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,10 +13,6 @@ parameters:
     type: string
     default: ~/src
 
-  yarn_cache_key:
-    type: string
-    default: raiden-{{ .Branch }}-{{ checksum "yarn-lock.yaml" }}
-
   test_report_directory:
     type: string
     default: reports/junit
@@ -97,10 +93,14 @@ jobs:
       - run: git submodule sync
       - run: git submodule update --init
       - restore_cache:
-          key: << pipeline.parameters.yarn_cache_key >>
+          name: Restore Yarn Package Cache
+          keys:
+            - raiden-v1-yarn-{{ checksum "yarn.lock" }}
+            - raiden-v1-yarn-
       - run: yarn install
       - save_cache:
-          key: << pipeline.parameters.yarn_cache_key >>
+          name: Save Yarn Package Cache
+          key: raiden-v1-yarn-{{ checksum "yarn.lock" }}
           paths:
             - ./node_modules
             - ~/.cache # Cypress stores its stuff here.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,6 +44,21 @@ executors:
     docker:
       - image: raidennetwork/lightclient-e2e-environment:<< pipeline.parameters.e2e_environment_executor_image_tag >>
 
+commands:
+  check-license:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -79,14 +89,14 @@ jobs:
 
   lint:
     executor: base-executor
-    working_directory: *working_dir
+    working_directory: << pipeline.parameters.working_directory >>
     steps:
       - attach_workspace: *attach_options
-      - run: pnpm run lint -r
+      - run: pnpm run lint --recursive
```
palango: Nice, I like using to long forms in scripts!

.circleci/config.yml
```diff
@@ -79,13 +79,16 @@ jobs:
           command: npm install license-checker@25.0.1
       - run:
           name: Check SDK Licenses
-          command: cd ~/src/raiden-ts && license-checker --production --onlyAllow 'MIT;BSD;Apache;MPL;LGPL;ZPL;ISC;WTFPL;Unlicense' --excludePackages 'uuid@2.0.1' --excludePrivatePackages # uuid@2.0.1 has an issue with the license information but it is licensed under MIT
+          working_directory: raiden-ts
```
palango: How does this work? Above we define `working_directory: ~/`, so shouldn't this be `src/raiden-ts`?

.circleci/config.yml
```diff
@@ -119,41 +119,40 @@ jobs:
...
     steps:
       - attach_workspace: *attach_options
       - run:
           name: Build dapp
-          command: pnpm run build --filter raiden-dapp -- --mode << parameters.mode >>
+          command: pnpm run build -- --mode << parameters.mode >>
       - run:
           name: Compress dapp
-          working_directory: raiden-dapp
           command: tar -czvf dapp.tar.gz dist
```
palango, line 158: How can it be that we don't have to adjust the command when we change the working dir?

.circleci/config.yml
```diff
@@ -59,6 +60,26 @@ commands:
...
+      - store_test_results:
+          path: reports/junit
+      - store_artifacts:
+          path: reports/junit
+      - run:
+          name: "Upload test coverage results to CodeCov"
+          command: >
+            bash <(curl -s https://codecov.io/bash) -C $CIRCLE_SHA1
+            -F << parameters.base_flag >>
+            -F << parameters.base_flag >>.<< parameters.test_kind_flag >>
```
palango, line 88: What do the double flags do? Make it so that I can filter for all sdk tests?

.circleci/config.yml
```diff
@@ -363,29 +360,38 @@ workflows:
             - build_dapp
 
   publish_production:
+    when:
+      and:
+        - not: << pipeline.git.branch >>
```
palango, line 364: Branch is not set for tags? Or when would we require both conditions?

palango (commented):
Great PR that cleans up a lot of stuff. Left some questions, but looks good from my side. 🏅

.circleci/config.yml
weilbith: This is outdated. But you are right, actually this should not have worked on this commit. :thinking:

.circleci/config.yml
weilbith: I was changed (checkout [this version](https://github.com/raiden-network/light-client/blob/447a642baadd42b9503288f707d6cf584971a1bd/.circleci/config.yml#L194). GitHub is just pretty bad in showing a good diff here since the order of the job definition has changed.

.circleci/config.yml
weilbith: First this feature was not added, but was there before. Seconds yes, this makes it easier to identify coverage changes by "location". See for example the table in [this report](https://github.com/raiden-network/light-client/pull/2291#issuecomment-716520206).

.circleci/config.yml
weilbith: Yes. So these both variables are exclusive. The pipeline gets either triggered by a tag or a branch update/creation (there are more triggers). So either of both is empty. So we could get away with just one of both expressions here. This is just more explicit and since the recent bad experience with tags and CircleCI I wanted to be sure this is as secure as it can be. :joy: 
Note that we can't use regex for the tag here. But since each job needs a tag filter anyways to run for a tag triggered pipeline (you remember?) this would be useless anyways. So this `when` just states that this job should run when the pipeline got triggered by a tag. The tag filtering is then done at the job list of the workflow.

palango (approved):

weilbith:
In PR #2289 we adopted the project from `pnpm` to `yarn`. When the PR got finalized, I'll merge it's changes into this PR. Eventually I could also rebase here. Not sure which is the best solution, since this will affect every single commit in this branch here. :see_no_evil:

weilbith:
I've rebased on `master` and resolved the mass of conflicts on each level. Le'ts see if the base workflow still runs through.

.circleci/config.yml
```diff
@@ -1,46 +1,111 @@
...
+    type: string
+    default: latest
+
+  working_directory:
+    type: string
+    default: ~/src
+
+  yarn_cache_key:
+    type: string
+    default: raiden-{{ .Branch }}-{{ checksum "yarn-lock.yaml" }}
```
palango, line 18: As discussed here the `Branch` doesn't make much sense: https://github.com/raiden-network/light-client/pull/2312#discussion_r516599517

weilbith: Good catch. Sorry that I missed that while resolving the conflicts. :see_no_evil: 

It is a very good feeling that the reviewer is really competent and careful. Thanks for being such a colleague! :pray:

weilbith: While looking again on the master version: why do we have two keys here, while [the second](https://github.com/raiden-network/light-client/blob/5a729c9e3b1ce6fd1ff1ad71771a66811951d0b5/.circleci/config.yml#L77) looks pretty awkward. Especially because we don't use this key a few lines below. Do you have any insights?

palango: I assumed this was kind of a fallback, but now that you mention it I'm not so sure any more.

weilbith: I asked Andre as he is the author and I got [this helpful link](https://circleci.com/docs/2.0/caching/#restoring-cache) to the documentation. So I'll recover that as well. :+1:

weilbith: :heavy_check_mark:

andrevmatos:
Can we merge the `build_sdk` step and do it in the `install` phase? I think we can shave 2min from CI with it.

.circleci/config.yml
```diff
@@ -1,76 +1,97 @@
...
-    branches:
-      ignore: /.*/
 
 executors:
   base-executor:
-    working_directory: *working_dir
+    <<: *executor_shared_options
     docker:
-      - image: circleci/node:lts-browsers
+      - image: cimg/node:<< pipeline.parameters.base_executor_image_tag >>
```
andrevmatos, line 43: Nitpick: I don't think putting only the tag on a parameter is very useful, it's more probable to change the whole image than to switch tags. Wouldn't be worth it to put the whole image in the parameter instead?

weilbith: Sure we can do that. :+1:

weilbith:
> Can we ditch the build_sdk step and do it in the install phase? I think we can shave 2min from CI with it.

Hmm. :thinking: 
I'm not too sure about sure. Where to we do we draw the border? Put everything into a single job that always runs together in all workflows? I think jobs should be like functions in code. Keep them small, act on the same level of abstraction etc pp.
Merging checkout and install made sense, because both are in fact the overall preparation for all other jobs. The build of the SDK is an actual standalone part. It is not only necessary for further jobs to proceed, but is a test/validation on itself: does the SDK build successfully in a standardized environment (e.g. did the developer forgot a dependency he installed locally. Although does the separation of jobs provide an easy analysis flow. Already at the mail report you see which part failed. The logs of the job are smaller and easier to follow. The parameters of the job are easier to maintain since they are focused to a single thing. It reduces complexity to say so.

.circleci/config.yml
weilbith: :heavy_check_mark:

.circleci/config.yml
```diff
@@ -84,95 +105,38 @@ jobs:
...
-          name: "Upload coverage"
-          command: bash <(curl -s https://codecov.io/bash) -F dapp -C $CIRCLE_SHA1
-
-  test_dapp_e2e:
-    executor: e2e-environment-executor
-    working_directory: *working_dir
-    steps:
-      - attach_workspace: *attach_options
+          name: Update $PATH
+          command: echo 'export PATH=$PATH:$HOME/node_modules/.bin' >> $BASH_ENV
```
andrevmatos, line 120: Meddling with `PATH` isn't pretty.. since we have `yarn` available, maybe it'd be worth it to use `yarn exec`? I think maybe it doesn't even require installing explicitly, and do it automatically

weilbith: Let me try. This might be a cool change. At least my comment above becomes regardless since we don't have a custom Docker image anymore.

weilbith: Btw: in the package.json we have a `npx` command to allow only yarn. So `npx` is apparently available. Why not simply using it here as well?

andrevmatos: Yes, you can as well. npx should be available wherever npm is, so it's safe to assume it'd be there, and it also can auto-download libs as needed

weilbith: I tested it locally with their Docker image and it looks fine. So I'll change that in the config and we should be fine.

weilbith: :heavy_check_mark:

andrevmatos (approved):
lgtm

PR #2298, opened 2020-10-29
dApp: State backup notification

### LINKED ISSUE #2178: I would like to get reminded frequently to make a backup of my state
## User Story
As Chris I would like to get reminded frequently to make a backup of my state so that I don't lose tokens.

## Acceptance criteria
- It is shown as a notification
- Show it once a day

taleldayekh:
Fixes #2178 

**Short description**
Reminds user to backup state when connecting and sends new reminder each 24h upon connecting.

**Definition of Done**

1.
2.


taleldayekh:
I still need to add the tests.

raiden-dapp/src/components/notification-panel/NotificationCard.vue
```diff
@@ -74,6 +82,16 @@ export default class NotificationCard extends Vue {
...
+
+  get linkRoute() {
+    switch (this.notification.dappRoute) {
+      case 'backup state route':
+        return () => this.navigateToBackupState();
+        break;
+      default:
+        break;
+    }
+  }
```
weilbith, line 94: Can I suggest you a better solution? 
I really like the Open-Closed-Principle. The idea is that the notification card is able to handle optional links of a notification. But you don't want to adjust the component each time a new link got added. It is cumbersome to have the developer remind this and you simply clutter the logic of the notification card component.
I would rather replace the `notification.link` with a `RouteName` (use the typing). And then just have a function that uses `this.$router.push(name: this.notification.link)` (or similar). Thereby you have written the logic once. As long as the idea of the functionality remains the same this remains "closed". But it is open for extension, because each notification can simply define this link name and it works out of the box. When you add a new notification or extend an existing one with a link you don't have to remember adding a switch-case here. You can get rid of the Mixin (which bloats up the final package/user load time) and due to the enumeration of the route names you are also type safe (note to then not define the links in the locals, but where you dispatch the configuration).
Sorry for the long comment and the critics. I hope you understand my arguments.

raiden-dapp/src/store/index.ts
```diff
@@ -54,6 +54,7 @@ const _defaultState: RootState = {
   config: {},
   userDepositTokenAddress: '',
   disclaimerAccepted: false,
+  latestStateBackupReminder: 0,
```
weilbith, line 57: Background: I remember that there was a "discussion" what representation/type to use here.
Unfortunately the name of the variable does not give enough information to explain itself. Is this the date in milliseconds when the last reminder was shown or..? Please check if you could help a stupid reader like me with a better name if possible. :wink: 

PS: I think the signature of this mutation (`stateBackupReminder(state: RootState, newReminderDate: number)`) shows perfectly the mismatch. Btw I think active names for mutations are preferrable. Also we did not in the past we should attempt to improve. Something like `update...`, `set...` or something else with a verb. `stateBackupReminder` what?!

raiden-dapp/src/views/TransferRoute.vue
```diff
@@ -40,15 +43,47 @@ import { Token, TokenModel } from '@/model/types';
...
   computed: {
+    ...mapState(['latestStateBackupReminder']),
     ...mapGetters(['tokens', 'channelWithBiggestCapacity', 'openChannels']),
   },
 })
 export default class TransferRoute extends Vue {
+  latestStateBackupReminder!: number;
   tokens!: TokenModel[];
+  currentTime: number = new Date().getTime();
+  oneDay: number = new Date().setHours(24);
```
weilbith, line 54: This is not part of the components state. It is a fixed constant. Please define it outside of the class and consider to name it in upper snake case like `ONE_DAY`.

raiden-dapp/src/views/TransferRoute.vue
```diff
@@ -40,15 +43,47 @@ import { Token, TokenModel } from '@/model/types';
...
   },
   computed: {
+    ...mapState(['latestStateBackupReminder']),
     ...mapGetters(['tokens', 'channelWithBiggestCapacity', 'openChannels']),
   },
 })
 export default class TransferRoute extends Vue {
+  latestStateBackupReminder!: number;
   tokens!: TokenModel[];
+  currentTime: number = new Date().getTime();
```
weilbith, line 53: Question: When do you want to change this value? Why not getting the time when it is needed instead of when the class gets created? This could eventually be even wrong. In this case it probably would work. But it is an error prone pattern that you should not get used to. :wink: Just imagine it would be a `v-if` component similar. Just do it in a single line in `mounted()`.

raiden-dapp/src/views/TransferRoute.vue
```diff
@@ -40,15 +43,47 @@ import { Token, TokenModel } from '@/model/types';
...
 
+  mounted() {
+    if (this.latestStateBackupReminder === 0) {
+      this.pushStateBackupNotification();
+    } else if (
+      this.currentTime >
+      this.latestStateBackupReminder + this.oneDay
+    ) {
+      this.pushStateBackupNotification();
+    }
```
weilbith, line 76: Both condition execution blocks do 100% the same. So please merge the both boolean expression with an OR. If you like to you can store each expression into a boolean variable with a expressive name. In the end it could then read if `if (neverNotified || lastNotificationOlderThanOneDay) {...}`.

raiden-dapp/src/views/TransferRoute.vue
```diff
@@ -40,15 +43,47 @@ import { Token, TokenModel } from '@/model/types';
...
+    ...mapState(['latestStateBackupReminder']),
     ...mapGetters(['tokens', 'channelWithBiggestCapacity', 'openChannels']),
   },
 })
 export default class TransferRoute extends Vue {
+  latestStateBackupReminder!: number;
   tokens!: TokenModel[];
+  currentTime: number = new Date().getTime();
+  oneDay: number = new Date().setHours(24);
+  stateBackupReminder = {
```
weilbith, line 55: The same as before. Either you say this is something fixed and do it outside, or it is always new for each notification and create in inside of the function. But it will never change. So it is not a state.

PS: Once we have resolved the messy RaidenService module we should have proper notification helper that reduce the code here and keep if more focused (in fact this component should not care about how the notification looks in detail, just send it).

weilbith (changes requested):
As I understood you are about to still add the tests? :smiley: 
All in all just minor comments. I'm so happy that you found a pretty way to introduce links into the notification panels. :partying_face:

raiden-dapp/src/assets/notification_state_backup.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 25 25"><defs><style>.cls-1{fill:#7f878d;}</style></defs><g id="state_backup" data-name="state backup"><path id="pointers" class="cls-1" d="M15.71,15.85,12.81,13V7.63a.69.69,0,1,0-1.38,0V13.2a.65.65,0,0,0,.2.52l3.1,3.1a.67.67,0,0,0,.49.21.7.7,0,0,0,.49-1.18Z"/><g id="circle"><path class="cls-1" d="M12.57,21.5a8.94,8.94,0,0,1-7.18-3.57.69.69,0,1,1,1.1-.84A7.62,7.62,0,1,0,12.57,4.88,7.67,7.67,0,0,0,5.44,9.81a.69.69,0,0,1-1.29-.49A9,9,0,1,1,12.57,21.5Z"/><polygon class="cls-1" points="4.88 14.92 8.07 9.38 1.68 9.38 4.88 14.92"/></g></g></svg>
```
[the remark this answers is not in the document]
taleldayekh: I agree, I'll create a notifications dir for the notifications icons. 👍

raiden-dapp/src/components/notification-panel/NotificationCard.vue
taleldayekh: This is much better, thank you for the suggestions.

raiden-dapp/src/components/notification-panel/NotificationCard.vue
```diff
@@ -29,7 +29,7 @@
         <span
           v-if="notification.link"
           class="notification-card__content__details__link"
-          @click="linkRoute"
+          @click="linkRoute()"
```
weilbith, line 32: Don't! Parenthesis mean to execute the function **now**. Why did you added them now? :confused:

andrevmatos: You may want to use the previou style as a consistency decision, but these parenthesis are only evaluated at event-firing time, so both syntax are almost the same, specially if the callback doesn't receive parameters, the only difference is that the previous call would receive a `$event` param, while explicitly stating the parenthesis will call it without arguments (which may or may not be what you expect, just be aware of it).

raiden-dapp/src/views/TransferRoute.vue
taleldayekh: Genau, good point. 👍

raiden-dapp/src/components/notification-panel/NotificationCard.vue
taleldayekh: Oh, my bad.

weilbith: Oh sorry. My bad. But yeah, maybe it works here, but "everywhere" else not. Sorry for the trouble :see_no_evil:

weilbith (approved):

PR #2300, opened 2020-10-29
Adding custom cypress attributes so that classes can keep changing

### LINKED ISSUE #2285: Add custom Cypress attributes to selectors
## Description
Currently, we are accessing all elements via our class selectors. This has the consequences that if we change a class name in our code it will break the Cypress e2e tests and there is no easy way to catch this early on.

We should add custom attributes to all elements which are used in Cypress. See Cypress and Vuetify docs for more details:
https://docs.cypress.io/guides/references/best-practices.html
https://vuetifyjs.com/en/getting-started/unit-testing/#mocking-vuetify

## Acceptance criteria
- Replace all class selectors with the custom attributes prefixed with `data-e2e-selector`.

## Tasks
- [ ] Early draft PR review to support @agatsoh

agatsoh:

Fixes #2285 

**Short description**

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-dapp/CHANGELOG.md
+++ b/raiden-dapp/CHANGELOG.md
@@ -2,7 +2,18 @@
 
 ## [Unreleased]
 
+### Fixed
+
+### Added
+
+- [#2285] Add custom Cypress attributes to selectors
+
+### Changed
+
+[#2285]: https://github.com/raiden-network/light-client/issues/2285
+
 ## [0.12.0] - 2020-10-22
+
 ### Fixed
 
 - [#2237] Displays RDN token on Withdrawal screen
@@ -40,6 +51,7 @@
 [#2224]: https://github.com/raiden-network/light-client/issues/2223
 
 ## [0.11.1] - 2020-08-18
+
 ### Fixed
 
 - [#2047] Conversion and token amount display in UDC deposit dialog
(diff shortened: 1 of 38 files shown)


taleldayekh (approved):
Great job, the naming is consistent and all looks good thanks for taking care of this 👍 

Just two questions. Did you figure out why `({ force: true })` is needed now when it was not needed before? And, I tried to run the tests locally but get this error, did you encounter this anything? I guess it is related to something locally for me? If they work for you and CI passes then all should be good.

Screenshot 2020-10-30 at 16 12 36

agatsoh:
@taleldayekh  I still havent figured out why we still have to use `{force: true}` , its mostly needed on the Disclaimer screen, and as far as I know there is absolutely no element covering anything. There might be one possibility that the disclaimer is quite large the the checkboxes that we try to click and the final do not come in the view port of the user.  Its was difficult for me to have a look because the test runs really fast and you dont even get the opportunity to see properly the components in the test.
I will try to analyse this sometime in future and try scrolling down and bit and then giving the click actions and see whether it solves our problems. Since its only in one screen and not the entire it should not be much of a problem but definitely if possible to to avoid and fix this.
just as a reference i will post the links to the docs where they mention this issue along with some allied issues.
https://docs.cypress.io/guides/references/error-messages.html

PR #2310, opened 2020-11-02
Unify eslint and prettier configs on monorepo

### LINKED ISSUE #1864: Introduce unified rules for the whole project
## Description

Now that we have pnpm installed we need to introduce unified TypeScript eslint rules for the whole monorepo so that the projects are in sync.

## Acceptance criteria
- 

## Tasks
- [ ]

andrevmatos:
Fixes #1864 

**Short description**
- Move common config to root's `.eslintrc.json` and `.prettierrc.json`.
- Remove the options now in the root dir from specific workspaces
- On workspaces, keep only specific configs (e.g. Vue rules in dApp)
- Fix lint issues on each package

Since this is a development/style only change, it doesn't need its changelog entry.

**Definition of Done**

1.
2.


raiden-ts/package.json
```diff
@@ -128,75 +120,13 @@
...
-      "jsdoc/require-hyphen-before-param-description": "warn",
-      "jsdoc/require-jsdoc": [
-        1,
-        {
-          "publicOnly": true
-        }
-      ]
-    }
-  },
   "eslintIgnore": [
```
weilbith, line 123: As we have discussed already, I think it is cleaner to put this list into a ignore file.

weilbith (commented):
This is so great. I love it. :heart_eyes: 
I'll finalize the work for the dApp. :+1:

raiden-cli/package.json
```diff
@@ -27,14 +27,7 @@
     "@types/node": "^14.14.6",
```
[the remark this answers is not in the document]
andrevmatos: I'd say it's better not to:
I'd like the packages to be as standalone as possible, specially if they're published as is and someone try to install them from a registry like `npmjs`, they should be usable standalone. Although `@types/node`  is only required at development time, its depended types may still be required to be available even when installed from registry and used in a TS/ESM project, therefore they should stay on the workspaces. 
Meanwhile, linting rules already (now) depend on the workspace tree (since most of the rules are now in the root), and therefore, the specific devDependencies required for it may go to the root, but I'd say we should keep it at a minimum and let yarn hoisting do its job if/when needed.

raiden-ts/package.json
andrevmatos: Sure. May you do it, please, so we don't conflict? Thanks

raiden-cli/package.json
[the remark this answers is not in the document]
weilbith: Okay, so only dev-dependencies should be shared via root?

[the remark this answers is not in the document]
andrevmatos: Yes, only devDependencies and which depends on something explicitly requiring the workspace tree (CI, linting, tools, etc). devDependencies used only on package development or dependant's development should stay in the workspace dependencies, as they are now.

raiden-ts/package.json
weilbith: :heavy_check_mark:

raiden-cli/package.json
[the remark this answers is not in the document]
andrevmatos: > For dev-deps I don't get the registry argument.

There are some devDependencies which are there because they are not required at runtime (imported, indirectly run, etc), but which are required if you are developing a project which depends on this package, and have installed it from the registry: e.g. some types; I want to have any devDependency which match this criteria to stay in the workspace's package.json; actually, I want most devDependencies to stay there, and only move up things which already require the workspace tree explicitly, like things which depend on shared tools/configs/etc.

raiden-ts/.eslinignore
```diff
@@ -0,0 +1,5 @@
+raiden-contracts/*
```
andrevmatos, line 1: @weilbith `.eslinignore` -> `.eslintignore`

raiden-dapp/tests/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "baseUrl": "..",
```
weilbith, line 4: Why? :thinking:

raiden-dapp/tests/unit/components/amount-input.spec.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
weilbith, line 1: Keep it like that. But I think ignoring a whole file is just a last resort. If the linting tool supports it (what it does here) I would prefer to ignore specific lines. Thereby we can explicitly know that the developer ignored this line on purpose. This suppresses all future warnings that could occur without getting noticed. In general this is not too much of an issue. And when the whole files is full of this pattern this also makes sense. So I think for many files you did the best decision.

This comment is meant to share different views/opinions. Please close it after reading or feel free to update the conversation with further arguments form your side.

weilbith (approved):
Just reviewed the last commit I haven't seen so far. Looks pretty good to I think. This became a quite huge PR. But it must be like that. :shrug: :see_no_evil:

raiden-dapp/tests/tsconfig.json
andrevmatos: So it bases on the root of the package (the same as the upper tsconfig). Required in order to properly reference on the `paths` alias. Maybe it'd work with the full relative path there as well, but this is more common, and also how it's done in the SDK.

raiden-dapp/tests/unit/components/amount-input.spec.ts
andrevmatos: I agree with you, but mostly on src files. For tests, my policy was to put it file-wide if it was more than 2 or 3 instances. Either way, using `any` here is a workaround, hopefully soon enough we'll have better tooling and will be able to remove most of these. Thanks for the headsup!

kelsos:
Awesome, glad to get notified that this was completed. Keep up the good work :+1:

PR #2312, opened 2020-11-02
Bump NodeJS to v14 LTS

### LINKED ISSUE #2227: cli: fix wrtc bugs biting us

### LINKED ISSUE #2262: Remove forceExit flag for E2E test of the SDK in CircleCI
## Description
To fix #2242 we introduced a temporally fix with #2261 that is "hiding" this error to make the CI pipeline working again.
This can only happen when the origin issue with #2227 has been resolved.

## Acceptance criteria
- WebRTC bug was fixed as requirement
- The `--forceExit` got removed at the `test_sdk_e2e` job at CircleCI

## Tasks
- [ ]

### LINKED ISSUE #2311: Upgrade infrastructure to Node 14
Upgrade infrastructure to Node 14

andrevmatos:
Fixes #2311 
~~Fixes `#2227`~~
Fixes #2262 

**Short description**
Also:
- Some fixes for proper `yarn` cache management on CircleCI
- Call `close()` on [WebRTC connection](https://github.com/raiden-network/light-client/blob/dde6594a55c4cb14e0268c9e9493de09404682a8/raiden-ts/src/transport/epics/webrtc.ts#L415-L416); ensures the connection is properly teared down, as `wrtc` doesn't segfault on close anymore on Node v14, allowing proper teardown of RTC channels
  - Edit: the apparent infinite loop bug on `wrtc` is still present on this branch/node14, so #2227 isn't solved, but this improves  it and specially the connection teardown phase
- It was noticed the e2e tests were trying to shutdown before last db flush, which made it fail even though it could exit properly; to fix this, `Raiden.stop` was made asynchronous, and resolves after `db` is properly flushed and gracefully closed;
  - This shouldn't be needed on browsers, they're fast enough; only on `memory` adapter on integration/e2e tests this could be seen
- With both issues above fixed, we can now have proper teardown of e2e SDK tests; there were some infra issues, which were also addressed on this PR: now, e2e tests complete and exit cleanly! :tada: 

image

**Definition of Done**

1.
2.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,23 +1,18 @@
 version: 2.1
 
-var_1: &docker_image raidennetwork/lightclient-node-pnpm:latest
-var_2: &cache_key raiden-{{ .Branch }}-{{ checksum "yarn.lock" }}
-var_3: &working_dir ~/src
-var_4: &dapp_working_dir ~/src/raiden-dapp
+var_1: &working_dir ~/src
+var_2: &dapp_working_dir ~/src/raiden-dapp
 
-anchor_1: &package_lock_key
-  key: *cache_key
-
-anchor_2: &attach_options
+anchor_1: &attach_options
   at: *working_dir
 
-anchor_3: &executor_parameter
+anchor_2: &executor_parameter
   parameters:
     executor:
       type: executor
       default: base-executor
 
-anchor_4: &run_on_release_tag_only
+anchor_3: &run_on_release_tag_only
   filters:
     tags:
       only: /^v\d+\.\d+\.\d+$/
@@ -28,7 +23,7 @@ executors:
   base-executor:
     working_directory: *working_dir
     docker:
-      - image: *docker_image
+      - image: circleci/node:lts-browsers
 
   e2e-environment-executor:
     working_directory: *working_dir
@@ -76,11 +71,17 @@ jobs:
     working_directory: *working_dir
     steps:
       - attach_workspace: *attach_options
-      - restore_cache: *package_lock_key
+      - restore_cache:
+          name: Restore Yarn Package Cache
+          keys:
+            - raiden-v1-yarn-{{ checksum "yarn.lock" }}
+            - raiden-v1-yarn-
       - run: yarn install
       - save_cache:
-          <<: *package_lock_key
+          name: Save Yarn Package Cache
+          key: raiden-v1-yarn-{{ checksum "yarn.lock" }}
           paths:
+            - ./node_modules
             - ~/.cache # Cypress stores its stuff here.
       - persist_to_workspace:
           root: *working_dir
@@ -263,8 +264,8 @@ jobs:
           name: Run end-to-end tests
           command: >
             source /etc/profile.d/smartcontracts.sh &&
-            yarn workspace raiden-ts run test:e2e --
-            --ci --runInBand --forceExit
+            yarn workspace raiden-ts run test:e2e
+            --ci --runInBand
             --reporters=default --reporters=jest-junit
           environment:
             JEST_JUNIT_OUTPUT_DIR: "reports/junit"
--- a/e2e-environment/run-e2e-tests.sh
+++ b/e2e-environment/run-e2e-tests.sh
@@ -27,6 +27,8 @@ docker run --detach --rm \
   raidennetwork/lightclient-e2e-environment \
   >/dev/null
 
+sleep 5
+
 echo "Getting DEPLOYMENT_INFO from docker image '$DOCKER_CONTAINER_NAME' ..."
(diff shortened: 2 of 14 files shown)


andrevmatos:
I've built the new image successfully here. I'll push the upgradeto Docker Hub as soon as this gets merged.

.circleci/config.yml
```diff
@@ -76,11 +71,17 @@ jobs:
     working_directory: *working_dir
     steps:
       - attach_workspace: *attach_options
-      - restore_cache: *package_lock_key
+      - restore_cache:
+          name: Restore Yarn Package Cache
+          keys:
+            - raiden-v1-yarn-{{ checksum "yarn.lock" }}
```
palango, line 77: The old `cache_key` had a `.Branch` in it as well. might make sense to keep that.

e2e-environment/run-e2e-tests.sh
```diff
@@ -27,6 +27,8 @@ docker run --detach --rm \
   raidennetwork/lightclient-e2e-environment \
   >/dev/null
 
+sleep 5
```
palango, line 30: Why is this needed?

palango (approved):
Nice PR. Left two nits - handle at your discretion.

.circleci/config.yml
andrevmatos: IMO, there's no point into caching different branches differently. Even because usually different PRs would be seen as different branches. If the `yarn.lock` matches (and actually even if don't), we want the yarn cache to be... _cached_, so it speeds up the install for new runs. Not sure why `.Branch` was there before, but I fail to see a reason for that.

e2e-environment/run-e2e-tests.sh
andrevmatos: On local runs, I've seen the LCs starting too quickly after the container got started, and failed because it'd not be able to hit matrix/synapse in a usable state. Giving some time for the containers to start did fix it for me.

PR #2332, opened 2020-11-06
Fix overflows and add expiration checks

### LINKED ISSUE #2058: Missing overflow check for lock.amount
The code below is not using the decoder:

https://github.com/raiden-network/light-client/blob/27d308d6279ec7f8d797ea8fc5829738992d905c/raiden-ts/src/transfers/epics/locked.ts#L137

This may result in a overflow if the fee is high enough, unlikely, but it doesn't hurt to check

### LINKED ISSUE #2059: Missing overflow check for locked_amount
Like https://github.com/raiden-network/light-client/issues/2058 , in the code below a overflow check is missing:

https://github.com/raiden-network/light-client/blob/27d308d6279ec7f8d797ea8fc5829738992d905c/raiden-ts/src/transfers/epics/locked.ts#L167

### LINKED ISSUE #2063: Missing overflow check for transferred_amount
Similar to #2058 and #2059 when a locked transfer is successful the transferred_amount will be increased, and that must not overflow. Because of this, when a locked transfer must only be create iff the code below would not overflow:

https://github.com/raiden-network/light-client/blob/60903231bc2bcebc366b8431061bb20610ca5579/raiden-ts/src/transfers/epics/locked.ts#L257-L259

To reinforce, the overflow must be checked when the locked transfer is created, and not at unlock time, that would be too late.

andrevmatos:
Fixes #2058 
Fixes #2059 
Fixes #2063 

**Short description**
Adds overflow checks when sending (and mediating) and receiving transfers. Also, require expirations to not be after settleTimeout blocks in the future, so they always expire before channel can get closed and settled.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/transfers/epics/locked.ts
+++ b/raiden-ts/src/transfers/epics/locked.ts
@@ -131,20 +131,36 @@ function makeAndSignTransfer$(
     'expiration too soon',
   );
 
-  // fee is added to the lock amount, check for overflow
-  const amountWithFee = decode(
-    UInt(32),
-    action.payload.value.add(fee),
-    'overflow of locked amount with fee',
-  );
+  // fee is added to the lock amount; overflow is checked on locksSum below
+  const amountWithFee = action.payload.value.add(fee) as UInt<32>;
+  const expiration = BigNumber.from(
+    action.payload.expiration ||
+      Math.min(
+        state.blockNumber + revealTimeout * 2,
+        state.blockNumber + channel.settleTimeout - confirmationBlocks,
+      ),
+  ) as UInt<32>;
+  assert(expiration.lte(state.blockNumber + channel.settleTimeout), [
+    'expiration too far in the future',
+    {
+      expiration: expiration.toString(),
+      blockNumber: state.blockNumber,
+      settleTimeout: channel.settleTimeout,
+      revealTimeout,
+    },
+  ]);
   const lock: Lock = {
     amount: amountWithFee,
-    expiration: BigNumber.from(
-      action.payload.expiration || state.blockNumber + revealTimeout * 2,
-    ) as UInt<32>,
+    expiration,
     secrethash: action.meta.secrethash,
   };
-  const locksroot = getLocksroot([...channel.own.locks, lock]);
+  const locks = [...channel.own.locks, lock];
+  const locksSum = totalLocked(locks);
+  assert(
+    UInt(32).is(channel.own.balanceProof.transferredAmount.add(locksSum)),
+    'overflow on future transferredAmount',
+  );
+  const locksroot = getLocksroot(locks);
 
   log.info(
     'Signing transfer of value',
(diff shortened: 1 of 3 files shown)


raiden-ts/src/transfers/epics/locked.ts
```diff
@@ -131,20 +131,36 @@ function makeAndSignTransfer$(
...
   );
 
-  // fee is added to the lock amount, check for overflow
-  const amountWithFee = decode(
-    UInt(32),
-    action.payload.value.add(fee),
-    'overflow of locked amount with fee',
-  );
+  // fee is added to the lock amount; overflow is checked on locksSum below
+  const amountWithFee = action.payload.value.add(fee) as UInt<32>;
```
palango, line 135: total nit, but I'd group the expiration checks together and move the amount down.

raiden-ts/src/transfers/epics/locked.ts
```diff
@@ -131,20 +131,36 @@ function makeAndSignTransfer$(
...
+  const expiration = BigNumber.from(
+    action.payload.expiration ||
+      Math.min(
+        state.blockNumber + revealTimeout * 2,
+        state.blockNumber + channel.settleTimeout - confirmationBlocks,
```
palango, line 140: Don't we enforce that `settleTimeout >= 2 * revealTimeout`? So in the edge case we deduct the confirmation window...
I can't really pinpoint, but this looks strange

andrevmatos: On LC, at least, we don't. `revealTimeout` is expected to be smaller than `settleTimeout`, but we aren't sure of it either. On LC, `revealTimeout` is a config global value, while `settleTimeout` comes from the blockchain and can be as low as 20 blocks. `expiration` can come from the `transfer/request` action (e.g. in case of a mediated transfer), and therefore is validated below, but if not provided, we need this just as a default value. Of course, it needs to be larger than `revealTimeout` (which defines the dangerZone), and this check is also done both here and on receiving, but there's no actual protocol/security requirement how large it needs to be, there need to be just enough _time_ for a transfer to succeed, therefore this check here ensures we chose the minimum between available expirations, and let the expiration check fail if it isn't fit. Some tests include revealTimeout=50 and settleTimeout=60, and it works with this code.

palango: But why do we need to subtract `confirmationBlocks`? Shouldn't `state.blockNumber + channel.settleTimeout` be valid as well?

andrevmatos: We need to give **some time** for the transfer to succeed. We give confirmationBlocks here, only as a default (in case `2 * revealTimeout < settleTimeout`)

palango: Isn't `revealTimeout` exactly the time we assume a tx will need (maximally) to get mined?

andrevmatos: It is, for transactions. We need to give some time before this (danger zone) for the off-chain protocol to work.

palango: Ok. now I get it. Thanks for talking me through!

palango (approved):

PR #2333, opened 2020-11-06
Fix potential undesired handling of SecretReveal if transfer receiving is disabled

### LINKED ISSUE #2297: Make unsafe (remaining lock timeout < reveal_timeout) transfers traceable
In #2210 we had some problems because the python client didn't continue a payment that it deemed unsafe (remaining lock timeout < reveal_timeout) and didn't show that information.

We have the same check and don't log a potential error either.

https://github.com/raiden-network/light-client/blob/d65362c267857d1643194d47915948760d4beedd/raiden-ts/src/transfers/epics/locked.ts#L494-L510

## Tasks
- [x] Log a message describing the problem

### LINKED ISSUE #2299: Don't handle reveals when 'receive' capability is off
## Description

Currently the `RECEIVE` capability stops the client from asking for the secret:
https://github.com/raiden-network/light-client/blob/d65362c267857d1643194d47915948760d4beedd/raiden-ts/src/transfers/epics/locked.ts#L494-L510

If a `SecretReveal` was received (shouldn't happen) the client would handle it.

## Acceptance criteria
- 

## Tasks
- [x] Add the same capability check to the `transferSecretRevealEpic`
- [x] Expand and adjust tests

andrevmatos:
Fixes #2297 
Fixes #2299 

**Short description**
If we had receiving disabled, upon receiving a transfer, we would simply not continue with the protocol by requesting the secret from initiator, **but** if a malicious actor wanted to trigger us to still receive a transfer (maybe to trigger us to waste SVT on monitoring or waste on on-chain secret registers), they could still Reveal to us (even without a request) and we would accept and continue with the protocol (unlock, etc). This PR fixes it by ensuring we don't handle reveals at all if receiving is disabled.
Also, we add proper logs in case we ignore a received transfer, so it's easier to debug why it got ignored.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/transfers/epics/locked.ts
+++ b/raiden-ts/src/transfers/epics/locked.ts
@@ -491,23 +491,31 @@ function receiveTransferSigned(
       );
 
       let request$: Observable<Signed<SecretRequest> | undefined> = of(undefined);
-      if (
-        getCap(caps, Capabilities.RECEIVE) &&
-        locked.target === address &&
-        // only request secret if transfer don't expire soon
-        locked.lock.expiration.sub(revealTimeout).gt(state.blockNumber)
-      )
-        request$ = defer(() => {
-          const request: SecretRequest = {
-            type: MessageType.SECRET_REQUEST,
-            payment_identifier: locked.payment_identifier,
-            secrethash,
-            amount: locked.lock.amount,
-            expiration: locked.lock.expiration,
-            message_identifier: makeMessageId(),
+      if (locked.target === address) {
+        let ignoredDetails;
+        if (!getCap(caps, Capabilities.RECEIVE)) ignoredDetails = { reason: 'receiving disabled' };
+        else if (!locked.lock.expiration.sub(revealTimeout).gt(state.blockNumber))
+          ignoredDetails = {
+            reason: 'lock expired or expires too soon',
+            lockExpiration: locked.lock.expiration.toString(),
+            dangerZoneStart: locked.lock.expiration.sub(revealTimeout).toString(),
           };
-          return signMessage(signer, request, { log });
-        });
+        if (!ignoredDetails) {
+          request$ = defer(async () => {
+            const request: SecretRequest = {
+              type: MessageType.SECRET_REQUEST,
+              payment_identifier: locked.payment_identifier,
+              secrethash,
+              amount: locked.lock.amount,
+              expiration: locked.lock.expiration,
+              message_identifier: makeMessageId(),
+            };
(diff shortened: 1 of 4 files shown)


raiden-ts/src/transfers/epics/locked.ts
```diff
@@ -491,23 +491,31 @@ function receiveTransferSigned(
...
+        else if (!locked.lock.expiration.sub(revealTimeout).gt(state.blockNumber))
+          ignoredDetails = {
+            reason: 'lock expired or expires too soon',
+            lockExpiration: locked.lock.expiration.toString(),
+            dangerZoneStart: locked.lock.expiration.sub(revealTimeout).toString(),
           };
-          return signMessage(signer, request, { log });
-        });
+        if (!ignoredDetails) {
+          request$ = defer(async () => {
```
palango, line 504: Why change it to `async`?

raiden-ts/src/transfers/epics/locked.ts
```diff
@@ -491,23 +491,31 @@ function receiveTransferSigned(
...
-            expiration: locked.lock.expiration,
-            message_identifier: makeMessageId(),
+      if (locked.target === address) {
+        let ignoredDetails;
+        if (!getCap(caps, Capabilities.RECEIVE)) ignoredDetails = { reason: 'receiving disabled' };
+        else if (!locked.lock.expiration.sub(revealTimeout).gt(state.blockNumber))
+          ignoredDetails = {
+            reason: 'lock expired or expires too soon',
+            lockExpiration: locked.lock.expiration.toString(),
+            dangerZoneStart: locked.lock.expiration.sub(revealTimeout).toString(),
```
palango, line 501: Maybe we should add the `state.block` here as well?

palango (approved):
Great! Left two nits, but feel free to ignore.

raiden-ts/src/transfers/epics/locked.ts
andrevmatos: It's already a function returning a promise, declaring it as `async` just makes it more explicit, and doing so on `defer`s callback protects us in case this function eventually doesn't return a promise (I had trouble with this in the past, where a function I thought returned a promise was actually returning undefined, which is understood by `defer` as `EMPTY`)

raiden-ts/src/transfers/epics/locked.ts
andrevmatos: Current `state.blockNumber` is already logged all over the place, no need to log it again here IMO

PR #2341, opened 2020-11-09
Feature/dapp handle notification confirmation state

### LINKED ISSUE #2334: dapp: handle blockchain event related notifications properly according to confirmation property
## Description

Talel and I introduced the cool feature that notifications can have the optional properties of a target block number when this event becomes done/ready/finished/... This allowed us to introduce automatically a block counter in the notification card. When the counter reached zero, it was shown as successful. In the meantime, the rest of the whole notification remained the same.

This has some things we forgot here. We correctly handle that there could a `success` or `failure` variant of the event we see form the SDK. When we see success event and it is blockchain related the confirmation state is always undefined for the first time. Depending on the confirmation blocks this remains like that (events gets emitted for each new block again) until the confirmation property changes to a boolean value. This can be either `true` if the associated transaction got confirmed or `false` if there was a reorg of the blockchain. 
There are now two issues. First we never handle the case that the transaction could get reorged. Second we never care about the confirmation state itself. As soon as the counter reaches zero, we made it a confirmed notification. 

So we need to evaluate the confirmation property of an event. This should not be handled by the "stupid" notification card which shows the counter if required (i.e. it is a blockchain related notification). But the module which commits the notification object to the store need to do. We have introduced the option to update already existing notifications. This will allows to replace the "pending" notification with a counter with a different notification object that is a success or informs about an reorg issue. This issue differs from the error notification caused by a `failure` event. Thereby the notification card can remains simple, showing the counter if there is a given target block number in relation tot he current block number. The event listener is then responsible to watch out for the confirmation state and then replace the notification (as we do always, also for continuous unconfirmed events, just that they replace the object with itself [kinda stateless here]) with a success of reorg one.

## Acceptance criteria
- 

## Tasks
- [ ]

weilbith:
Fixes #2334

**Short description**
I'm not too happy with the implementation. I'm not sure if it is related to the in general RaidenService notification pushing code or.... I'm open to change everything.

Please give me feedback if this PR should continue and extend the other notifications that are blockchain related. We haven't done it at all for now, also not for the recent features.

**Definition of Done**

1. Open a new channel
2. Open right away (ignore loading pop-up) to the notifications
3. Observe the counter and description of the notification

There is no easy way to test a real blockchain reorg scenario...

### DIFF
--- a/raiden-dapp/src/components/notification-panel/NotificationDescriptionDisplay.vue
+++ b/raiden-dapp/src/components/notification-panel/NotificationDescriptionDisplay.vue
@@ -12,7 +12,7 @@
 </template>
 
 <script lang="ts">
-import { Component, Prop, Vue } from 'vue-property-decorator';
+import { Component, Prop, Watch, Vue } from 'vue-property-decorator';
 import AddressDisplay from '@/components/AddressDisplay.vue';
 
 @Component({
@@ -27,12 +27,13 @@ export default class NotificationDescriptionDisplay extends Vue {
   @Prop({ required: true })
   description!: string;
 
-  isAddress(address: string): boolean {
-    return this.addressRegEx.test(address);
+  @Watch('description', { immediate: true })
+  updateSplitDescription(newDescription: string): void {
+    this.splitDescription = newDescription.split(this.addressRegEx);
   }
 
-  mounted() {
-    this.splitDescription = this.description.split(this.addressRegEx);
+  isAddress(address: string): boolean {
+    return this.addressRegEx.test(address);
   }
 }
 </script>
--- a/raiden-dapp/src/store/notifications/mutations.ts
+++ b/raiden-dapp/src/store/notifications/mutations.ts
@@ -1,4 +1,5 @@
 import { MutationTree } from 'vuex';
+import { NotificationImportance } from './notification-importance';
 import { NotificationContext } from './notification-context';
 import {
   NotificationPayload,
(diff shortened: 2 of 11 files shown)


weilbith:
I have no clue why the test coverage for the SDK has dropped. Unfortunately the improved tests coverage of the dApp can't equalize it. So I took the chance to improve the dApp tests even more according to the recent changes.

raiden-dapp/src/store/notifications/mutations.ts
```diff
@@ -34,6 +35,7 @@ export const mutations: MutationTree<NotificationsState> = {
       received: new Date(),
       display,
       context: payload.context ?? NotificationContext.NONE,
+      importance: payload.importance ?? NotificationImportance.LOW,
```
taleldayekh, line 38: I'm still not sure why we have these importance levels?

raiden-dapp/src/components/notification-panel/NotificationDescriptionDisplay.vue
```diff
@@ -26,14 +26,15 @@ export default class NotificationDescriptionDisplay extends Vue {
 
   @Prop({ required: true })
   description!: string;
+  
+  @Watch('description', { immediate: true })
```
taleldayekh, line 30: Nice fix! 👍

taleldayekh (approved):
Nice and proper, thank you @weilbith

raiden-dapp/src/store/notifications/mutations.ts
weilbith: So far this is only used to show highly important notifications with the snackbar stuff.

weilbith:
@taleldayekh what do you answer to my initial PR question?

weilbith:
I opened the issue that was asked for. I now rebase the PR to include the fixup commits. When the CI runs through successfully, I'll merge it afterwards.

PR #2350, opened 2020-11-11
ci: Run e2e tests on each PR, remove nightly run

palango:
**Short description**

The e2e tests are quick enough to run them in every PR. This should also make problems visible earlier.

**Definition of Done**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -319,9 +319,15 @@ workflows:
       - test_sdk_integration:
           requires:
             - build_sdk
+      - test_sdk_e2e:
+          requires:
+            - build_sdk
       - test_dapp_unit:
           requires:
             - build_dapp
+      - test_dapp_e2e:
+          requires:
+            - build_sdk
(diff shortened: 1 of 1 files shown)


weilbith (commented):
Did you tried it?
In the past one of the issues was that attaching the workspace with the build sdk and dapp to the e2e environment executor has not worked out. At least not for the dApp. You remember that old PR?
An alternative is to use the recent feature of CircleCI to stack Docker images in an executor, while all but the first just provide services to the first one. But this was not working for me.

palango:
> In the past one of the issues was that attaching the workspace with the build sdk and dapp to the e2e environment executor has not worked out. At least not for the dApp. You remember that old PR?

Ahhh, I didn't even think about this. Sorry for the noise.

weilbith:
You don't need to close that. You can concentrate here so solve this issue.
Maybe this got even resolved with the better work of `yarn`. Just try it.

PR #2368, opened 2020-11-17
Workaround wrtc segfault on teardown by exiting after stop resolves

### LINKED ISSUE #2361: Multiple scenario failures because of "node did not stop cleanly"
BF2, MS1, MS2, MS3 and PFS8 failed tonight with the message `Exception: Node X did not stop cleanly`

I could not find any obvious problems in the logs.

What's a bit strange is that the scenario runs (especially the MS ones) take little more than 5 mins, but that error should only be thrown after 10mins:
```py
    def stop(self, timeout=600):  # 10 mins
        assert self._process is not None, "Can't call .stop() before .start()"
        self._process.send_signal(signal.SIGINT)
        if self._process.wait(timeout):
            raise Exception(f"Node {self._index} did not stop cleanly: ")
```

### Logs
- [MS2.zip](https://github.com/raiden-network/light-client/files/5536150/MS2.zip)
- [MS3.zip](https://github.com/raiden-network/light-client/files/5536152/MS3.zip)

andrevmatos:
Fixes #2361

**Short description**
On NodeJS 14+, fortunately, the process does not segfault on closing connection, so we can keep it (as it's correct), but does on teardown due to a double-free. The workaround is to force-exit the process when shutting down after stop resolves (db flushes), but before it tries to teardown on its own (which still causes the segfault).

**Definition of Done**

**Steps to manually test the change (CLI)**

1. Run 2 nodes connected to each other
2. Wait for the RTC connection to be established
3. Tells one of them to shutdown (either via SIGINT/SIGTERM or `POST /api/v1/shutdown`)
4. Check process exits cleanly, as well as partner after telling it to shutdown.

### DIFF
--- a/raiden-cli/src/index.ts
+++ b/raiden-cli/src/index.ts
@@ -210,18 +210,18 @@ function createLocalStorage(name: string): LocalStorage {
   return localStorage;
 }
 
-function unrefTimeout(timeout: number | NodeJS.Timeout) {
-  if (typeof timeout === 'number') return;
-  timeout.unref();
-}
-
 function shutdownServer(this: Cli): void {
   if (this.server?.listening) {
     this.log.info('Closing server...');
     this.server.close();
   }
-  // force-exit at most 10s after stopping raiden
-  unrefTimeout(setTimeout(() => process.exit(0), 10000));
+  // it's safe to call stop multiple times; this one to await for db flush;
+  // then we wait a little for any pending request to complete and force-exit cleanly after that,
+  // to skip wrtc segfault on process teardown
+  new Promise((resolve) => setTimeout(resolve, 10))
+    .then(() => this.raiden.stop())
+    .then(() => new Promise((resolve) => setTimeout(resolve, 1000)))
+    .then(() => process.exit(0));
 }
 
 function shutdownRaiden(this: Cli): void {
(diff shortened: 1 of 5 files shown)


raiden-cli/src/index.ts
```diff
@@ -210,18 +210,18 @@ function createLocalStorage(name: string): LocalStorage {
...
     this.server.close();
   }
-  // force-exit at most 10s after stopping raiden
-  unrefTimeout(setTimeout(() => process.exit(0), 10000));
+  // it's safe to call stop multiple times; this one to await for db flush;
+  // then we wait a little for any pending request to complete and force-exit cleanly after that,
+  // to skip wrtc segfault on process teardown
+  new Promise((resolve) => setTimeout(resolve, 10))
+    .then(() => this.raiden.stop())
+    .then(() => new Promise((resolve) => setTimeout(resolve, 1000)))
```
palango, line 223: Why do we need this? Couldn't we just kill the process after `stop` is finished?

andrevmatos: Usually, there may be some requests in-flight when this kicks in. Namely, matrix's set as offline. It's not critical to wait for this request to finish though, that's why it doesn't have a wait for it, but I thought it'd be convenient to wait a little more before forcing exit.

palango (approved):

Issue #2220, opened 2020-10-09
I want to know, that I have to withdraw the tokens from the UDC on the UDC screen


christianbrb:
## User Story
As Chris I want to know, that I have to withdraw the tokens from the UDC on the UDC screen so that I am not getting confused.

## Acceptance criteria
- It should follow the design
- The information area on the top is not getting implemented
- We use the already existing color-variables (incl. new account screen button colors)
- No text is smaller than 12px
- User Contract Deposit Balance instead of USER CONTRACT DEPOSIT BALANCE
- Without the area "Withdrawals in progress" -> Issue #2308 

## Tasks
- [x] Meeting with @sashseurat 
- [x] Design @sashseurat 

## Design
[The click-dummy](https://www.figma.com/proto/5H8EOoC2SntoXZuTtUZsyY/LC-UDC-Screen-Optimisation?node-id=14%3A27&viewport=-1799%2C1629%2C1&scaling=min-zoom)

![LC - No Planned Withdrawals](https://user-images.githubusercontent.com/32264127/98224364-aff8d900-1f53-11eb-9f1e-f44b30f3ba0d.png)

taleldayekh:
@christianbrb Not very obvious how to best solve this atm. Was thinking about the problem today and also discussing a bit with @weilbith . We decided to push this to when @sashseurat is back from vacations to discuss possible solutions with her.

taleldayekh:
@sashseurat Design?

sashseurat:
Click-dummy showcasing the solutions for UX/UI issues is here ready for review:
https://www.figma.com/proto/5H8EOoC2SntoXZuTtUZsyY/LC-UDC-Screen-Optimisation?node-id=42%3A0&viewport=-96%2C461%2C0.25&scaling=min-zoom

## The approach is:
- Defining the information hierarchy via clear visual structures.
- Using simple UX copy.
- The text giving info on the screen is not too prominent to minimize distraction from the main actionability on the screen.
- Balance is displayed in a dedicated, clear, and prominent location.
- Balance is the information users will need to have access to making informed decisions on this screen.
- Deposit / Withdraw buttons have labels for maximum accessibility and clearness.
- "Balance is too low" notice is shown in appropriate subtleness as it is not a critical error.
- Added planned withdrawal section dedicated to managing easily all items that are either pending or ready.
- When the user confirms withdrawal in the UDC screen via withdraw dialog - the planned withdrawal section has a new item in progress.
[...]

PR #2369, opened 2020-11-17
Feature/dapp planned udc withdrawal extension

### LINKED ISSUE #2308: I want to know how much blocks are left until I can withdraw my RDN / SVT from the main account
## User Story
As Chris I want to know how much blocks are left until I can withdraw my RDN / SVT from the main account so that I can get them.

## Acceptance criteria
- It is shown on the UDC screen
- I want to see the number of remaining blocks until the planned withdraw is complete
- I want to see the amount of RDN / SVT on the planned withdrawal
- We use our existing colors
- There is another issue for the rest of the screen https://github.com/raiden-network/light-client/issues/2220
- A refactoring to split the stores is done

## Design
[The click-dummy](https://www.figma.com/proto/5H8EOoC2SntoXZuTtUZsyY/LC-UDC-Screen-Optimisation?node-id=14%3A27&viewport=-1799%2C1629%2C1&scaling=min-zoom)
- Click on the Withdraw button to see the complete design

One part:
grafik

## Remark
- Are several planned withdrawals possible at a time? -> No it is not.

## Tasks
- [x] Design @sashseurat (can be done together with #2220)
- [ ] Write down the software design @weilbith

weilbith:

Fixes #2308

**Short description**
The first two commits includes minor issues which have produced noise.. Sorry for that.
Well, then go for it and do a withdraw and wait a while. I did it myself and it worked for me.
Note that while the new component from the last commit follows 100% the linked design model of Sash, the width inside the UDC screen is not yet correct. This is part of #2220 that reworks the structure and design of this screen.

**Definition of Done**

1. Go to the UDC screen of your account
2. Verify that you have no planned withdraws
3. Initiate a withdraw
4. Verify that you see a pending planned withdraw with down counting remaining blocks
5. Wait that the planned withdraw becomes ready and changes its appearance
6. Follow the link to actually withdraw the tokens
7. Go back to the UDC screen and verify that there is no planned withdrawal anymore (this eventually needs 1-2 blocks more to happen, since the blockchain events are not yet seen)


weilbith:
Note that the changelog hook is failing because there is this stupid commit in the SDK directory. I added an entry for the dApp which this PR is for.

weilbith:
Note that the changelog hook is failing because there is this stupid commit in the SDK directory. I added an entry for the dApp which this PR is for.

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -457,6 +462,24 @@ export default class RaidenService {
     });
   }
 
+  private async updatePlannedUserDepositWithdrawals(event: ObservedValueOf<Raiden['events$']>) {
+    if (event.type === 'udc/withdraw/success') {
+      if (event.payload.confirmed === false) {
+        this.store.commit('userDepositContract/clearPlannedWithdrawal');
```
taleldayekh, line 468: Not sure I understand why it has to be cleared for each event if the payload is unconfirmed?

raiden-dapp/src/components/PlannedUdcWithdrawalInformation.vue
```diff
@@ -0,0 +1,116 @@
+<template>
+  <div
+    class="planned-udc-withdrawal-information d-flex justify-center align-center px-2"
+    :style="{ color, backgroundColor }"
+  >
+    <template v-if="noWithdrawInProgress">
```
taleldayekh, line 6: Just out of curiosity, what are the benefits of using the `template` element instead of a, for example, `div`?

raiden-dapp/src/components/PlannedUdcWithdrawalInformation.vue
```diff
@@ -0,0 +1,116 @@
...
+  @Prop()
+  plannedWithdrawal!: PlannedUdcWithdrawal | undefined;
+
+  @Prop({ required: true })
+  udcToken!: Token;
+
+  @Prop({ required: true })
+  blockNumber!: number;
+
+  readonly assetSourceDirectory = '@/assets/planned-udc-withdrawal';
```
taleldayekh, line 70: Smart! :)

raiden-dapp/src/components/PlannedUdcWithdrawalInformation.vue
```diff
@@ -0,0 +1,116 @@
...
+
+  get isPending(): boolean {
+    return this.plannedWithdrawal !== undefined && this.remainingBlocksUntilReady > 0;
+  }
+
+  get isReady(): boolean {
+    return this.plannedWithdrawal !== undefined && this.remainingBlocksUntilReady <= 0;
+  }
+
+  get color(): string {
```
taleldayekh, line 91: Personally, I prefer the style to stay in the `styles` part of the component and use the conditionals to switch between `class` names instead if possible. Then I know that I only have to bother with the style section if I want to look for style-related stuff. Isn't that the beauty with Vue? :)

taleldayekh (approved):
Looks great. Good job! ❤️

raiden-dapp/src/services/raiden-service.ts
weilbith: What do you mean?
`confirmed === undefined` -> transaction not yet confirmed (not enough blocks on top of it, risk that block with transaction could get reorged)
`confirmed === true` -> transaction confirmed (enough blocks on top, unlikely that blockchain reorgs below)
`confirmed === false` -> block with transaction was reorged an not part of the (longest) blockchain anymore

Still open questions?

taleldayekh: Ah, that clarifies it, thank you!

raiden-dapp/src/components/PlannedUdcWithdrawalInformation.vue
weilbith: A `v-if` and according `v-else` directive can be only attached to one element in the template. So this basically adds the directive to all its childs, but itself is invisible. So I do not need to repeat if for each child, besides that I can't assign a `v-else` multiple times. The difference to a `div` is, that the directive gets only applied to the div itself. The div is its own object in the DOM and gets rendered as that. So the whole flow and DOM looks different.

```html
<template>
   <span>a</span>
   <img alt="b" />
</template>
```
becomes
```html
<span>a</span>
<img alt="b" />
```

But 
```html
<div>
   <span>a</span>
   <img alt="b" />
</div>
```
remains as it is.

This becomes relevant, because in the if-else scenario, I can define at the parent component the flex flow and style, no matter which case is true. If I put them into divs, the flow gets destroyed and I would need to repeat it or make it reusable.

Sorry for the long answer. But did it become clear?

taleldayekh: Ok nice, was not aware of that! 👍

raiden-dapp/src/components/PlannedUdcWithdrawalInformation.vue
weilbith: Hmm. Vue is not opinionated about that at all. But you have a good point. This is a really nice hint and I will definitely implement it. I like it especially, because it allows me to get rid of fore- and background color and group them in the classes. Thank you very much. :hugs:

weilbith: Done. Including an improvement to not split link text and icon (causes to replicate the click function and is stupid). The past issue was simple the vertical alignment. But I solved it better now, resolving in better "logic" and less code.

weilbith:
I cleaned up the branch and pushed again. But CI hangs/got not triggered. I'll amend the last commit to be able pushing it again.

PR #2393, opened 2020-11-24
Feature/dapp udc withdrawal navigation in notification

### LINKED ISSUE #2372: dApp: udc withdrawal plan and automatic final withdrawal user flow improvement
## Description

During the design of showing the progress of planned UDC withdrawals we did one mistake: the last final transaction that get done automatically.

This is what happens in the SDK:
- user triggers withdrawal
- SDK does withdrawal plan transaction
- plan gets written to contract, period where services can claim IOUs starts
- period ends and user is able to claim finally the tokens
- SDK automatically detects end of period and pushes second transaction to claim remaining tokens

What do we show:
- before: no withdrawal in progress
- after trigger: withdrawal in progress in pending with block counter for period
- after period: withdrawal in progress in ready state with link to withdrawal (claim screen)
- after final withdrawal tx: no withdrawal in progress

The issue:
There is a short period in which the SDK detects that the period is over and does the transaction until seeing the mined claim transaction. In this period we show the user the ready state and the navigation link. If he follows the link he will see no tokens to transfer from Raiden to main account, since the tokens are not yet claimed from contract. But when the tokens are automatically claimed by the SDK the link to the withdrawal screen disappears (e.g. when the user closes the screen/app while the period is running).

Decided on solution:
- the ready state does not show a link but a information that the token get claimed now.
- the notification for the final token claim from the contract includes link to navigate to withdrawal/transfer screen (this also shows/works when the user was offline/away for a while)

## Acceptance criteria
- 

## Tasks
- [ ]

weilbith:
Fixes #2372

**Short description**
The most relevant part is in the first two commits. The rest is not actually related to this issue. But since the PR was small enough, I decided to include the changes here. In fact they are mostly about removing all warnings from the unit tests and make it more clean again.

__Note:__
Some parts of how we handle notification icons and the payload in general is not too nice. So to get this feature done, I just followed the current structure for now. But I opened new issues to target this problems in general.

**Definition of Done**

1. Initiate an UDC withdrawal plan
2. Wait 100 blocks until the withdrawal plan becomes ready
3. Confirm the new message of the progress info component and that there is no link anymore
4. Wait for the automatic claim and the notification that gets send when the according Tx gets confirmed
5. Confirm that the notification has a navigation link to the account withdrawal screen

1.2. Run the unit tests of the dApp and confirm that there is no red warning output anymore, but just a green wall

### DIFF
--- a/raiden-dapp/src/components/PlannedUdcWithdrawalInformation.vue
+++ b/raiden-dapp/src/components/PlannedUdcWithdrawalInformation.vue
@@ -15,8 +15,8 @@
       <img
         class="planned-udc-withdrawal-information__status-icon mr-2"
         :src="require(`@/assets/planned-udc-withdrawal/status-icon-${statusIconName}.svg`)"
-        :width="iconSize"
-        :height="iconSize"
+        width="13px"
+        height="13px"
       />
 
       <amount-display
@@ -27,26 +27,14 @@
 
       <v-spacer />
 
-      <span v-if="isPending" class="planned-udc-withdrawal-information__pending-message">
+      <span v-if="isPending" class="planned-udc-withdrawal-information__message">
         {{ remainingBlocksUntilReady }}
-        {{ $t('udc.information.blocks-remaining') | upperCase }}
+        {{ $t('udc.information.pending-message') | upperCase }}
       </span>
 
-      <a
-        v-if="isReady"
-        class="planned-udc-withdrawal-information__navigation-link d-flex align-center"
-        @click="navigateToWithdrawal"
-      >
-        {{ $t('udc.information.confirmed-navigation-link') | upperCase }}
-
-        <img
-          v-if="isReady"
-          class="ml-2"
-          :src="require(`@/assets/planned-udc-withdrawal/navigation-arrow.svg`)"
-          :width="iconSize"
-          :height="iconSize"
-        />
-      </a>
+      <span v-if="isReady" class="planned-udc-withdrawal-information__message">
+        {{ $t('udc.information.ready-message') | upperCase }}
+      </span>
     </template>
   </div>
 </template>
@@ -70,7 +58,6 @@ export default class PlannedUdcWithdrawalInformation extends NavigationMixin {
   blockNumber!: number;
 
   readonly assetSourceDirectory = '@/assets/planned-udc-withdrawal';
-  readonly iconSize = '13.33px';
 
   get noWithdrawInProgress(): boolean {
     return this.plannedWithdrawal === undefined;
@@ -114,13 +101,8 @@ export default class PlannedUdcWithdrawalInformation extends NavigationMixin {
   }
 
   &--ready {
-    $color: #00ff66;
-    color: $color;
+    color: #00ff66;
     background-color: #003214;
-
-    a {
-      color: $color;
-    }
   }
 
   &__amount {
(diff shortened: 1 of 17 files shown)


raiden-dapp/src/components/account/AccountContent.vue
```diff
@@ -139,10 +139,7 @@ export default class AccountContent extends Mixins(NavigationMixin) {
         },
       });
 
-      const mainAccount = this.$raiden.getMainAccount();
```
palango, line 142: what does this mean when `usingSubkey` is `undefined`? That we don't know or that we don't use subkeys?

raiden-dapp/src/components/PlannedUdcWithdrawalInformation.vue
```diff
@@ -15,8 +15,8 @@
       <img
         class="planned-udc-withdrawal-information__status-icon mr-2"
         :src="require(`@/assets/planned-udc-withdrawal/status-icon-${statusIconName}.svg`)"
-        :width="iconSize"
-        :height="iconSize"
+        width="13px"
```
palango, line 18: why hardcode the image size?

raiden-dapp/src/locales/en.json
```diff
@@ -153,8 +153,8 @@
     "withdrawal-planned": "The withdrawal from the user deposit has been planned. You will get notified when the withdrawal is complete.",
     "information": {
       "no-withdrawal-in-progress": "No withdrawal in progress",
-      "blocks-remaining": "Blocks remaining...",
-      "confirmed-navigation-link": "ready・continue in withdrawal screen"
+      "pending-message": "Blocks remaining...",
+      "ready-message": "Ready・Tokens get claimed automatically"
```
palango, line 163: Is "・" the symbol we commonly use for this?

palango (commented):
Nice PR with great commit descriptions (as always 🎖️ )! Left some questions and I still need to test this.

raiden-dapp/src/components/PlannedUdcWithdrawalInformation.vue
weilbith: Because we don't need to vary it. Before it was a variable since it was a shared values for multiple icons. Now there is only one left, so this is not needed anymore. :shrug:

raiden-dapp/src/locales/en.json
weilbith: That was the decision of the designer and Christian approved it. So I just copied it.

raiden-dapp/src/components/account/AccountContent.vue
weilbith: We agreed to put an assert here to express in code that at this point it is not expected (and theoretically impossible) that this property is undefined here.

weilbith: And roll back again. Our linter does not accept an assert, since this is part of the `console` namespace.
Anyway, Paul and I discussed this again. The subkey property gets set within the connect function of the RaidenService. This means whenever the app is connected, this property must be defined. So since the block we are discussing here is within a conditional block that checks if the app is connected, this statement will always evaluate a boolean value and it can't ever be undefined. At least as the code is at the moment. And do not see a reason why this should change, since the RaidenService can't know it if is using a sub-key or not until the connect function gets called.

weilbith:
Flaky SDK integration tests? :thinking: 
I just ran the CI pipe again to test that...

palango (approved):
You just showed that it works in the demo. LGTM

weilbith:
So you missed the hidden cut while I was jumping over the 100 blocks withdraw period. :stuck_out_tongue_winking_eye: 
Just joking. Thanks for the review. :+1:

PR #2395, opened 2020-11-24
Disable receiving if blocks don't arrive in a timeline manner

### LINKED ISSUE #1256: Disable receiving if blocks don't arrive in a timely manner

andrevmatos:
Fixes #1256

**Short description**
This is done only if receiving is not force-enabled in config, so user's config have precedence. If blocks get stale (i.e. no new blocks after the time were 3 to 4 blocks should have been received), it'll auto-disable receiving until a new block comes.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/epics.ts
+++ b/raiden-ts/src/epics.ts
@@ -1,4 +1,16 @@
-import { Observable, from, of, combineLatest, using, throwError, merge } from 'rxjs';
+import {
+  Observable,
+  from,
+  of,
+  combineLatest,
+  using,
+  throwError,
+  merge,
+  timer,
+  defer,
+  EMPTY,
+  concat,
+} from 'rxjs';
 import {
   catchError,
   filter,
@@ -11,11 +23,18 @@ import {
   scan,
   distinctUntilChanged,
   ignoreElements,
+  switchMap,
+  exhaustMap,
+  mapTo,
+  endWith,
+  shareReplay,
+  withLatestFrom,
 } from 'rxjs/operators';
 import { MaxUint256 } from '@ethersproject/constants';
 import negate from 'lodash/negate';
 import unset from 'lodash/fp/unset';
 import isEqual from 'lodash/isEqual';
+import constant from 'lodash/constant';
 
 import { RaidenState } from './state';
 import { RaidenEpicDeps, Latest } from './types';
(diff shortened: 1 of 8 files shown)


andrevmatos:
I needed the average `blockTime` for this PR, and saw #2398 could make use of it as well, so I refactored the PR to also calculate and expose the `latest$.blockTime` as the public `Raiden.blockTime$` observable. It's only calculated every 10 blocks by default, since we want to avoid fetching the `block` structure every block only to get its `timestamp`, but it should be enough.

raiden-ts/src/raiden.ts
```diff
@@ -193,6 +197,7 @@ export class Raiden {
     this.state$ = latest$.pipe(pluckDistinct('state'));
     // pipe action, skipping cached
     this.action$ = latest$.pipe(pluckDistinct('action'), skip(1));
+    this.blockTime$ = latest$.pipe(pluckDistinct('blockNumber'));
```
palango, line 200: I don't get this. Shouldn't this be `blockTime`?

andrevmatos: Woosh.. good catch :see_no_evil:

andrevmatos: It's weird that this wasn't catched by the type system.. I think `pluck` type inference evaluates to `unknown`, and then that draws the type from the declared `blockTime$: Observable<number>` member, narrowing the inference, while it should better be narrowed as `never`.. I may look into it later, but already queued a fix.. Will push with a rebase after the end of your review

raiden-ts/src/epics.ts
```diff
@@ -106,28 +106,66 @@ function getBlockTime$(
...
+      if (config.caps !== null) config.caps = { ...defaultConfig.caps, ...userConfig.caps };
+      if (config.caps === null || config.caps[Capabilities.RECEIVE] != null) return of(config);
+      // if user config caps is not disabled, calculate dynamic default values
+      else
+        return combineLatest([udcBalance$, getStale$(state$, blockTime$, config)]).pipe(
+          map(([udcBalance, stale]) => ({
+            ...config,
+            caps: {
+              [Capabilities.RECEIVE]:
+                !stale && config.monitoringReward?.gt(0) && config.monitoringReward.lte(udcBalance)
```
palango, line 161: Maybe move this to a separate function to make it easier to read.

palango (approved):
Great! Left some minor remarks, but you can decide on them.

raiden-ts/src/epics.ts
```diff
@@ -35,6 +54,58 @@ import * as TransportEpics from './transport/epics';
...
+  return concat(
+    // like startWith, but if somehow this gets re-subscribed, uses lastInfo[2] as initialBlockTime
+    defer(() => (lastInfo[0] > 0 && lastInfo[2] ? of(lastInfo[2]) : of(initialBlockTime))),
+    state$.pipe(
+      pluckDistinct('blockNumber'),
+      filter((blockNumber) => blockNumber >= lastInfo[0] + fetchEach),
+      exhaustMap((blockNumber) => {
+        const prevInfo$ =
+          lastInfo[0] > 0 ? of(lastInfo) : getBlockInfo$(Math.max(1, blockNumber - fetchEach));
+        const curInfo$ = getBlockInfo$(blockNumber);
```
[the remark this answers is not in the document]
andrevmatos: We do cache it, it'll be `lastInfo` aka `prevInfo$` on next call. We do only a single request every 10 blocks: the current blockInfo, and compare that with the previous request, to calculate it; the only exception for this is  (the triple conditional above), on first request ever, where we need to request both `prevInfo` and `curInfo`, since we haven't any by then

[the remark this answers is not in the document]
palango: Ah, I missed that.

raiden-ts/src/epics.ts
andrevmatos: :heavy_check_mark:

PR #2396, opened 2020-11-24
Allow batch messages received in wrtc to be processed

fredo:

This is a small PR to allow batched messages to be processed. Currently, the Python Client batches messages unconditionally and sends them via any communication channel possible. To be compatible the light client needs to separate the different batched messages and process them one by one.


andrevmatos (approved):
Lgtm, ty

PR #2411, opened 2020-11-26
Bugfix/dapp notification icon unefined

### LINKED ISSUE #2391: dapp: notification icon handling is buggy
## Steps to reproduce it
1. Remove an icon from a notification
2. Run the dApp

## Expected
- The notification is shown
- If no icon is available a fallback icon is shown

## Actual
- The notification is not shown

## Remarks
At some point we introduced icons as an optional property of a notification. But it is not handled as optional. This means if the icon is not set (which is by code allowed and accepted), the notification card is broken and does not get displayed.
Unfortunately we don't get supported by the linters here which do a bad job in checking the templates. Maybe a hint to put most logic into the component class itself to circument this.
In general is the treatment of optional fields during the creation of notifications a horrible process which has grown over time. Also do we not test the full flow of notifications from the SDK to the UI. Which is fine. But this usually means such mistakes only get detected when you try it manually. And since some notifications are hard to test since they are complicated to trigger (tx reorgs for example), this gets discovered usually too late.

weilbith:
Fixes #2391

**Short description**
The first commit includes some test refactors in preparation, so just read the commit individually.
The icon was provided by Sash.
Unfortunately it is not really possible to actually test the rendered icon. But the test do fail if the icon related asset could not been resolved.

**Definition of Done**

1. The only notification without an icon atm is the balance prove sent one.
2. Manipulate the open channel notification to do not use an icon.

### DIFF
--- a/raiden-dapp/CHANGELOG.md
+++ b/raiden-dapp/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- [#2391] Fix notification icon handling for special scenarios
+
+[#2391]: https://github.com/raiden-network/light-client/issues/2391
+
+
 ## [0.14.0] - 2020-11-25
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


andrevmatos (approved):
lgtm!

Issue #211, opened 2019-07-23
I want to get hubs proposed which are well connected (Autopilot)


christianbrb:
## User Story

As Chris I want to get hubs proposed which are well connected so that I can connect to a hub through which I can send transfers to a potentially high number of nodes.

## Acceptance criteria
- It is possible to select one hub
- It is possible to enter the address of a hub
- As it is time-critical it needs to be easy to implement

## Tasks
- [ ] Design @sashseurat 
- [ ] Implement in the SDK (@andrevmatos)
- [ ] Implement in the dApp

## Ideas
- List dropdown
- Using the PFS list
- Adding a button and use a modal 

## Screenshot of the current implementation

![grafik](https://user-images.githubusercontent.com/35398162/98559220-a4d0e080-22a6-11eb-8d1c-e7256ba01f37.png)

andrevmatos:
Duplicated of #155, but this title is better explanatory, closing that in favor of this one.

christianbrb:
@andrevmatos We have estimated 100 points, but maybe due to our limited knowledge of the matrix protocol

christianbrb:
#691 is a short term workaround

christianbrb:
#691 is a short term workaround

andrevmatos:
Ideally, this wouldn't involve much matrix protocol meddling, but instead rely on PFS suggesting well connected and high uptime partners for us to connect to, possibly also taking into account higher probabilities on default configurations of being counter-deposited.
A local algorithm can also be developed, but as it'd only have access of instantaneous blockchain & matrix data, it'd always be less precise than long-living PFS data. Eventually, some on-chain reputation system is also expected to arise for this.

christianbrb:
We can do it partly
- Having a basic list of suggestions

@Dominik1999 It would be great if we could have https://github.com/raiden-network/raiden-services/issues/684 This would greatly simplify the onboarding of new users for the Light Client and would avoid us implementing an autopilot (Connection Manager).

christianbrb:
Let's first create a small solution and wait for https://github.com/raiden-network/raiden-services/issues/684 to be implemented
1. Partner / Hub is online
2. Clients with 3 or 4 channels 
a) Connected with the connection manager
b) Counter-deposited
3. The highest number of channels

@Dominik1999 We should talk about the standard setting for counter-depositing to enable receiving from very well connected hubs.

christianbrb:
[...]

Issue #2324, opened 2020-11-05
sdk-integration-tests: runtime regression


palango:
Recently runtime of the `sdk-integration-tests` went from 6mins to 12mins. We should investigate if we can recover some/all of those losses.

@andrevmatos noted that https://github.com/raiden-network/light-client/pull/2289 might be the culprit:
> A nice and easy way to optimize that by a lot would be to make our integration tests use-case based, instead of testing each function separately
> e.g. a happy case test: channel open -> deposit -> transfer -> withdraw -> close -> settle -> unlock
> than failure cases for individual steps, possibly making better use of the snapshoting feature of ganache (we already have utils for it)
> This way, we could decrease the number of integration tests by a lot, and allow finer-granularity tests to depend on a given previous state (e.g. sneaking in a get channels$ test after channel got opened on the happy case test
> Most of the time is spent on setting up and resetting these integration tests
> I know “mixing” tests isn’t very good, but since the setup times here are so long, and we already have extensive unit tests, maybe we could consider this refactoring..pretty sure we could get it down to a couple of minutes at most
> The subkey integration test is an example of a more extensive test which works very well, and is fast despite testing maybe half of what the other tests does (but here using the subkey)

andrevmatos:
#2417 improved a lot on this, integration tests there are taking 7min, but we should still refactor this at some point in the future

PR #2412, opened 2020-11-26
Add suggestedPartners method

andrevmatos:
SDK part of #211 

**Short description**
Adds a `Raiden.suggestPartners` method, which fetches and validates the suggested partners from PFS.

It also fixes a small issue when not all environment variables were set, partial config handled to `Raiden.create` contained `undefined` values which overwrote default ones. Now, this gets properly cleaned before creating default config, allowing the dApp to be served with something like `--mode=testnet` to skip loading demoenv001 variables.

This was required because current demo001 PFS doesn't support suggested partners.

**Definition of Done**

1. On development mode, get the SDK instance on DevTools console (e.g. in a var named `raiden`)
2. call the method to get the response: `await raiden.suggestPartners('0x59105441977ecD9d805A4f5b060E34676F50F806')`

### DIFF
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -12,10 +12,13 @@ import { createLogger } from 'redux-logger';
 
 import constant from 'lodash/constant';
 import memoize from 'lodash/memoize';
+import isUndefined from 'lodash/isUndefined';
+import omitBy from 'lodash/omitBy';
 import { Observable, AsyncSubject, merge, defer, EMPTY, ReplaySubject, of } from 'rxjs';
-import { first, filter, map, mergeMap, skip, pluck } from 'rxjs/operators';
+import { first, filter, map, mergeMap, skip, pluck, timeout } from 'rxjs/operators';
 import logging from 'loglevel';
 
+import { fromFetch } from 'rxjs/fetch';
 import {
   TokenNetworkRegistry__factory,
   TokenNetwork__factory,
@@ -63,7 +66,15 @@ import {
   transferKeyToMeta,
 } from './transfers/utils';
 import { pathFind, udcWithdraw, udcDeposit } from './services/actions';
-import { Paths, RaidenPaths, PFS, RaidenPFS, IOU } from './services/types';
+import {
+  Paths,
+  RaidenPaths,
+  PFS,
+  RaidenPFS,
+  IOU,
+  SuggestedPartner,
+  SuggestedPartners,
+} from './services/types';
 import { pfsListInfo } from './services/utils';
 import { Address, Secret, Storage, Hash, UInt, decode } from './utils/types';
 import { isActionOf, asyncActionToPromise, isResponseOf } from './utils/actions';
@@ -85,6 +96,7 @@ import { RaidenError, ErrorCodes } from './utils/error';
 import { RaidenDatabase } from './db/types';
 import { dumpDatabaseToArray } from './db/utils';
 import { createPersisterMiddleware } from './persister';
+import { jsonParse } from './utils/data';
 
 export class Raiden {
   private readonly store: Store<RaidenState, RaidenAction>;
@@ -183,10 +195,7 @@ export class Raiden {
     this.resolveName = provider.resolveName.bind(provider) as (name: string) => Promise<Address>;
     this.log = logging.getLogger(`raiden:${address}`);
 
-    const defaultConfig = makeDefaultConfig(
-      { network },
-      config && decode(PartialRaidenConfig, config),
-    );
+    const defaultConfig = makeDefaultConfig({ network }, config);
 
     // use next from latest known blockNumber as start block when polling
     provider.resetEventsBlock(state.blockNumber + 1);
@@ -341,7 +350,7 @@ export class Raiden {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     storage?: { state?: any; storage?: Storage; adapter?: any; prefix?: string },
     contractsOrUDCAddress?: ContractsInfo | string,
-    config?: PartialRaidenConfig,
+    config?: { [k: string]: unknown },
     subkey?: true,
   ): Promise<InstanceType<R>> {
     let provider: JsonRpcProvider;
@@ -399,13 +408,14 @@ export class Raiden {
         },
       ],
     );
+    const cleanConfig = config && decode(PartialRaidenConfig, omitBy(config, isUndefined));
 
     return new this(
       provider,
       network,
       signer,
       contractsInfo,
-      { db, state, config },
+      { db, state, config: cleanConfig },
       main,
     ) as InstanceType<R>;
   }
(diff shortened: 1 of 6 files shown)


raiden-ts/src/raiden.ts
```diff
@@ -399,13 +408,14 @@ export class Raiden {
         },
       ],
     );
+    const cleanConfig = config && decode(PartialRaidenConfig, omitBy(config, isUndefined));
```
weilbith, line 411: Pretty neat code. Sometimes `lodash` chains are quite useful. :grimacing: 
Please close this conversation after reading.

raiden-ts/src/raiden.ts
```diff
@@ -1387,6 +1397,35 @@ export class Raiden {
...
+   * @param options.pfs - PFS to use, instead of configured or automatic
+   * @returns Ordered array of suggested partners, with address and scoring values according to PFS
+   */
+  public async suggestPartners(
+    token: string,
+    options: { pfs?: RaidenPFS } = {},
+  ): Promise<SuggestedPartner[]> {
+    assert(Address.is(token), [ErrorCodes.DTA_INVALID_ADDRESS, { token }], this.log.info);
+    const tokenNetwork = this.state.tokens[token];
+    assert(tokenNetwork, ErrorCodes.RDN_UNKNOWN_TOKEN_NETWORK, this.log.info);
```
weilbith, line 1415: Do you think it is worth to test this error? Should be a fairly easy test case.
It is not about making sure the code is working, but it is more a behavior check that we expect. Since this is an interface function where the client can us feed with anything. So we have the requirement to this function that it needs to validate the token address before doing the request.

raiden-ts/src/raiden.ts
```diff
@@ -1387,6 +1397,35 @@ export class Raiden {
...
+    const pfs = decode(PFS, options.pfs ?? (await this.findPFS())[0]);
+
+    return await fromFetch(`${pfs.url}/api/v1/${tokenNetwork}/suggest_partner`, {
+      method: 'GET',
+      headers: { 'Content-Type': 'application/json' },
+    })
+      .pipe(
+        timeout(this.config.httpTimeout),
```
weilbith, line 1423: Do we eventually do any retries here? I mean the method is asynchronous. It must not block the application with a correct design.

raiden-ts/src/raiden.ts
```diff
@@ -1387,6 +1397,35 @@ export class Raiden {
...
+        mergeMap((response) => response.text()),
+        map((text) => decode(SuggestedPartners, jsonParse(text))),
```
weilbith, line 1425: What if we have some unexpected decode issues? Let's say there was a version update, the format changed and we are not prepared or... just imagine anything. This would just forward the origin error to the user right? :thinking:

raiden-ts/src/utils/types.ts
```diff
@@ -343,3 +343,20 @@ export function last<T extends any[]>(arr: T): Last<T> {
...
+  ? BigNumberish
+  : T extends string
+  ? string
+  : T extends boolean
+  ? boolean
+  : T extends number
+  ? number
+  : T extends null | symbol
+  ? T
+  : { [K in keyof T]: Decodable<T[K]> };
```
weilbith, line 362: Pretty impressive approach. I would love if we could quickly talk about it so I can understand the full scope. Would be good to invite Paul as well. Just a short session.

raiden-ts/tests/integration/raiden.spec.ts
```diff
@@ -1479,4 +1479,49 @@ describe('Raiden', () => {
     // subkey is emptied of ETH
     expect((await sub.getBalance()).isZero()).toBe(true);
   });
+
+  test('suggestPartners', async () => {
```
weilbith, line 1483: This test case is packed with fixtures and preparations. This makes the test itself hard to read and understand what it is actually testing. Although adding new tests for this function is hard, since all the fixtures are not re-usable. Would be amazing if you would consider to improve here. :upside_down_face:

weilbith (approved):
Cool work. Actually wanted to tackle it myself this morning. :laughing: 
None of the comments are blocking. But it would be cool if you could tackle them before merging.

raiden-ts/src/raiden.ts
andrevmatos: I don't think this is worth to be tested. We test the happy case, and the same pattern is used elsewhere and tested for the unhappy case, and this is just simple validation.

raiden-ts/src/raiden.ts
andrevmatos: No, the request here is one-shot, it may reject in case of failures, and it's up to the user to retry. You could offer a retry button in case of errors, but errors here shouldn't happen unless on network connectivity issues.

raiden-ts/src/raiden.ts
andrevmatos: Unexpected decode issues will reject the same as network errors, but with its proper error messages. We may include an e2e test for this, once this gets implemented on the dApp. If there's unexpected issues, the best we can do is forward it to the user, they're unexpected afterall.

raiden-ts/tests/integration/raiden.spec.ts
andrevmatos: Unfortunately, these integration tests are very specific and hard to make generic fixtures for. #2324 tracks a refactoring of them, we could try to improve it there, but specifically here, I hardly can see how it'd be.

raiden-ts/src/raiden.ts
weilbith: Then I think we do a bad job in testing the expected behavior.

raiden-ts/src/raiden.ts
weilbith: Hmm. Not too happy with it.

raiden-ts/tests/integration/raiden.spec.ts
weilbith: Well, for example decare such fixed objects like `suggestResponse` and the mock implementations outside of this test case? They are not the test. 80% of the code is unrelated and I have to parse it myself in the head to see what this test is actually doing. with this pattern unit tests become incredibly hard to to maintain.

andrevmatos: These are integration tests. Defining mock values somewhere else helps unclog the test, but otoh makes it harder to wrap your mind around what the test does and what's expected data for a given endpoint/method, and splitting it is less useful when the data is useful/relevant only for a single test, but I agree, this can/should be improved, but then it's a matter of refactoring the tests and there's that issue for it.

raiden-ts/src/raiden.ts
andrevmatos: I understand, but if we'd code for every scenario where an upstream behavior/schema/protocol/api could change, this would be unfeasible.

PR #2424, opened 2020-12-01
dApp: sort transactions latest first in transaction list

### LINKED ISSUE #2415: Bad transfer history ordering
## Steps to Reproduce
1. Send some transfers
2. Look at the transfer history

## Expected Result
- Latest transfer shown first, as I'm potentially more interested in those

## Actual Result
- Latest transfer shown last, making it necessary to scroll

Screenshot 2020-11-27 at 12 22 25

taleldayekh:
Fixes #2415

**Short description**
Fixes issues with transactions in the transactions list appearing in random order. Now the get sorted latest first.

**Definition of Done**

1. Make some transfers and check the order of them.
2.

### DIFF
--- a/raiden-dapp/CHANGELOG.md
+++ b/raiden-dapp/CHANGELOG.md
@@ -4,9 +4,11 @@
 
 ### Fixed
 
+- [#2415] Fix sorting latest transaction on top in transactions list
 - [#2391] Fix notification icon handling for special scenarios
 - [#2410] Fix transfer history list not showing third entry initially
 
+[#2415]: https://github.com/raiden-network/light-client/issues/2415
 [#2391]: https://github.com/raiden-network/light-client/issues/2391
 [#2410]: https://github.com/raiden-network/light-client/issues/2410
 
--- a/raiden-dapp/src/components/transaction-history/TransactionList.vue
+++ b/raiden-dapp/src/components/transaction-history/TransactionList.vue
@@ -23,9 +23,9 @@
 import { Component, Prop, Vue } from 'vue-property-decorator';
 import { mapState } from 'vuex';
 import { Transfers } from '../../types';
-import { RaidenTransfer } from 'raiden-ts';
 import { Token } from '@/model/types';
 import Transaction from '@/components/transaction-history/Transaction.vue';
+import { RaidenTransfer } from 'raiden-ts';
 
 @Component({
   components: {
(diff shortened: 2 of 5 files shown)


palango:
Thanks for the PR, but I'm unsure if that fixes the problem. As in the screenshot below, there seems to be no ordering at all, while your fix indicates that the ordering was just inverted.
Screenshot 2020-12-02 at 11 15 24

Could you also add a test for this.

taleldayekh:
> Thanks for the PR, but I'm unsure if that fixes the problem. As in the screenshot below, there seems to be no ordering at all, while your fix indicates that the ordering was just inverted.
> Screenshot 2020-12-02 at 11 15 24
> 
> Could you also add a test for this.

Sure I will add a test. Tried it manually worked as expected then when changing from milliseconds to date. But I'll add tests as well. 👍

taleldayekh:
@palango actually, we already have a test for the sorting [here](https://github.com/raiden-network/light-client/blob/c58b292f2ed6494939362ba1478951c30f1fda78/raiden-dapp/tests/unit/components/transaction-history/transaction-list.spec.ts#L65). However I extended the test data with more dates.

palango:
> @palango actually, we already have a test for the sorting [here](https://github.com/raiden-network/light-client/blob/c58b292f2ed6494939362ba1478951c30f1fda78/raiden-dapp/tests/unit/components/transaction-history/transaction-list.spec.ts#L65). However I extended the test data with more dates.

Why didn't that one find the problems above?

taleldayekh:
> > @palango actually, we already have a test for the sorting [here](https://github.com/raiden-network/light-client/blob/c58b292f2ed6494939362ba1478951c30f1fda78/raiden-dapp/tests/unit/components/transaction-history/transaction-list.spec.ts#L65). However I extended the test data with more dates.
> 
> Why didn't that one find the problems above?

getMilliseconds() seems to return the current milliseconds in a given date, so this:

```javascript
let date = new Date("June 5, 1986 20:22:10:565");  
console.log(date.getMilliseconds())
```

will give me `565` which is irrelevant for comparing dates.

getTime, however, returns me the milliseconds from Jan 1, 1970. The above example with getTime() would give me `518379730565` which can be used for comparison.

The reason why the tests still passed I believe is that the milliseconds are not added at all in the test data, so the date in our test data only looks like this `new Date("June 5, 1986 20:22:10")`.

palango (approved):
Thanks for the explanation. That makes sense! Please add a changelog entry as well!

PR #2429, opened 2020-12-03
dApp: hub proposal when selecting hub

### LINKED ISSUE #211: I want to get hubs proposed which are well connected (Autopilot)

taleldayekh:
Fixes #211 

**Short description**
This PR introduces suggested partners for the dApp when connecting to a token network.
Screenshot 2020-12-11 at 17 44 44

Screenshot 2020-12-11 at 18 08 44

**Definition of Done**

1. Chose to connect to a new well-connected token (a list of three hubs should be displayed).
2. Chose to connect to a token with no partners (a message about not being able to fetch a list is displayed).


raiden-dapp/src/locales/en.json
```diff
@@ -529,8 +529,12 @@
     "select-button": "Select Hub",
     "service-token-tooltip": "Mint and deposit {token} to the User Deposit Contract to pay for the Monitoring Service.",
     "service-token-tooltip-main": "Deposit {token} to the User Deposit Contract to pay for the Monitoring Service.",
-    "service-token-balance-too-low": "Your {token} balance is too low to pay for the Monitoring Service. Use the button above to mint and deposit more tokens.",
-    "service-token-description": "Balance to pay the Monitoring Service"
+    "service-token-description": "BALANCE FOR MONITORING SERVICE"
```
palango, line 532: why is this all caps?

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -738,6 +739,10 @@ export default class RaidenService {
     const tokenBalance = await this.raiden.getTokenBalance(tokenAddress, raidenAccount);
     return tokenBalance;
   }
+
+  async getSuggestedPartners(token: string): Promise<SuggestedPartner[]> {
+    return await this.raiden.suggestPartners(token);
```
palango, line 746: Just for my understanding, shouldn't `await this.raiden.suggestPartners(token)`  already resolve to `SuggestedPartner[]`? Why do we need another await?

raiden-dapp/src/locales/en.json
taleldayekh: Dunno, designers decision 😉

raiden-dapp/src/services/raiden-service.ts
taleldayekh: Do you mean it should be sufficient to return right away like: `return this.raiden.suggestPartners(token)`?

raiden-dapp/tests/unit/utils/data-generator.ts
```diff
@@ -121,3 +122,20 @@ export function generateNotificationPayload(
...
+
+/**
+ * @param partialSuggestedPartner - SuggestedPartner overrides
+ * @returns SuggestedPartner mocked object
+ */
+export function generateSuggestedPartner(
+  partialSuggestedPartner: Partial<SuggestedPartner> = {},
+): SuggestedPartner {
+  return {
+    address: '0x123',
```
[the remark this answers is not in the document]
taleldayekh: Ok, I will update with a valid address 👍🏼

raiden-dapp/tests/unit/views/select-hub-route.spec.ts
```diff
@@ -116,12 +116,9 @@ describe('SelectHubRoute.vue', () => {
     );
   });
 
-  test('displays error if UDC capacity is not sufficient', async () => {
+  test('disable button if UDC capacity is not sufficient', async () => {
```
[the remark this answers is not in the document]
taleldayekh: I'll adjust that 👍🏼

raiden-dapp/src/locales/en.json
palango: Shouldn't we then use something like `text-transform: uppercase`?

raiden-dapp/src/services/raiden-service.ts
palango: Yeah, or not return a promise

taleldayekh: As I understand it, since the parent functions return a promise, this one would have too until it bubbles up in the component where the data is used.

raiden-dapp/src/locales/en.json
taleldayekh: Ah, yes, of course, that makes more sense.

raiden-dapp/src/services/raiden-service.ts
andrevmatos: Yes, you must type the return as a Promise for any async function. You can `await` on the call or not, the behavior is almost the same due to promise's automatic unfolding. There's some slight differences when handling rejections, I usually prefer the explicit `await` because of that, but as I said, here it's almost the same.

raiden-dapp/src/components/HubList.vue
```diff
@@ -0,0 +1,170 @@
...
+  selectHub(): string {
+    return this.selectedHub;
+  }
+
+  async mounted() {
+    try {
+      this.suggestedHubs = await this.$raiden.getSuggestedPartners(this.tokenAddress);
+      this.loadingHubs = false;
+    } catch (err) {
+      this.loadingHubs = false;
```
palango, line 63: You could also but this inside a `finally` block to reduce duplication

raiden-dapp/tests/unit/components/hub-list.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    expect(spinner.exists()).toBe(true);
+  });
+
+  test('displays message if no hubs available', async () => {
+    const wrapper = createWrapper(undefined);
+    await flushPromises();
+
+    const hubList = wrapper.find('.hub-list__no-hubs');
+
+    expect(hubList.text()).toContain('hub-list.error');
```
palango, line 53: Mh, should we handle errors and no hubs differently?

One example might be that no nodes are online, but the request succeeded.

palango (approved):
Great, another of the missing features implemented!

I left two more comments. I think we should differentiate between errors (PFS didn't answer) and no results (PFS didn't return nodes), but this can be done in a follow-up at a later point.

Regarding the presence problems. From all I looked into and saw, this is the presence bug striking. The nodes aren't seen online by anyone but the PFS, so nothing can be done in this PR.

raiden-dapp/src/components/HubList.vue
taleldayekh: That's a good idea, I'll put it in a finally block.

raiden-dapp/tests/unit/components/hub-list.spec.ts
taleldayekh: Yes, let's do this in a follow-up PR since I'm not quite sure what errors are relevant to display. So for now we just display this generic error.

PR #2436, opened 2020-12-07
Shorter expiration

### LINKED ISSUE #2409: Lower transfer expiration for dApp
In https://github.com/raiden-network/team/issues/894 and https://github.com/raiden-network/team/issues/891 we made the case for a shorter time between receiving a secret and the on-chain transaction in case we don't receive an unlock.

A decision has been made, and we need to implement it for the dApp (and possibly the CLI).

- SDK
  - [x] Make lock expiration configurable
  - [x] Add test for the minimal expiration
- dApp
  - [x] Use lower expiration by default
- CLI
  - [x] Decide if lower expirations should be used

palango:

Fixes #2409 

**Short description**

Instead of hard-coding the lock expiry to 2 times the reveal timeout, this introduces the `expiryFactor`, which multiplied with the reveal timeout returns the lock expiry.

~~An alternative would be to use absolute block numbers here, I can change that if people don't like the relative value.~~

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-dapp/src/services/raiden-service.ts
+++ b/raiden-dapp/src/services/raiden-service.ts
@@ -74,6 +74,9 @@ export default class RaidenService {
           ...(process.env.VUE_APP_CONFIRMATION_BLOCKS && +process.env.VUE_APP_CONFIRMATION_BLOCKS
             ? { confirmationBlocks: +process.env.VUE_APP_CONFIRMATION_BLOCKS }
             : undefined),
+          ...(process.env.VUE_APP_EXPIRY_FACTOR && +process.env.VUE_APP_EXPIRY_FACTOR
+            ? { expiryFactor: +process.env.VUE_APP_EXPIRY_FACTOR }
+            : undefined),
         },
         subkey,
       );
--- a/raiden-ts/src/transfers/epics/locked.ts
+++ b/raiden-ts/src/transfers/epics/locked.ts
@@ -127,9 +127,7 @@ function makeAndSignTransfer$(
   const channel = getOpenChannel(state, { tokenNetwork, partner });
 
   assert(
-    !action.payload.expiration ||
-      // require mediated transfers to have at least confirmationBlocks before danger zone
-      action.payload.expiration >= state.blockNumber + revealTimeout + confirmationBlocks,
+    !action.payload.expiration || action.payload.expiration > state.blockNumber + revealTimeout,
     'expiration too soon',
   );
   const expiration = BigNumber.from(
--- a/raiden-ts/src/transfers/epics/locked.ts
+++ b/raiden-ts/src/transfers/epics/locked.ts
@@ -101,6 +101,7 @@ function getOpenChannel(
  * @param config - Config object
  * @param config.revealTimeout - The reveal timeout for the transfer.
  * @param config.confirmationBlocks - Confirmations block config
+ * @param config.expiryFactor - The factor that the reveal timeouts gets multiplied with to calculate the lock expiration
  * @param deps - {@link RaidenEpicDeps}
  * @param deps.log - Logger instance
  * @param deps.address - Our address
@@ -111,7 +112,7 @@ function getOpenChannel(
 function makeAndSignTransfer$(
   state: RaidenState,
   action: transfer.request,
-  { revealTimeout, confirmationBlocks }: RaidenConfig,
+  { revealTimeout, confirmationBlocks, expiryFactor }: RaidenConfig,
   { log, address, network, signer }: RaidenEpicDeps,
 ): Observable<transferSecret | transferSigned> {
   // assume paths are valid and recipient is first hop of first route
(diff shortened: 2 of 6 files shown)


palango:
@andrevmatos Can you give this a quick lookand let me know if you agree with the approach. Then I write the tests tomorrow.

andrevmatos (approved):
I like the approach. Maybe we should default to 1.1, so it's also used on the CLI, I don't see a point in making them diverge, even if both should accept transfers using both factors (as well as raiden-py, as you've investigated), but let it as is if you have a good reason to keep them different. LGTM anyway. Thanks for looking into it.

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -71,6 +71,9 @@ export default class RaidenService {
           ...(process.env.VUE_APP_SETTLE_TIMEOUT && +process.env.VUE_APP_SETTLE_TIMEOUT
             ? { settleTimeout: +process.env.VUE_APP_SETTLE_TIMEOUT }
             : undefined),
+          ...(process.env.VUE_APP_EXPIRY_FACTOR && +process.env.VUE_APP_EXPIRY_FACTOR
+            ? { expiryFactor: +process.env.VUE_APP_EXPIRY_FACTOR }
+            : { expiryFactor: 1.1 }),
```
palango, line 76: @andrevmatos What's the best way to add validation here? I'd liek to make sure this value is >= 1.1

palango:
> Maybe we should default to 1.1, so it's also used on the CLI, I don't see a point in making them diverge, even if both should accept transfers using both factors (as well as raiden-py, as you've investigated), but let it as is if you have a good reason to keep them different.

No good reason, I changed it.

raiden-dapp/src/services/raiden-service.ts
andrevmatos: You would need a custom codec for this. I'm not sure it's worth it, since this is input by the developer.
If you wanna do it anyway, you would need the `expiryFactor` in `const RaidenConfig` to be a variable declared (inline or separete) as a `t.Type<number, number>`, `pipe`ing `t.number.decode` into a `fold` with your custom validator (e.g. `(u) => u > 1`). Check `BigNumberC` in `src/utils/types.ts` and `io-ts` docs.

palango: Yeah, the improvement doesn't seem to be worth the added complexity.

raiden-ts/src/transfers/epics/locked.ts
```diff
@@ -126,15 +127,13 @@ function makeAndSignTransfer$(
   const channel = getOpenChannel(state, { tokenNetwork, partner });
 
   assert(
-    !action.payload.expiration ||
-      // require mediated transfers to have at least confirmationBlocks before danger zone
-      action.payload.expiration >= state.blockNumber + revealTimeout + confirmationBlocks,
+    !action.payload.expiration || action.payload.expiration >= state.blockNumber + revealTimeout,
```
andrevmatos, line 130: Shouldn't this be ?
```suggestion
    !action.payload.expiration || action.payload.expiration > state.blockNumber + revealTimeout,
```

raiden-ts/tests/unit/epics/send.spec.ts
```diff
@@ -87,7 +87,7 @@ describe('send transfer', () => {
       lock: {
         secrethash,
         amount: value.add(fee),
-        expiration: expect.any(BigNumber),
+        expiration: BigNumber.from(Math.ceil(1.1 * raiden.config.revealTimeout)),
```
andrevmatos, line 90: Shouldn't current blockNumber be here somewhere?

raiden-ts/src/transfers/epics/locked.ts
palango: absolutely!

raiden-ts/tests/unit/epics/send.spec.ts
```diff
@@ -87,7 +88,9 @@ describe('send transfer', () => {
       lock: {
         secrethash,
         amount: value.add(fee),
-        expiration: expect.any(BigNumber),
+        expiration: BigNumber.from(Math.ceil(1.1 * raiden.config.revealTimeout))
+          .add(requestBlock)
+          .sub(1),
```
palango, line 93: I don't really get why I need to subtract one here.

andrevmatos: Hmm, me neither.. can you debug `requestBlock` against the current `state.blockNumber`?

andrevmatos (approved):
I'd check the reason behind that `sub(1)`, but codewise LGTM. Thanks

palango:
> I'd check the reason behind that `sub(1)`, but codewise LGTM. Thanks

So the reason is floating point math. `1.1 * 50 = 55.00000000000001`, so `Math.ceil(1.1 * 50) = 56`. I changed the PR to use `Math.round` now, but are open to other suggestions.

PR #2450, opened 2020-12-14
dapp: Display overall capacity in TransferRoute view

### LINKED ISSUE #2426: Misleading token amount shown in transfer screen
## Thanks for filing a bug report :-)

Conclusion:
We go with - **"Show only the sum of the values, if a user tries to put a higher value warn user that they can only send x."**

Possible solutions discussion
- Show minimum amount (is misleading dApp is not intended only for direct transfers).
- Can't switch between channels because of PFS.
- Show two values (max amount that can be sent in a single transfer, and the total balance).

## Steps to Reproduce
1. Have three nodes A, B, C
2. Open channel from A to B, Deposit 1 T
3. Deposit into Channel to A from B, 0.5 T
3. Open channel from B to C, deposit 0.5 T
3. Send 1 T from, A to B

## Expected Result
- B should show 2 T available

## Actual Result
- B shows 0.5 T available

A is left, B is the right dApp
Screenshot 2020-12-02 at 12 01 10

THis is the channels details page from dApp B
Screenshot 2020-12-02 at 12 02 54

palango:
This makes the displayed amount more intuitive and changes with
received/sent transfers. The amount validation in transfer view input is
still validated with the biggest channel capacity.

This also removes incorrect `noChannels` properties from
`TransferHeaders` and `TransferInputs` components.

It also includes a fix for the `channelWithBiggestCapacity` store
getter, whcih didn't order the channel correctly.

Fixes #2426

**Definition of Done**

1. Open two channels
2. Check that the transfer view displays the sum of the capacities in both channels

### DIFF
--- a/raiden-dapp/CHANGELOG.md
+++ b/raiden-dapp/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [#2415] Fix sorting latest transaction on top in transactions list
 - [#2391] Fix notification icon handling for special scenarios
 - [#2410] Fix transfer history list not showing third entry initially
+- [#2426] Display total available capacity in transfer view
 
 ### Added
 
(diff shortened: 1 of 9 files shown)


taleldayekh (approved):
Behaves and looks as expected, thank @palango 👍🏼

PR #2455, opened 2020-12-16
Feature/dapp strict manual update procedure

### LINKED ISSUE #2382: Message to update the dApp does not appear
## Steps to Reproduce
1. Access the dApp on staging
2. Merge a new PR on master

## Expected Result
- I am getting a message, that I need to click to update the dApp

## Actual Result
- I am not getting the message

```
- Browser & version
- Browser console errors
- Device / OS
```

### LINKED ISSUE #2414: dApp: extend service worker to allow offline usage
## Description
The dApp should be usable even if the server is down. Furthermore it should only request the server for updates and else just use the cached version of the application.
While our service worker claims to do this, it is clearly not working. Finally this is the result of the Chrome PWA test.
This is a big step into the direction of make this app used as PWA by Chrome and Firefox mobile (and more...).

Related to: #2382

## Acceptance criteria
- service handler caches all resources
- service handler serves app and only fetches resources if there was an update
- offline section for PWA checks run through

## Tasks
- [ ]

### LINKED ISSUE #2440: dApp - strictly manual update procedure
## Description

Follow up of #2414

At the moment the updates to the dApp happen in some kind of automated manner. The updates are tight to a new (according to byte order) server worker script. This means each time the dApp gets build, not matter if anything has changes, this means an "update". The service worker itself checks every 60s if there is an update for itself. Note that webbrowsers do so themselves every 24h automatically anyways. The check for an update of a service worker also implies (and is not devisable) the download of the new script, its parsing and installation. Then it waits in the pipe until it can displays the currently active service worker. This happens automatically when the active service worker thread got stopped (e.g. new start of web-browser, revisit domain after all related tabs where closed etc.). Or it can be manually triggered to get immediately activated. This is basically what the current update notification does. Note that this does only activate the new service worker with an updates cache, but this does not unload the active version until the next load of the dApp.

Summarized we do not strictly serve the dApp from cache once a version got downloaded. Updates are tight to service worker updates which happen uncontrolled, can't be confirmed by the user, can't be denied, get applied automatically anyways.

## Acceptance criteria
- [x] The dApp must be served 100% offline after the first initial pre-caching.
- [x] Update for a new version of the dApp must be independent form the service worker update procedure.
- [x] The user must be notified of a new version before it gets fetched.
- [x] An update gets only applied when a user actively confirms it.
- [x] A confirmed update gets applied actively, means it requires to shutdown the SDK, apply update and load new version.
- [x] The shutdown of the SDK **must not** remove the local state including the key (currently sleeping bug in code)
- [x] The service worker must be able to detect that a user cleared the browser cache, which should result into a fresh new version.
- [x] The service worker must react to a (partially) deleted or corrupted cache during the runtime of the dApp.

## Tasks
- [ ] 

---

Note that this issue got created a while after there was already a massive amount of work and research invested into this base "feature", until it has been decided to make this its own issue. Therefor it includes more detailed criteria.

weilbith:
> This is an epic PR. Not because it is contains above average changes of code, but because it is something quite new to all (?) members of the team and contains a lot of new information. Grab a cup of tea or coffee and reserve some time. Make sure you are not stressed already and take breaks from time to time to stay at 100% brain capacity.
> PS: Sorry for all the spelling and grammar mistakes plus awkward sentences. It is a lot and I don't want to spend iterations of reading over and improve. I hope you can still understand anything, else don't hesitate to ask.

Fixes #2440
Fixes #2414
Fixes #2382
Successor of #2439

**~~Short~~ Long description**

Alright. This will take some time to write/read. The PR is of quite some importance for the mainnet release and is necessary for some legal requirements. Let's wrap up those requirements, followed by some basic knowledge facts regarding service workers and then some details to the solution approach. I try to not repeat myself with the descriptions within the commit messages. This whole feature took me quite some time and I want to make sure each reviewer is prepared as good as I can do.

<details>

For legal issues in terms of being the provider of the dApp it has been decided that the application should run as much independent from the web-server as possible. Also it is possible to build the dApp from source all the time, this comes with some downsides, especially convenience restrictions for the user. Therefore we attempt to make the dApp a progressive web app. But in a more strict manner. While such apps support offline usage in one or the other way, they usually always re-connect to the web-browser to sync data and get updates. A typical example would be a lost mobile network connection in the subway, but keeping the application continuously working. For us it is different in terms of that the application should be available fully locally once loaded the first time. From then on the server is not contacted to get automatic updates of the local version nor any other data exchange. The web-server therefore just provides a already build bundle of the application to the web for convenience. The application needs the network only to connect to the Ethereum Node and the Matrix servers. Though it should be still possible to check if there is an update of the application available on the web-server. The user should be notified of such an optional update. He must explicitly confirm that he wants to update. Only then web-server gets contacted again to download the new assets.

The heart of progressive web apps and especially the offline usage is the so called service worker. The service worker is (oh wonder) a worker of the web-browser. A worker is a separate thread related to a domain. This means he is not part of the main thread that controls the page/application itself. This means this thread has no access to the window (object), document, local storage, cookies, ... What it has access to is the cache API (in terms of data storage). While it also has some capabilities of pushing notifications, its major functionality comes with the fact that it can interfere network connections. This means each time the main thread access for example the `fetch` function of the web-browser to load some data, the service worker can hook in. Therefore it can listen to all requests a page does and if a request matches any kind of conditions (could be also just all), it can answer itself with a response object. The main thread does not see where the response comes from. Here comes a typical example of a service worker cache functionality. The service worker listens to all requests the client does and filters for example images. If it has never "seen" this image before, it will continue to forward the message to the server to get the file. Then it saves the file into the cache of the domain (which is like a small database) and respond to the client with the response including the file. The next time the user queries the same image, the service worker immediately respond with the file from the cache instead of connecting to the server. This saves for example band-width and makes the page faster. Following this concept allows the application to provide some offline functionality. This works in both directions. Means it can also also cache request of the client like saving a document until network is available again to then do all the save requests.
Further to note is that it is possible to exchange message events between the main thread and a worker. This is of importance for any kind of functionality that can't be simply solved by a fetch response.
The for us interesting feature branch is the so called pre-caching. This means we don't cache only after the client requested a resource the first time. But we maintain a list of assets that will be get all downloaded at once when the service worker starts. So the worker itself queries the server for all these assets and stores them to the cache. Depending on the content of the defined list of assets, this makes the application super fast and responsive, as long as it does not do other network requests that are not cached/cachable. Note that this happens in background. So it will not delay the initial page load of the user. When the assets are available that got requested initially, this works like a normal load. All future resources are then already available later when needed. For us this is the key to generate a list of all assets our application consists of. All the dozens of javascript, css, html chunks, font typesets, images, ... All data gets stored into the cache. And each possible asset request the page can do is available offline on local host. So an icon for a new notification that was never displayed before is already available all the time. The only remaining network usage comes from the communication with the Ethereum node and the Matrix servers.

The important part which causes all the "trouble" for us is, that we must prevent automatic updates of the assets in the cache. Once a version got pre-cached it must stay as it is. This is not against, but pretty uncommon for such applications, which means we can't use existing solutions (already not out of the box). The service worker is meant to support the main thread for temporally network outages and make the application faster in background. But not to cut off the web-server connection at all. Let's see how I can put in the facts. I guess first it is necessary to know how a service workers lifecycle looks like. The service worker gets registered by the main thread. This means it calls a browser API with the URL to the service worker script (not like usual JS that gets loaded via HTML tags). The browser then downloads the file, install it and then activate it. When a service worker got activated it can start to inject the requests of the main thread. You also say that the worker now controls the client. The installation phase is like the preparation part. The thread of the service worker runs in parallel to the main thread (obviously). The next time the page gets loaded the "old" thread gets restarted in activated mode already. Means the install and activation steps are not executed again. But the page reload means that the service worker registration gets executed again. This means the browser downloads the script file again. Now it does a byte-by-byte comparison to the currently active service worker. If it is the same, it dismiss it. If it differs, it will install the new service worker, **but** does not activate it. The old service worker is still active. This is made to prevent that the old worker gets killed while still doing important stuff like caching offline save requests for the client or anything else. This means there are now two threads (plus the main one), but only the active one controls the applications. When the thread got stopped for example when the browser got closed and then the next time the page gets loaded, it will recognize this situation, remove the old active worker and activate the new worker. This step can be fasten by "skipping" the waiting phase of the newly installed service worker. This is a function call that can be done while browser and the page are actively running. The service worker script file can be cached as every other resource. Note that this is a different cache than the one the service worker controls. This is the global cache of the browser where it caches stuff according to the time-to-life properties defined by the web-server. **But** service worker files are treated special. This means after 24h, no matter which `ttl` value the server has set, the browsers ignores the cache automatically and query the server for the most recent version.
Alright. Let's come closer to our issue. Important to know is that the pre-caching is happening in the install and activate phase. Here the assets gets downloaded, put into the cache, old file versions get cleared and then once activated it will service the client from cache. The list of assets the worker should pre-cache is packed into the script. For us this means when webpack builds our application it generates a list of all the chunks it generates with all the versions and then inject this list into the script of the service worker. This means each new build of a new version will result into a new service worker. Even "worse" each compilation of the service worker source script (e.g. in TypeScript) will lead to a output script that differ in bytes to the successor. I don't know the details (maybe some date injection?!), but that is the case for our setup. Anyways this isn't too important. Important is that if we have a service worker that pre-caches all assets of the application when it gets installed and activated, will install the most recent version of the application. And since we are out of control of when the web-browser gets the most recent version of the script from the server, we get automatic updates. Also if we do not skip the waiting service worker, the next day, the next time the page gets loaded the user could end up with the new version. As long as we haven't published/build a new version on our web-server, all this doesn't matter. No matter how uncontrolled the browser fetches the most recent version, the file is 100% the same and the "old" service worker lives for ever as long as it does not manually get killed. But if the user loaded the application, then we release a new version, then this becomes an issue.
The conclusion of that is that we need to decouple the updates of the service worker that happens "uncontrolled" to controlled updates of the cached version. So no matter how often the service worker gets updated because we build/released new version, the cached assets remain as they were loaded the first time. So each newly installed service worker fist checks if there is a local version available already. If not, like for the first load, it installs/pre-caches the whole application according to what the web-server serves atm. But if the cache is already filled, means it was installed in the past already, we skip that phase. In the active phase of the service worker all behave the same: they just serve from cache.
An important injection is the following. Although a service worker skips the pre-caching part and re-uses the existing one, this still doesn't work. The worker is using a route to match the requests of the client. This way it is still possible to access other parts on the web-server (e.g. the docs) that are not in the control scope of the service worker. But the asset list is part of the service worker script as explained before. This means we would end up with a new service worker that uses an old cache, but matches requests against a new list that does not link to the cache storage. Therefore we need to persist this asset list after a successful (new) pre-caching. This then allows any new worker to correctly replace the old service worker while serving the local version. Since the cache storage does allow to save response objects one, the persistence of the asset list must be done in the indexdb database of the web-browser.
All together this makes the application stably served "offline". So now it needs to establish a update mechanism again. Therefore the server must provide a way to check for the version he provides. This can be compared to the local version to determine if there is an update available. An actively running service worker can't do the update of the cache (because it was compiled with the old list of assets). So it will require to install the new version of the service worker. This requires the current service worker to tell the next service worker to do the update. The only medium they have access to to persist information is the cache itself. Therefore it as simple as the current service worker deletes the whole cache (plus the asset list in the database), it unloads itself, so that on the reload of the page the web-browser must install the current version of the worker script. That will find an empty cache will automatically lead it to pre-cache the current version. Done.
The now remaining critical section is that we need to make sure that the cache is working. If the cache gets lost or files are missing, we have an issue. And since there is no update available, the user has no option escape this scenario (if he don't want to access the developer tools). He is stuck on a broken local version that does not get updated. This can for example happen when he (or maybe in automated manner) clears the data/cache of the web-browser. A partially broken cache should never happen, also we need to handle this case. We need to detect this scenario and introduce mandatory "updates". While before the user had the option to stick with the current local version and continue using the app, he now **must** update to get a working app again. This might be the same version as before if there was no new release on the web-server or a new version.

Alright. All the implementation details and some smaller issues and facts are within the commit messages. They are quite exhaustive, even for my standards. :see_no_evil:

Unfortunately it is not possible or just very hard (and we don't have the resources) to test all these features in an automated end-to-end environment. It requires bumping of the package version, re-builds while tests are running, full control over the browsers threads etc. Ofc I wrote all the unit tests where possible. But the vast majority of what got implemented here must be tested manually. I did that hundred times locally in different web-browsers. I tried to break it as much as I can (and I did...) to verify that it is as stable as possible. So below I'll put a longer list of scenarios to run. It is time to grab the next cup of coffee (maybe tea if you are still motivated). Please first read all the commits one-by-one. I re-implemented everything (well I did so dozen iterations) trying to do it in meaningful and coherent chunks bundled as commits. But be aware that the history might not be 100% straight forward. In early commits I sometimes structure code already in shapes that might look unintuitive first, but I did so with knowing the "future" already. So add your questions and comments for each commit. Then in the end display all commits at once and go through your comments again to see if they got resolved now.
</details>

**Definition of Done**

Note that the tests are meant to run after each other.

**"Offline" usage**
<details>

- build the dApp from source `yarn build`
- serve it locally (e.g. `npx serve ./dist`)
- open the app in your browser
- stop the server again (after app fully loaded)
- make sure the app is working by connecting, clicking around, open menus, ...

</details> 

**Re-build stable**

<details> 

- re-build the dApp again `yarn build` (don't change anything)
- serve it locally
- go to app in browser, wait a while (at least 1min), reload it, restart browser, wait 24h (just kidding...), repeat the checks for all these steps as described below
- verify that there is no pop-up for an update
- verify that the cache is still the old one
  - chrome based: developer tools -> application -> Cache Storage (left side) -> expand and select precache of app -> see dates in list of cache
  - firefox: developer tools -> storage -> Cache Storage (left side) -> expand and select precache of app -> see dates in list of cache
- verify that the service worker is still the old one (registration date of worker) [in dev-tools -> application]
- verify that all is still as it was before

</details> 

**New Version**

<details> 

- bump the version of the package (just increase it to anything in the `package.json`)
- re-build dApp (feel free to include any other changes to the sources to cause an actual new version you can see)
- service it
- go to the browser
- wait (~~max 1min~~, updated due to review, needs to wait 1h now or reload the dApp) for the update snackbar to appear
- make sure the app is still working, everything is clickable
- restart the dApp
- verify that the update snackbar appears again
- accept disclaimer and connect (just for some extended test)
- verify that your changes to the application (if you did) are not there, still
  the old version
- eventually re-check the date of the cache if you don't remember
- accept the update
- make sure this "disconnects" (so you navigate back to home)
- verify that the page reloads
- check the cache and the service worker dates to verify that it was updated
- connect the app and watchout for you changes (of you introduced any)
- if you are motivated you can build another new version (should offer new optional update) or also try to reset to the old version (should do nothing in app)

</details> 

**Clear Cache**

<details> 

- have the dApp open in browser and make sure the server is running
- you need to be fast now!
- go to the cache storage in the developer tools and clear it
- within 10seconds click on any link in the application
- the screen should block blurred and the snackbar informs about a mandatory update
- do the update
- clear the cache again
- wait without doing anything (max 10s)
- verify that the you need to do a mandatory update again
- do the update
- clear the cache again
- close the page and load it again
- the page should be empty with a standard error of the web-browser
- reload the page/click on the button of the error
- verify that the application loads again

</details> 

**Stable Cache Check / Worker Switch**

<details> 

- have the dApp open in the browser and unregister the service worker
- load the page again
- make sure the service worker got replaced now
- make sure you don't get any broken cache updates or similar
- unregister the service worker again
- build a new version of the dApp
- consider to remove an entry of the `routes.ts` (which will result into a shorter/different asset list
- service the new version and update within the dApp
- make sure you get no mandatory update due to broken cache
- try to access the route you have removed and make sure this is still possible

</details> 

**Install PWA**

<details> 

- this requires a chrome based browser like Chromium, Chrome, Opera, Edge, Brave, ..
- verify that there is a banner at the top that advertises to install the app
- click install and accept the pop-up of the browser
- the app should be loaded in a standalone window colored like Raiden (light blue)
- close this window and also the browser (optional)
- check your desktop or use a smart application launcher and watch-out/search for the Raiden dApp
- make sure you can start this app
- feel free to uninstall it (title bar of app window, context menu/three dots, uninstall)

</details>


weilbith:
While writing the description for the PR I discovered a bug. I'll tackle that now. It is still okay to have this open here as it will be just an addition to it. And since the review will probably take some time, it just save some moments while I'm working on the fix.

weilbith:
Alright, I finally fixed the issue I discovered. I'll update the PR description now. More details as always in the commit messages.

christianbrb:
@andrevmatos -> @palango and @taleldayekh are going to review this one

raiden-dapp/vue.config.js
```diff
@@ -89,12 +93,14 @@ module.exports = {
...
     }
-  },
-  pwa: {
-    workboxPluginMode: 'InjectManifest',
-    workboxOptions: {
-      swSrc: './src/sw.js',
-      swDest: 'service-worker.js',
-    },
+
+    if (process.env.NODE_ENV === 'production') {
```
palango, line 106: Mh, should we maybe enable this on staging as well for testing purposes?

raiden-dapp/src/plugins/service-worker-assistant-plugin.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * @param _Vue - global Vue instance to act on
+ * @param _options - eventual configuration for the plugin (ignored)
+ */
+export async function ServiceWorkerAssistantPlugin(
+  _Vue: typeof Vue,
+  _options?: null,
+): Promise<void> {
+  const serviceWorkerAreSupported = 'serviceWorker' in navigator;
```
palango, line 27: This should either be `serviceWorkerIsSupported` or `serviceWorkersAreSupported`

raiden-dapp/src/plugins/service-worker-assistant-plugin.ts
```diff
@@ -0,0 +1,20 @@
...
+  const serviceWorkerShouldBeRegistered = process.env.NODE_ENV === 'production';
```
palango, line 28: I commented on a similar thing in another commit already. Can we somehow centralize this condition, so we have one place in the codebase for this?

raiden-dapp/vue.config.js
weilbith: Well, you can do this manually if you really need it. Though I don't recommend it. 
For staging this does not work as this would prevent updates. Staging version are only between to package versions/releases. This means once you loaded the first time a staging version after a new release, this will be sticky until the next release has passed. Means you won't see any other staging version in-between.
For testing I think we can't. Unit tests don't have this environment available. And the end-to-end tests in Cypress lack in control of the environment. Means we don't get a clean setup. I somewhere documented. Though we can (and probably should) try to get it working in the CI, hoping that the cache we have (yes includes Cypress) does not include such stuff. @taleldayekh how did you solved the localStorage issue back then? We once had the issue that Cypress does not clear the data between test runs.

raiden-dapp/public/manifest.json
```diff
@@ -1,6 +1,12 @@
 {
-  "name": "raiden-dapp",
-  "short_name": "raiden-dapp",
+  "name": "Raiden Light Client dApp",
```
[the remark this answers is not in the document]
weilbith: This file is always updated manually. That's why it is part of the repository. Nothing generated should be here. 
We edit this when ever we want to. If you think some values could be better just go for it. I was just following some recommendations and what I thought fits best. I'm open for any kind of suggestions.

raiden-dapp/vue.config.js
taleldayekh: It was solved like this: https://github.com/raiden-network/light-client/blob/eb712da309d6c2ed22f93cf657121d53b08bbba2/raiden-dapp/tests/e2e/specs/scenario.spec.ts#L54

raiden-dapp/src/service-worker/worker/index.js
```diff
@@ -0,0 +1,31 @@
+import { PrecacheController, PrecacheRoute } from 'workbox-precaching';
+import { registerRoute } from 'workbox-routing';
+import { isCacheEmpty } from './utilities';
+
+async function onInstall(event) {
+  this.initialCacheWasEmpty = await isCacheEmpty();
+
+  // Do not update if there is still a(n old) version cached.
```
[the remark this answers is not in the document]
weilbith: This comment does not exist anymore on the most recent version.

raiden-dapp/src/plugins/service-worker-assistant-plugin.ts
weilbith: It is the same line of code...

raiden-dapp/public/manifest.json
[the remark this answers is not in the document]
taleldayekh: Ok, I though it was updated automatically before. Then all good. 👍🏼

raiden-dapp/vue.config.js
weilbith: So we eventually could do this for the Cache API and IndexedDB as well.  Though this will not test the service worker its features, but just makes sure nothing is broken or unexpected updates are proposed. The Cypress environment is not able to test the update procedure. At least not with a feasible amount of work (I discussed this with Christian).
I would love to have a talk with us 3 to evaluate what we could do best for making sure nothing gets broken over time. Apparently I forgot many things. But the right questions should help.

raiden-dapp/vue.config.js
```diff
@@ -1,10 +1,19 @@
...
 const CopyWebpackPlugin = require('copy-webpack-plugin');
+const VersionFile = require('webpack-version-file-plugin');
+const { DefinePlugin } = require('webpack'); // eslint-disable-line import/no-extraneous-dependencies
 const { InjectManifest } = require('workbox-webpack-plugin');
 
 const sourceDirectoryPath = path.resolve(__dirname, 'src');
 const distributionDirectoryPath = path.resolve(__dirname, 'dist');
 
+function getPackageVersion() {
+  const packageFilePath = path.resolve(__dirname, 'package.json');
```
palango, line 12: Maybe import the JSON instead of reading the file explicitly?
```js
var pjson = require('./package.json');
console.log(pjson.version);

File: raiden-dapp/src/service-worker/assistant.ts [hunk 7]
```diff
@@ -1,15 +1,34 @@
...
   }
 
+  public update = (): void => {
+    navigator.serviceWorker.controller?.postMessage(ServiceWorkerAssistantMessages.UPDATE);
+  };
+
+  private onMessage = (event: MessageEvent): void => {
+    if (!event.data) return;
+
+    switch (event.data) {
```
taleldayekh, line 23: Out of curiosity, why did you decide to go with a `switch` statement here when you are not really switching between things? Or maybe you are adding more stuff in a later commit?

raiden-dapp/src/service-worker/assistant.ts
```diff
@@ -0,0 +1,28 @@
+/* istanbul ignore file */
+import { Store } from 'vuex';
+import compareVersions from 'compare-versions';
+import { CombinedStoreState } from '@/store/index';
+
+const VERSION_FILE_PATH = (process.env.BASE_URL ?? '/') + 'version.json';
+
+export default class ServiceWorkerAssistant {
+  constructor(private store: Store<CombinedStoreState>) {
+    setInterval(this.updateAvailableVersion, 1000 * 60);
```
palango, line 12: Once per minute seems quite often IMHO.

raiden-dapp/src/components/UpdateSnackbar.vue
```diff
@@ -13,39 +12,32 @@
 <script lang="ts">
 /* istanbul ignore file */
 import { Component, Vue } from 'vue-property-decorator';
+import { mapGetters } from 'vuex';
 import BlurredOverlay from '@/components/overlays/BlurredOverlay.vue';
```
palango, line 19: Is this still needed after the changes?

raiden-dapp/src/components/UpdateSnackbar.vue
```diff
@@ -13,39 +12,32 @@
...
-    });
-  }
-  beforeDestroy() {
-    window.removeEventListener('swUpdated', this.handleSWUpdate);
-  }
-
-  handleSWUpdate(event: ServiceWorkerUpdatedEvent) {
-    this.swRegistration = event.detail;
-    this.updateAvailable = true;
+  get visible(): boolean {
```
palango, line 35: What's the reasoning to define `visible`, when you could just use `versionUpdateAvailable`?

raiden-dapp/src/service-worker/assistant.ts
weilbith: Because I know the future. :wink: 
Also was this the code approach before this PR and the MDN suggest this approach. Also all these arguments are really weak. In first place because I knew there are more messages to come.

raiden-dapp/src/service-worker/assistant.ts
```diff
@@ -25,4 +44,9 @@ export default class ServiceWorkerAssistant {
       console.warn(`Failed to get (a valid) version: ${error.message}`); // eslint-disable-line no-console
     }
   };
+
+  private reloadWindow = (): void => {
+    // Delay reload so that the service worker can unregister safely if necessary.
+    setTimeout(() => window.location.reload(), 1000);
```
palango, line 60: Maybe mention why you chose a one second delay here.

raiden-dapp/package.json
```diff
@@ -37,7 +37,9 @@
     "vuetify": "^2.3.21",
```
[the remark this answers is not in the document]
weilbith: Unfortunately I can't see which commit you mean.  Could you quickly link the has here? Thanks.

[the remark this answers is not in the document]
palango: Ah damn, didn't think about that: 39164672ce0ecb05d89c2e8aad0ca830aea31626

raiden-dapp/src/components/UpdateSnackbar.vue
palango: Fixed by later commit

raiden-dapp/src/components/UpdateSnackbar.vue
palango: same here

raiden-dapp/src/service-worker/messages.ts
```diff
@@ -6,6 +6,7 @@
 
 export enum ServiceWorkerMessages {
   RELOAD_WINDOW = 'reload_window',
+  CACHE_IS_INVALID = 'case_is_invalid',
```
palango, line 10: s/case/cache

raiden-dapp/src/service-worker/assistant.ts
weilbith: That was the former used value. I just continued to use it. I'm totally open for any value you guys like to. I have no strong feelings about it. Please make a suggestion.
Though 1min is useful for testing. Maybe making it configurable would be helpful. The question would be how. Any suggestions?

raiden-dapp/src/components/UpdateSnackbar.vue
weilbith: Yes ofc? What do you mean? Why shouldn't it? Else the linter would complain btw...

weilbith: I fail to get a permanlink to the line. But it is just a couple of lines below in the decorator of the class.

raiden-dapp/src/service-worker/worker/utilities.js
```diff
@@ -2,6 +2,16 @@
 import { cacheNames } from 'workbox-core';
 import { ServiceWorkerMessages } from '../messages';
 
+/**
+ * Checks if the service worker is contralling any client.
```
palango: nit: s/contralling/controlling

raiden-dapp/src/components/InstallBanner.vue
```diff
@@ -0,0 +1,49 @@
...
+      </v-btn>
+    </template>
+  </v-banner>
+</template>
+
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator';
+
+@Component
+export default class Name extends Vue {
```
taleldayekh, line 27: Should this really be called `Name`?

raiden-dapp/src/components/UpdateSnackbar.vue
weilbith: Because at a later point in time I need to extend this for mandatory updates as well.
As I wrote I knew the future of the code. Therefore things sometimes look slightly off in some of the commits. I'm sorry for that. But I barely saw the advantage to optimize each commit as it would be the final one if I already knew what makes sense in terms of the following commits/features.

raiden-dapp/src/locales/en.json
```diff
@@ -734,5 +734,11 @@
       "blockie-alt": "Token address blockie",
       "information-template": "{symbol} | {name}"
     }
+  },
+  "install-banner": {
+    "install": "Install",
+    "dismiss": "Dismiss",
+    "message-mobile": "Add the Raiden dApp to your Home screen",
+    "message-desktop": "Install the Raiden dApp to your desktop"
```
taleldayekh, line 742: Replace `to` with `on`, "on your desktop" :)

raiden-dapp/src/service-worker/assistant.ts
weilbith: Okay. So I would write it here and you can argue if that make sense to you:
1s should be very long from the execution time as the unregistration of the service worker is a direct follow up. In terms anything else is taking the CPU 1s should be still enough so the server worker gets the CPU again to unsubscribe. But 1s is short enough that the user feels not like the app is laggy. As this is an update he should  be fine that it is not instant. 

Now thinking about it we could improve it and wait until no service worker is controlling the client anymore (like for the check if there is already a controller during registration). That would make it bullet-proof. 

What do you think?

raiden-dapp/src/service-worker/messages.ts
weilbith: :see_no_evil:

raiden-dapp/src/components/InstallBanner.vue
weilbith: Uuups. Good catch. That is a template from my template to create a Vue component file. xD
I'll fix that!

raiden-dapp/package.json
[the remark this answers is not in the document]
weilbith: [Here is the mentioned issue](https://github.com/microsoft/TypeScript/issues/20595).
Is it enough to have it in the conversation here? Or where can we put it else? :thinking:

raiden-dapp/src/components/InstallBanner.vue
taleldayekh: I'm surprised it worked to import it anyways.

raiden-dapp/package.json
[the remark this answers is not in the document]
palango: I think having the link here is enough. Thanks!

raiden-dapp/src/components/InstallBanner.vue
weilbith: Why should it? We just import modules as their default export. So `import lalaStupidWhatever from 'serious-library'`. Just works. Remember that `import` statements are just like syntactic suggar. The `require()` function is there maybe more obvious as it is a function call and you store the result into a constant variable. Imports like `import { foo, bar } from 'foobar-module'` are just spreading the returned object. So a module that exports multiple things is just a function that return an object. So the import of `'foobar-module'` returns something like `{ foo: something, bar: something, more: something, ... }`. And we just disassamble it. If we would now use a property that does not exist on this objects (means there is no export with this name), we get an error. But for the default export this is just a new variable we create and we can name it like how ever we want. The name of the default export does not matter.

taleldayekh: Ok, then it makes sense.

raiden-dapp/src/service-worker/worker/index.js
```diff
@@ -0,0 +1,96 @@
...
+}
+
+async function onRouteError() {
+  if (await isCacheInvalid.call(this)) {
+    if (await isAnyClientAvailable.call(this)) {
+      sendMessageToClients.call(this, ServiceWorkerMessages.CACHE_IS_INVALID);
+    } else {
+      update.call(this);
+      // Note that the user now has to reload the page himself. Since there is
+      // no client available, nobody can reload the page automatically.
```
palango, line 84: Can we display a message in this case?

palango:
Finished a first pass now. With the detailed descriptions and the presentation it was quite easy to get through! Thank you!

So far I only found small typos. However, codecov complains a lot about the following files. Is there any chance to test the (better) without complete mocking?
- `InstallBanner.vue `
- `service-worker-assistant-plugin.ts`
- `service-worker/assistant.ts`

raiden-dapp/src/service-worker/worker/index.js
weilbith: How?
This is basically what the comment states. There is no client. So no window, no document, no nothing.
But maybe I'm overseeing something. :shrug:

raiden-dapp/src/plugins/service-worker-assistant-plugin.ts
weilbith: :heavy_check_mark:

raiden-dapp/vue.config.js
weilbith: :heavy_check_mark:

raiden-dapp/src/service-worker/messages.ts
weilbith: :heavy_check_mark:

raiden-dapp/src/service-worker/worker/utilities.js
weilbith: :heavy_check_mark:

raiden-dapp/src/components/InstallBanner.vue
weilbith: :heavy_check_mark:

raiden-dapp/src/service-worker/assistant.ts
weilbith: I'm implementing the smarter and safe solution now and test it. Thanks for pushing me for a better solution!

weilbith: Even better. I did some wrong assumptions. Unregistering a service worker makes it only switching into the redundant state. But as long as there is now new service worker in the pipe that got installed and can be activated, the redundant worker is still the controlling one. Means we can still exchange messages with the client(s). Therefore we can savly first unregister the service worker itself and then notify the client to reload the window. To be 100% sure I tested this with an timeout of 20s between the instructions of the service worker to unregister and sending the message. Finally the assistant must not wait for anything with a wild guess timeout or check any conditions. Just reload the window right away.
Checkout the commit for details (even i think this comment might be even longer...)

weilbith: I'm closing this optimistically as I think I have successfully targeted your concerns in the best way possible.

raiden-dapp/src/service-worker/assistant.ts
weilbith: We have decided to change this to having an initial version check when the page loads and then every hour only. Going to implement this.

weilbith: Implemented. I adopted the test description of the PR for this.
As you have agreed with me on this solution in RC, I'm going to close this conversation since the implementation was trivial.

palango:
All tests besides "Stable Cache Check / Worker Switch" passed for me.

For "Stable Cache Check / Worker Switch" I saw the following behavior:
1. Running the dApp and unregistering the service worker works fine.
2. I built a new dApp version with the `notifications` route disabled. After reloading the page the cache is the same (see screenshots), but for some reason I get stuck when I click on the notifications icon.
3. After another reload the notifications screen works again, while the cache is still at the old version.


### dApp cache before step 2
Screen Shot 2021-01-12 at 15 35 10

### dApp state during step 2, and the console output
Screen Shot 2021-01-12 at 15 37 31
Screen Shot 2021-01-12 at 15 37 46

### dApp state in step 3
Screen Shot 2021-01-12 at 15 38 13
Screen Shot 2021-01-12 at 15 38 21

raiden-dapp/src/service-worker/worker/index.js
palango: This was just a verification. If we cannot do something we cannot.

weilbith:
@palango thanks for the feedback. I'll see if I can figure out what is going wrong there. At the moment I'm only speculating. Really nice that you caught.

weilbith:
Alright. The as discovered by Paul got fixed. Besides also some other smallish improvements, usually put as fixup commits.
Now we will await Talel his final review (probably he is back tomorrow). Then we will have a talk about the testability. Therefore I'll investigate how/if we can mock the navigator in the test environment with Jest.

weilbith:
Please be warned that I'll do a small rebase a couple of commits back to correct the CircleCI fix. I can't do it as a fixup commit as the message is wrong as well.

weilbith:
In a meeting we agreed that this PR should come to an end. This has been tested exhaustively manually plus we have some basic unit tests. Though it will be the target to increase the tests before the next major release. But all unit tests we write will only ensure that the current implementation remains fixed. It won't take the fear that we do not see some special case where the user could get stuck. We need to continue investigation how to make this as secure as possible. Follow up issues are about to be created.

---

I'll start to resolve the merge conflicts now.

weilbith:
Sorry that it took so long. I ran into some issues during my first attempt. Let's hope CI is passing. :see_no_evil: 
Note to myself: fixup commits are great, but when they are too far in the future of the to fix commit, they become confusing as the diff is different.

raiden-dapp/vue.config.js
weilbith: See [this comment](https://github.com/raiden-network/light-client/pull/2455#issuecomment-765267101).

weilbith:
I'm temporally enabling the nigthlies pipeline here. Just to make everything is alright and there is no evil awakening once merged.

weilbith:
Alright, [the nightlies](https://app.circleci.com/pipelines/github/raiden-network/light-client/8258/workflows/8c973e25-2f45-4f9e-83af-eb8436162e57) where successful. But the default workflow failed due to the unit tests of the SDK. This flakiness is off target this PR. So I'll just remove the nightlies again.

Talel told me that he has no time to do a final review due to higher priority tasks. Therefore I'll asking Paul instead.

palango:
New commit look good. Thanks for adding the tests and staying with this PR in general!

Just one little question with the version number, otherwise we can merge this from my side 🥳

raiden-dapp/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "raiden-dapp",
-  "version": "0.14.0",
+  "version": "0.14.1",
```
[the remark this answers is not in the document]
weilbith: Oh sh**. Thanks! :pray:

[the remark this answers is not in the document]
weilbith: :heavy_check_mark:

palango (approved):
🏅

weilbith:
Thanks to Paul he remembered that there were some spelling mistakes colleagues have mentioned during the demo. I just added them as fixup commits and will integrate them now.

weilbith:
Alright. Everything is done. The remaining CodeCov hook will be targeted later if possible as already discussed earlier. Let's merge this. :crossed_fingers:

PR #2471, opened 2020-12-29
Bugfix/dapp remember token during navigation

### LINKED ISSUE #2430: Token switched when leaving 'channels' screen
## Steps to Reproduce
1. Select a token that's not first in the token list
2. Go to the channels list
3. Click the 'back' button

## Expected Result
- Should be back on overview page for the selected token

## Actual Result
- The first token from the token list is shown in the overview page

```
- Browser & version
- Browser console errors
- Device / OS
```

weilbith:
Fixes #2430 

**Short description**
This just tweaks the back button navigation function to remember the selected token.
While doing this I recognized that sometimes we still assume the home route it the home route. But unfortunately it is the transfer route. This just works as the navigation guard automatically forwards the user if he is already connected. This should not be the way to go for. After all this is just another indicator that we should finally rename our home and transfer route to better reflect their purpose.

**Definition of Done**

1. Go to the transfer route
2. Select a different token (not the first one!)
3. Go to the channel list view of the token
4. Navigate back using the back button of the header
5. Verify that the same token is still selected, not the automatic first one (you should also see the token as parameter in the browser address bar)
6. Go to connect a new token
7. Navigate back as before and repeat the verification

### DIFF
--- a/raiden-dapp/CHANGELOG.md
+++ b/raiden-dapp/CHANGELOG.md
@@ -10,6 +10,7 @@
 - [#2426] Display total available capacity in transfer view
 - [#2431] Disable UDC 'withdraw' button when no eth is is raiden account
 - [#2476] Fix persitence to remember disclaimer acceptance if selected
+- [#2430] Remember token selection during navigation
 
 ### Added
 
@@ -33,6 +34,7 @@
 [#2431]: https://github.com/raiden-network/light-client/issues/2431
 [#2476]: https://github.com/raiden-network/light-client/issues/2476
 [#2448]: https://github.com/raiden-network/light-client/issues/2448
+[#2430]: https://github.com/raiden-network/light-client/issues/2430
 
 ## [0.14.0] - 2020-11-25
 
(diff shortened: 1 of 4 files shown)


agatsoh:
@weilbith there is a `raiden-dapp/src/mixins/navigation-mixin.ts`  codecov warning. Is it possible to look into it 
navigation_mixin_codecov_warning

weilbith:
> @weilbith there is a raiden-dapp/src/mixins/navigation-mixin.ts codecov warning. Is it possible to look into it.

Thanks for the hint. 
I argue for not adding tests here. This is just a helper function for better to read code in the components. The unit tests for the components do have tests for the expected navigations. But they test it on a higher level, means mocking the injected router. Thereby this Mixins is just the bridge that doesn't get tested here which is fine as it does include any business logic by itself. Each failure of it would be visible in the unit tests. And here it is only about the default case.
Thereby I'll use my administrator permissions to merge this anyways.

agatsoh (approved):
looks good to me. Also tested on my local.

weilbith:
I'm rebasing to resolve the merge conflict.

PR #2453, opened 2020-12-16
Send messages preemptively and use RTC channel as presence complement

andrevmatos:
Fixes #2303 
Fixes #2352 
Fixes #2433 
Fixes #2443 

**~Short~ Description**
This ended up being larger than expected, since it involved a big transport refactoring and a lot of server and client-side debugging. Fortunately, the end result is very good from my perspective, and adds a lot of performance improvements and preemptive bugfixes:
- Presence bug is fixed: we call `matrix.setPresence` explicitly after a random time after start, which causes a new presence update event to be emitted to all clients and federated servers, refreshing our presence for the rest of the session
  - A similar call is issued whenever capabilities change, since this was the only way to trigger peers to re-fetch our profileInfo and get the up-to-date avatarUrl
- WebRTC signaling messages are send using the standard messaging epics, instead of calling the helpers directly
- Instead of the complex waiting logic we had before sending a message, now messages are always sent as soon as possible, and instead, we hold the `messageSend.success` action, which was waited for before retry loops, until some condition is met which tells us with a higher confidence the message was received, e.g. peer is online
  - Send the messages early allows it to work even if the peer appeared offline, since they would get the message anyway
- The points above allows WebRTC signaling messages and channels to be established even if peers appears offline to each other, while "waiting" for the same conditions as message success action before retrying; the result is that the channels can be now established "asynchronously" (e.g. offer sent while callee was still offline) and while peers look offline to each other
  - Because RTC channel negotiation doesn't depend on the matrix presence anymore, it becomes an additional source of truth for presences: if an RTC channel is established, we consider partner as `online` even if they look offline in matrix
  - Also, now we avoid creating RTC channels with any node we fetch presence for, and instead do it only for neighbors and target/initiator on transfer sends. Hopefully, in the future, we can also remove the later.
- When using rooms, only the lower address creates the room and invite: this streamlines this behavior with PC. While our previous did work fine, it was more prone to races and could take longer to establish a reliable on-matrix communication room.
- Batch and serialize per-address p2p messages and per-room global messages:
  - This can reduce a lot the number of parallel matrix requests, which have shown to bottleneck the transport
  - This also reduces the number of API requests needed, since batching can send multiple messages in a single request
  - Once we get rid of the rooms code and messaging, it can be made even more efficient, since `sendToDevice` can send messages of the same type to multiple recipients in a single API request
- Adds better retries, batchSize and timing data to `messageSend.success` actions and add log calls with these info after a `messageGlobalSend` request is completed.
[...]

PR #2482, opened 2021-01-04
Transport and types improvements

andrevmatos:
Some commits offloaded from #2453 

**Short description**
- Ensure `deviceId` is set also for the fallback `matrix.register` call (preparation for enforcing it on `sendToDeviceMessage`
- Consider a peer with an established `RTC` channel as `online` complementary to matrix's presence
- `setPresence` upon capability changes to ensure peer's presence update is triggered
- Split some complex presence logic on their own functions, and simplify some types

None of this is worth changelog entries nor compatibility-breaking, just implementation details but preparation for what's seen on the other PR.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/transport/epics/presence.ts
+++ b/raiden-ts/src/transport/epics/presence.ts
@@ -22,7 +22,7 @@ import isEqual from 'lodash/isEqual';
 import pick from 'lodash/pick';
 import { getAddress } from '@ethersproject/address';
 import { verifyMessage } from '@ethersproject/wallet';
-import { MatrixClient, MatrixEvent } from 'matrix-js-sdk';
+import { MatrixClient, MatrixEvent, User } from 'matrix-js-sdk';
 
 import { assert } from '../../utils';
 import { RaidenError, ErrorCodes } from '../../utils/error';
@@ -32,7 +32,7 @@ import { RaidenEpicDeps } from '../../types';
 import { RaidenAction } from '../../actions';
 import { RaidenState } from '../../state';
 import { getUserPresence } from '../../utils/matrix';
-import { retryWhile } from '../../utils/rx';
+import { pluckDistinct, retryWhile } from '../../utils/rx';
 import { matrixPresence } from '../actions';
 import { channelMonitored } from '../../channels/actions';
 import { parseCaps, stringifyCaps } from '../utils';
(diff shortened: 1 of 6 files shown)


raiden-ts/src/transport/epics/presence.ts
```diff
@@ -161,6 +161,68 @@ type MatrixPresenceEvent = Omit<MatrixEvent, 'getContent'> & {
...
+        {
+          userId: user.userId,
+          available,
+          ts: last_active_ago ? Date.now() - last_active_ago : user.lastPresenceTs ?? Date.now(),
+          caps: parseCaps(profile.avatar_url),
+        },
+        { address: recovered },
+      );
+    }),
+    retryWhile(intervalFromConfig(config$), { onErrors: [429] }),
```
palango, line 218: Should we include 500 here as well?

raiden-ts/src/transport/epics/presence.ts
```diff
@@ -161,6 +161,68 @@ type MatrixPresenceEvent = Omit<MatrixEvent, 'getContent'> & {
...
+    map(([profile, { presence, last_active_ago }, rtc]) => {
+      // errors raised here will be logged and ignored on catchError below
+      assert(profile?.displayname, 'no displayname');
+      // ecrecover address, validating displayName is the signature of the userId
+      const recovered = verifyMessage(user.userId, profile.displayname) as Address | undefined;
+      assert(recovered === address, `invalid displayname signature: ${recovered} !== ${address}`);
+      if (presence !== user.presence)
+        log.warn('Presence mismatch', { user, presence, last_active_ago });
+      // react on both presence events and rtc channel presence
+      const available = AVAILABLE.includes(presence) || !!rtc;
```
palango, line 207: Is just the existence of the `RTCDataChannel` here enough to infer that it is usable, i.e. that the partner is online?

raiden-ts/src/utils/actions.ts
```diff
@@ -617,61 +617,23 @@ export function createReducer<S, A extends Action = Action>(initialState: S) {
...
-        handle: <
-          AC extends AnyAC & NotHandled<ACs, AD>,
-          H extends Handler<AC>,
-          AD extends AnyAC = AC
-        >(
-          ac: AC | AC[],
-          handler: H,
-        ) => ExtReducer<ACs | AC, Iterate<X>>;
-      };
+  type HandleNew<ACs> = <
```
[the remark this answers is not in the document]
andrevmatos: No, this was already possible before, but it hit an type recursion limit and had to include a workaround to limit it (that 1..15 type iterator). This got cleaned on recent TS versions, but the logic is the same.

raiden-ts/src/transport/epics/presence.ts
andrevmatos: Yes, because as soon as the `RTCDataChannel` isn't usable, a new `rtcChannel` action with `undefined` as payload gets emitted which clears it from the `rtc` mapping, so all channels on it should be always usable.

raiden-ts/src/transport/epics/presence.ts
andrevmatos: I don't think 500 is needed here, because there's a `catchError` right below. If server misbehaves, it's fine to handle and skip a single presence update.

palango (approved):
Nice PR with nicely separated commits!

PR #2491, opened 2021-01-07
Show better message when PFS doesn't suggest partners

### LINKED ISSUE #2485: Unclear message if PFS doesn't suggest partners
Currently the PFS only suggests partners that are online at the moment of the request. This leads to problems in low-traffic networks, where there might be no nodes currently online.

In this case the user is presented with an error in the dApp currently, instead of telling him that there are no suggestions currently.

## Expected Result
- In this special case the user should be shown a better message, mentioning the actual problem instead of an generic error.

## Actual Result
- A generic error message.

palango:

Fixes #2485

**Short description**

- Make the sdk return an empty list when no PFS returns suggested partners
- Display a better message in the dApp in this case

**Definition of Done**

1. Select a token where it's unlikely that nodes are online in the network (like any older scenario player token)
2. Check that the PFS request is done and return an empty list
3. Check that the suggested peers list displays a message mentioning that no nodes are suggested.

### DIFF
--- a/raiden-dapp/src/components/HubList.vue
+++ b/raiden-dapp/src/components/HubList.vue
@@ -2,8 +2,11 @@
   <div class="hub-list">
     <span class="hub-list__header">{{ $t('hub-list.header') }}</span>
     <spinner v-if="loadingHubs" class="hub-list__loading" />
-    <div v-else-if="!loadingHubs && !suggestedHubs.length" class="hub-list__no-hubs">
+    <div v-else-if="!loadingHubs && requestFailed" class="hub-list__no-hubs">
       {{ $t('hub-list.error') }}
+    </div>
+    <div v-else-if="!loadingHubs && suggestedHubs.length === 0" class="hub-list__no-hubs">
+      {{ $t('hub-list.no-results') }}
     </div>
     <div v-for="(suggestedHub, index) in suggestedHubs.slice(0, 3)" v-else :key="index">
       <div class="hub-list__item">
@@ -44,6 +47,7 @@ import { SuggestedPartner } from '@/types';
 export default class HubList extends Vue {
   truncate = Filters.truncate;
   loadingHubs = true;
+  requestFailed = false;
   suggestedHubs: SuggestedPartner[] = [];
   selectedHub = '';
 
@@ -59,6 +63,7 @@ export default class HubList extends Vue {
     try {
       this.suggestedHubs = await this.$raiden.getSuggestedPartners(this.tokenAddress);
     } catch (err) {
+      this.requestFailed = true;
     } finally {
       this.loadingHubs = false;
     }
--- a/raiden-dapp/src/locales/en.json
+++ b/raiden-dapp/src/locales/en.json
@@ -535,7 +535,8 @@
   },
   "hub-list": {
     "header": "Current top 3 suggested hubs (best first)",
-    "error": "Error: List cannot be fetched",
+    "error": "Error: List of hubs cannot be fetched from the Pathfinding Service.",
+    "no-results": "In this token network, the Pathfinding Service does not find a node which is online. Try again later.",
     "select-button": "Select"
   },
   "open-channel": {
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -1462,7 +1462,7 @@ export class Raiden {
     assert(tokenNetwork, ErrorCodes.RDN_UNKNOWN_TOKEN_NETWORK, this.log.info);
     const pfss = options.pfs ? [decode(PFS, options.pfs)] : await this.findPFS();
 
-    let firstError: Error | undefined;
+    let firstResponse: Error | SuggestedPartner[] | undefined;
     return from(pfss)
       .pipe(
         concatMap((pfs) =>
@@ -1474,20 +1474,30 @@ export class Raiden {
             mergeMap(async (response) => response.text()),
             map((text) => {
               const suggestions = decode(SuggestedPartners, jsonParse(text));
-              assert(suggestions.length, ['empty partner suggestion from PFS', { pfs: pfs.url }]);
+
+              if (!suggestions.length) {
+                firstResponse = [];
+                throw new Error('PFS returned no suggested partners');
+              }
               return suggestions;
             }),
             catchError((err) => {
               // store first error and omit to retry next pfs in list
               this.log.info('Could not fetch PFS suggested partners', pfs, err);
-              if (!firstError) firstError = err;
+              if (!firstResponse) firstResponse = err;
               return EMPTY;
             }),
           ),
         ),
         first(), // throws if no first result can be fetched/decoded
         // if first errored, throw first seen error or pass current through
-        catchError((err) => throwError(firstError ?? err)),
+        catchError((err) => {
+          if (Array.isArray(firstResponse)) {
+            return of(firstResponse);
+          } else {
+            return throwError(firstResponse ?? err);
+          }
+        }),
       )
       .toPromise();
   }
--- a/raiden-dapp/CHANGELOG.md
+++ b/raiden-dapp/CHANGELOG.md
@@ -12,6 +12,7 @@
 - [#2476] Fix persitence to remember disclaimer acceptance if selected
 - [#2430] Remember token selection during navigation
 - [#2474] Fix TransferHeader behaviour with no available capacity
+- [#2485] Show better message in case the PFS doesn't suggest partners
 
 ### Added
 
@@ -43,6 +44,7 @@
 [#2458]: https://github.com/raiden-network/light-client/issues/2458
 [#2474]: https://github.com/raiden-network/light-client/issues/2474
 [#2446]: https://github.com/raiden-network/light-client/issues/2446
+[#2485]: https://github.com/raiden-network/light-client/issues/2485
 
 ## [0.14.0] - 2020-11-25
 
--- a/raiden-dapp/tests/unit/components/hub-list.spec.ts
+++ b/raiden-dapp/tests/unit/components/hub-list.spec.ts
@@ -18,7 +18,10 @@ const suggestedPartners = [
   generateSuggestedPartner({ address: addressThree }),
 ];
 
-const createWrapper = (suggestedPartners: SuggestedPartner[]): Wrapper<HubList> => {
+const createWrapper = (
+  suggestedPartners: SuggestedPartner[],
+  request_error = false,
+): Wrapper<HubList> => {
   const vuetify = new Vuetify();
 
   return mount(HubList, {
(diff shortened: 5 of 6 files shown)


raiden-dapp/src/locales/en.json
```diff
@@ -536,6 +536,7 @@
   "hub-list": {
     "header": "Current top 3 suggested hubs (best first)",
     "error": "Error: List cannot be fetched",
+    "no-results": "Currently no nodes are available. Please try later again or connect to a known node.",
```
palango, line 539: @christianbrb This working is not perfect, any suggestions?

raiden-ts/src/raiden.ts
```diff
@@ -1471,20 +1471,29 @@ export class Raiden {
...
           ),
         ),
         first(), // throws if no first result can be fetched/decoded
         // if first errored, throw first seen error or pass current through
-        catchError((err) => throwError(firstError ?? err)),
+        catchError((err) => {
+          if (Array.isArray(firstResponse)) {
+            return of(firstResponse);
+          } else {
+            throw(firstResponse ?? err);
```
palango, line 1494: @andrevmatos I had to change to this from `throwError(firstError ?? err))`. Is this correct, and can you explain me how both differ?

raiden-dapp/src/locales/en.json
```diff
@@ -536,6 +536,7 @@
   "hub-list": {
     "header": "Current top 3 suggested hubs (best first)",
     "error": "Error: List cannot be fetched",
```
christianbrb, line 538:
```suggestion
    "error": "Error: Route cannot be fetched from the Pathfinding Service",
```
Does it make sense?

raiden-dapp/src/locales/en.json
christianbrb, line 539:
```suggestion
    "no-results": "In this token network, the Pathfinding Service does not find a node which is online. Try again later.",
```

raiden-dapp/src/locales/en.json
christianbrb, line 538:
```suggestion
    "error": "Error: List of hubs cannot be fetched from the Pathfinding Service.",
```

raiden-dapp/src/locales/en.json
christianbrb: Done, made suggestions

palango: Thanks!

raiden-ts/src/raiden.ts
```diff
@@ -1471,20 +1461,29 @@ export class Raiden {
             mergeMap(async (response) => response.text()),
             map((text) => {
               const suggestions = decode(SuggestedPartners, jsonParse(text));
-              assert(suggestions.length, ['empty partner suggestion from PFS', { pfs: pfs.url }]);
+
+              if (!suggestions.length) {
+                firstResponse = [];
+              }
```
weilbith, line 1467: Are you sure this works? If we get an empty array back and we do not manually throw (as we did with the assert before) this will just return the empty array. This means the `first()` operator just goes straight ahead and does not query any further PFS. And I think that's what we want. All you logic about the `firstResponse` can be also an empty array will never be used.
I think you should extend the test cases. So you want to have multiple PFS. The first returns an empty list. Then one case where all following PFS throw, so you expect to get the empty list of the first. And one test case where any future PFS in the list return a positive array, so you expect that one to get returned in the end.
The CodeCov remark on this PR is a hint that you either lack in tests cases or - what I stated - your code in line 1481 never gets applied.

Sorry if I'm totally wrong. :see_no_evil:

weilbith (changes requested):

raiden-ts/src/raiden.ts
weilbith: Eventually this is easily solvable by adding a filter like `first(suggestions => first.length)`. Not too sure about the rest as we still have no error. xD

raiden-dapp/src/components/HubList.vue
```diff
@@ -2,9 +2,12 @@
   <div class="hub-list">
     <span class="hub-list__header">{{ $t('hub-list.header') }}</span>
     <spinner v-if="loadingHubs" class="hub-list__loading" />
-    <div v-else-if="!loadingHubs && !suggestedHubs.length" class="hub-list__no-hubs">
+    <div v-else-if="!loadingHubs && requestErrored" class="hub-list__no-hubs">
```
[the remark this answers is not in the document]
palango: You have a point. Seems like I errored here. 😏

raiden-ts/src/raiden.ts
palango: No, you're right. For how I intended it to work there's at least a early return missing, but there might be additional problems., I'll write more tests.

raiden-ts/src/raiden.ts
andrevmatos: @palango `throw` doesn't look like an exported function from `rxjs`, and instead the `throw` operator, which should work here since rxjs catches all exceptions and errors the output, but the intended behavior here would be for you to `return throwError(...)`, which will return a new observable which errors with given value upon subscription.

raiden-ts/src/raiden.ts
```diff
@@ -1474,20 +1464,29 @@ export class Raiden {
...
           ),
         ),
         first(), // throws if no first result can be fetched/decoded
         // if first errored, throw first seen error or pass current through
-        catchError((err) => throwError(firstError ?? err)),
+        catchError((err) => {
+          if (Array.isArray(firstResponse)) {
+            return of(firstResponse);
+          } else {
+            throw firstResponse ?? err;
```
andrevmatos, line 1487: You should `return throwError(...)` here, preferably

raiden-ts/src/raiden.ts
andrevmatos: As @weilbith stated, you should set `firstResponse` to the empty array **and** throw on the map, to get the next pfs in queue to be tried. An alternative would be to abuse the error handling in JS inside `map` to `throw suggestions`, which is the empty array (and not an error, as usually advised), so it can be both set to `firstResponse` and omitted on the `catchError`

raiden-ts/src/raiden.ts
```diff
@@ -1474,20 +1474,30 @@ export class Raiden {
             mergeMap(async (response) => response.text()),
             map((text) => {
               const suggestions = decode(SuggestedPartners, jsonParse(text));
-              assert(suggestions.length, ['empty partner suggestion from PFS', { pfs: pfs.url }]);
+
+              if (!suggestions.length) {
+                firstResponse = [];
+                throw Error();
```
andrevmatos, line 1480: There's a log call on the `catchError`, it may be interesting to put some message here to get logged on this case.

raiden-ts/src/raiden.ts
```diff
@@ -1474,20 +1474,30 @@ export class Raiden {
             mergeMap(async (response) => response.text()),
             map((text) => {
               const suggestions = decode(SuggestedPartners, jsonParse(text));
-              assert(suggestions.length, ['empty partner suggestion from PFS', { pfs: pfs.url }]);
+
+              if (!suggestions.length) {
+                firstResponse = [];
```
andrevmatos, line 1479: Please, set this only if `!firstResponse`, to avoid overwriting the previous error (possibly).

andrevmatos: The `throw` can be outside of this condition (although inside the length condition), since it's needed on both cases to continue retrying the other pfss.

raiden-ts/src/raiden.ts
```diff
@@ -1474,20 +1474,31 @@ export class Raiden {
             mergeMap(async (response) => response.text()),
             map((text) => {
               const suggestions = decode(SuggestedPartners, jsonParse(text));
-              assert(suggestions.length, ['empty partner suggestion from PFS', { pfs: pfs.url }]);
+
+              if (!firstResponse && !suggestions.length) {
+                firstResponse = [];
+                this.log.info('PFS returned no suggested partners', pfs);
```
andrevmatos, line 1480: This log call will make it get logged twice: one here, and one inside the `catchError`

raiden-ts/src/raiden.ts
andrevmatos: `throw` must not be conditional to `!firstResponse`, or else it won't retry the next PFS if the array is empty but `firstResponse` is already set. The whole block should be something like:
```
if (!suggestions.length) {
  if (!firstResponse) firstResponse = suggestions;
  throw new Error('your message to be logged');
}
return suggestions;
```

andrevmatos: Or, if you wanna abuse the `throw` operator to reuse the log and var set inside the catchError block:
```ts
if (!suggestions.length) throw suggestions; // throw empty array to be set to firstResponse
return suggestions;
```

palango: We cannot use `(!firstResponse)` as a condition because that would mean an error from PFS1 would never be overwritten by an empty result of PFS2.

raiden-ts/src/raiden.ts
palango: Added it to the error.

raiden-ts/src/raiden.ts
andrevmatos: This is still an issue on latest commit.

raiden-ts/src/raiden.ts
```diff
@@ -1474,20 +1474,30 @@ export class Raiden {
             mergeMap(async (response) => response.text()),
             map((text) => {
               const suggestions = decode(SuggestedPartners, jsonParse(text));
-              assert(suggestions.length, ['empty partner suggestion from PFS', { pfs: pfs.url }]);
+
+              if (!suggestions.length) {
+                firstResponse = [];
+                throw Error('PFS returned no suggested partners');
```
andrevmatos, line 1480: Nitpick: prefer `throw new Error()`, since Error is a constructor.

raiden-ts/src/raiden.ts
palango: I think the argument from above holds here as well:
> We cannot use `(!firstResponse)` as a condition because that would mean an error from PFS1 would never be overwritten by an empty result of PFS2.

raiden-ts/src/raiden.ts
andrevmatos: This is fine, `firstResponse` is **first** response, either an error or an empty array.

andrevmatos: But I think then it's already nitpick, I think both behaviors are acceptable, so leave it as you find better.

palango: But our priority is to rather return an empty list than an error, because that means more information for the user. So we want to overwrite an earlier error with an empty list.

This is what https://github.com/raiden-network/light-client/pull/2491/files#diff-7c34034053ef8211fe51e9a7d5693c5b6249d7183ea9d8ce4d7327537e2b700aR1640 tests.

raiden-ts/src/raiden.ts
palango: ✅

weilbith (approved):
Looks good to me. Thanks for your work on all the feedback. Especially the cases we all test are good (although I wonder if we could outsource common parts as "fixtures"). :pray:

---

I must admit I'm not 100% happy with the SDK code. It works and is simple enough to understand. But it does not feel straight forward. But good enough and this was not introduced by this PR, just continued.

PR #2492, opened 2021-01-07
Introduce busy key value pair to track busy status of each channel se…

### LINKED ISSUE #2418: Channel close button incorrectly inactivated
## Steps to Reproduce
1. Open two channels for the same token
2. Go to the channels page
3. Close one channel

## Expected Result
- I can directly close the other channel as well

## Actual Result
- The other channels "Close" button is greyed out
  - Seems like it's re-activated as soon as the "close" transaction is confirmed. see screenshots

## Screenshots

Screenshot 2020-11-30 at 17 00 35
Screenshot 2020-11-30 at 17 02 30

```
- Firefox 83, macOS 10.15.7
```

agatsoh:
…parately

Fixes #2418  

**Short description**

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-dapp/tests/unit/components/channels/channel-list.spec.ts
+++ b/raiden-dapp/tests/unit/components/channels/channel-list.spec.ts
@@ -41,6 +41,7 @@ describe('ChannelList.vue', () => {
         } as Token,
         channels: TestData.mockChannelArray,
         expanded,
+        busy: { undefined: false },
       },
       stubs: ['raiden-dialog'],
       mocks: {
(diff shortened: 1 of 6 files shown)


raiden-dapp/tests/unit/components/channels/channel-list.spec.ts
```diff
@@ -84,20 +85,21 @@ describe('ChannelList.vue', () => {
 
   test('disable actions while "busy", show spinner on "selectedChannel"', async () => {
```
taleldayekh, line 86: Feels like this test is testing multiple things in one test case. It is pretty difficult to make sense of the things being tested. I would consider splitting it up in separate test cases. I think the fact that there has to be comments in the tests indicates that they perhaps should be split. ;)

taleldayekh (approved):
Good job @agatsoh, I'm approving the PR because it looks good and works as expected. I left one suggestion for improvements.

raiden-dapp/tests/unit/components/channels/channel-list.spec.ts
taleldayekh: I know it was like this from before, but if you find the time it would be nice to clean it up a bit.

agatsoh: @taleldayekh  I feel the things checked are part of the same transaction, the channel is first made `busy` => `true` and then made `busy` => `false` which is the same transaction as we perform it in the UI, its just that `busy` => `false` happens automatically for us without human action but on the UI it is part of the same transaction, so should remain in the same test.
What I missed is changing the name of the test which has become misleading and I am changing it.

taleldayekh: Ok, understood. Then go ahead and merge it. Good job!

Issue #2493, opened 2021-01-08
Investigate Nightly failures


palango:
Failure: https://app.circleci.com/pipelines/github/raiden-network/light-client/8093/workflows/738785ab-ac1d-4c3b-ad31-f81969f0a844/jobs/49904

Might be related to https://github.com/raiden-network/light-client/commit/bbb3bcfd05dca7652ff90703347aefa1b7dc31d6

In https://app.circleci.com/pipelines/github/raiden-network/light-client/8093/workflows/738785ab-ac1d-4c3b-ad31-f81969f0a844/jobs/49902 it looks like the correct downgrade version of `source-map-loader` is used:
```
warning "workspace-aggregator-76136779-9a2a-4900-b3b2-eb4e1c99a3b1 > raiden-dapp > source-map-loader@1.1.3" has unmet peer dependency "webpack@^4.0.0 || ^5.0.0".

palango:
@weilbith Any other ideas for what to look out? Otherwise I'd probably increase the cache version and try running with that.

weilbith:
I'll take a closer look soon.

PR #2494, opened 2021-01-08
Ignore - Test PR do debug nightlies at CI

weilbith:
Related to: #2493


palango:
We found that the latest version of `copy-webpack-plugin` is also incompatible with webpack 4.x:

https://github.com/webpack-contrib/copy-webpack-plugin/issues/575

PR #2515, opened 2021-01-20
chore: Ignore typedoc updates

palango:
This was pulled into https://github.com/raiden-network/light-client/pull/2508 , so let's ignore it for now.

See `renovate` options here: https://docs.renovatebot.com/configuration-options/

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -8,6 +8,7 @@
       "packagePatterns": [
         "*"
       ],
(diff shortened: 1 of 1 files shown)


andrevmatos (approved):
lgtm. usually a better way to ignore package updates on renovate is to close the update PR, but since this is batched with the others, you can't do it here. let's keep in mind to upgrade this soon and remove this line when possible

PR #2527, opened 2021-01-25
dapp: add check for required browser features

### LINKED ISSUE #2497: dApp - disable if browser does not support required features
## Description

The dApp is based on some modern features of web-browsers like service workers. To keep the implementation simple without the need to maintain a lot of backwards compatible code or messages, we just "disable" the applications in such browsers. This should like we already detect if JavaScript is enabled/available and if not do not even start to render/mount the application.
This also keeps legal requirements issues simple as we only have a single "version" of the application that fully aligns with all requirements no matter what.

## Acceptance criteria
- a list of features we require
- extended check for required features outside of Vue 

## Tasks
- [ ]

weilbith:
Fixes #2497 

**Short description**
Please try it yourself in your web-browser an make sure that the application gets served for you.

**Definition of Done**

1. Run the dApp
2. Make sure you get not shown the error message.

### DIFF
--- a/raiden-dapp/CHANGELOG.md
+++ b/raiden-dapp/CHANGELOG.md
@@ -23,6 +23,7 @@
 - [#2399] Hint for user to stay online while receving transfers
 - [#2458] Click on notification snackbar to open notification panel
 - [#2446] Add `VUE_APP_SUBKEY_ORIGIN_URL` env variable to recover backups
+- [#2527] Add check if browser supports all necessary features or block dApp
 
 ### Changed
 
@@ -51,6 +52,7 @@
 [#2418]: https://github.com/raiden-network/light-client/issues/2418
 [#2414]: https://github.com/raiden-network/light-client/issues/2414
 [#2440]: https://github.com/raiden-network/light-client/issues/2440
+[#2527]: https://github.com/raiden-network/light-client/issues/2527
 
 ## [0.14.0] - 2020-11-25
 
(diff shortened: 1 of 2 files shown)


palango:
Looks good on the first view, but you should add a changelog entry.

And do you know of any browser that doesn't support the feature set to test the bad case?

weilbith:
> Looks good on the first view, but you should add a changelog entry.

Good idea. I'll do so!

weilbith:
> And do you know of any browser that doesn't support the feature set to test the bad case?

Hmm. Maybe an old Internet Explorer which stopped development some years ago.
@christianbrb as the only Windows user in the team: could you try install the old IE?

weilbith:
@palango I got an old Firefox v41.0 from an achieve. This works. Means I get the message and not the application.
Unfortunately this fu**** up my actual Firefox setup (the config files, plugins, user data,,,). I'll need some time to fix it.

@christianbrb forget about my comment. I did it already.  --^

palango:
> @palango I got an old Firefox v41.0 from an achieve. This works. Means I get the message and not the application.
> Unfortunately this fu**** up my actual Firefox setup (the config files, plugins, user data,,,). I'll need some time to fix it.

Oh damn, sorry for that! But thanks for testing.

palango (approved):

weilbith:
I'll rebase to resolve the merge conflicts.

PR #2531, opened 2021-01-25
Updating disclaimers

### LINKED ISSUE #2355: Update the disclaimer
## Description
Currently, the disclaimer is getting reworked. It needs to be updated once it is done.

## Acceptance criteria
- 

## Tasks

Needs to be done first:
- [x] Add a page to the GitHub repo with the Terms & Conditions (https://docs.google.com/document/d/1Ms3qHF54dw8_AhbsulRAkFe9h1FAg0pttPjLXcv7tRI/edit?ts=60096d59#)
- [x] Update the disclaimer (https://docs.google.com/document/d/1Y3IbX6aOGhcTslxG0jgKvA5Rd6WNau9kBBfwgbbnYbk/edit)
    - [x] Update in the dApp -> Link to the terms & conditions on the GitHub repo
    - [x] Update in the SDK -> Link to the terms & conditions on the GitHub repo
    - [x] Update in the Readme -> Link to the terms & conditions on the GitHub repo
- [x] Add a link to the dApp, which links to the terms & conditions on the GitHub repo (below the link to the privacy policy)
~~- [ ] Update the privacy policy of brainbot.li (https://docs.google.com/document/d/1PGzSOdqdnAZY-Lk81ejeHCFojaV0fFdq4H4meiEcDJE/edit#)~~

taleldayekh:
Fixes #2355 

**Short description**

@christianbrb Could you please double check so that the content is correct and meets the acceptance criteria. Look at the screenshots below for the dApp and CLI/API disclaimer and the SDK README for the SDK disclaimer.

Regarding the terms and conditions. I don't believe I have access to the brainbot.li site. I added the new README in the repo and referenced it in the CLI/API disclaimer and SDK disclaimer, the idea now is to reference the same terms and conditions on brainbot.li in the dApp, would it be acceptable to just point to the same README as for the CLI/API and SDK here in the repo instead?

@weilbith I'm assigning you to review the changes in the component.

Screenshot 2021-01-25 at 16 18 42

Screenshot 2021-01-25 at 17 48 01

**Definition of Done**

1.
2.


raiden-dapp/src/views/DisclaimerRoute.vue
```diff
@@ -1,10 +1,52 @@
...
-      <div class="disclaimer__content__paragraphs font-weight-light">
-        <p v-for="(paragraph, index) in $t('disclaimer.paragraphs')" :key="index">
-          {{ paragraph }}
-        </p>
+      <div class="disclaimer__content__text">
+        <span class="disclaimer__content__text__header">
+          {{ $t('disclaimer.user-info.header') }}
+        </span>
+        <span class="disclaimer__content__text__body">{{ $t('disclaimer.user-info.body') }}</span>
+        <div>
```
weilbith, line 9: What is the outside `<div>` for? :thinking:
If there is something surrounded needed, maybe a paragraph would suites better.

raiden-dapp/src/views/DisclaimerRoute.vue
```diff
@@ -1,10 +1,52 @@
...
+            <li>
+              {{ bulletPoint.text }}
+              <a :href="disclaimerUrlMappings[bulletPoint.url]" target="_blank">
+                {{ bulletPoint.link }}
+              </a>
+            </li>
+          </ul>
+        </div>
+        <span v-if="imprint" class="disclaimer__content__text__header">
+          {{ $t('disclaimer.terms.header') }}
```
weilbith, line 24: `if imprint then print terms` - this is a little confusing. Shouldn't what is called `terms` now be actually called `imprint`? Or does the `imprint` consists of the `terms`? I'm not to sure about the last, but if not or if it only consists of the `terms` and nothing more, I would consider to rename them to the same.

raiden-dapp/src/views/DisclaimerRoute.vue
```diff
@@ -1,10 +1,52 @@
...
+        <div>
+          <ul
+            v-for="(bulletPoint, index) in $t('disclaimer.user-info.bullet-points')"
+            :key="index"
+            class="disclaimer__content__text__bullet-points"
+          >
+            <li>
+              {{ bulletPoint.text }}
+              <a :href="disclaimerUrlMappings[bulletPoint.url]" target="_blank">
+                {{ bulletPoint.link }}
```
weilbith, line 18: Suggestion (and I think we actually discussed this solution together):
I think the `url` property is quite confusing. From within the translations this just makes no sense. You read it and don't know it is a special key that gets converted. As you have written it, the `url` is actually a proper translation which could differ for all languages. I would like to (re)suggest to convert the array into an object and use just the names.
So in the locals this would looks like:

```js
bullet-points: {
   portal: {
     text: "foo",
     link: "bar",
   },
   docs: {
     ...
   },
   ... 
}
```

And then in the template you can just do `v-for="(bulletPoint, name) in $t(...`.
In fact this only changes a little to your current implementation and mapping in the component. But I personally think that these subtile changes represent a better structure. It has a stronger separation of translation and data. Although the linking works most-likely the same.

---

If you still prefer your approach and want to stick with it, please at least rename the `url` translation key as it is quite confusing. But maybe I'm the only one...

weilbith (changes requested):

weilbith (commented):
Having translation and language independent data links is always a tricky thing. You have a really good first approach! :+1:

raiden-dapp/src/views/DisclaimerRoute.vue
taleldayekh: The `div` was added because Vue does not allow me to have two `v-for` at the same "level". The console outputs: `Duplicate keys detected: '0'. This may cause an update error.`. I think it seems more correct to wrap an `<ul>` in a div rather than a paragraph?

raiden-dapp/src/views/DisclaimerRoute.vue
taleldayekh: You are right, Christian informed me that there are actually two different texts now, these will need to be adjusted in the footer as well.

raiden-dapp/src/views/DisclaimerRoute.vue
taleldayekh: I can make it an object instead, no problem. :)

raiden-dapp/src/views/DisclaimerRoute.vue
```diff
@@ -14,23 +14,26 @@
           >
             <li>
               {{ bulletPoint.text }}
-              <a :href="disclaimerUrlMappings[bulletPoint.url]" target="_blank">
+              <a :href="disclaimerUrlMappings[index]" target="_blank">
```
weilbith, line 17: This is not the index anymore. Rather the `name` or something similar. :shrug: 
But the more important is that you can resolve [this](https://github.com/raiden-network/light-client/pull/2531#discussion_r566167769) now. Before both `v-for` where using the same numbers (indexes) as keys. Now they do no-more. So you can get rid of the `<div>` which is potential layout risk. If it should be still an issue I would recommend to at least use a `<p>` as this is an inline element. I don't know what "magic" Vuetify apply automatically else.

raiden-dapp/src/views/DisclaimerRoute.vue
weilbith: Please take a look [here](https://github.com/raiden-network/light-client/pull/2531#discussion_r566718964)

taleldayekh: 👍🏼

raiden-cli/src/disclaimer.json
```diff
@@ -1 +1 @@
...
...                                       │\n │ Privacy warning                                                        │\n │ Please be aware, that by using the Raiden Light Client, among others   │\n │ your Ethereum address, channels, channel deposits, settlements and the │\n │ Ethereum address of your channel counterparty will be stored on the    │\n │ Ethereum chain, i.e. on servers of Ethereum node operators and ergo    │\n │ are to a certain extent publicly available. The same might also be     │\n │ stored on systems of parties running Raiden nodes connected to the     │\n │ same token network. Data present in the Ethereum chain is very unlikely│\n │ to be able to be changed, removed or deleted from the public arena.    │\n │                                                                        │\n │ Also be aware, that data on individual Raiden token transfers will be  │\n │ made available via the Matrix protocol to the recipient, intermediating│\n │ nodes of a specific transfer as well as to the Matrix server operators,│\n │ see Raiden Transport Specification.                                    │\n └────────────────────────────────────────────────────────────────────────┘\n\n"
```
christianbrb, line 1: Can you remove the explicit line breaks? /n There is a risk, that the disclaimer is not readable on every screen (especially on mobile).

raiden-dapp/src/views/DisclaimerRoute.vue
taleldayekh: Done!

raiden-dapp/src/views/DisclaimerRoute.vue
taleldayekh: Done!

weilbith (approved):
Perfect :ok_hand:

raiden-dapp/src/locales/en.json
```diff
@@ -404,7 +438,8 @@
     }
   },
   "application": {
-    "privacy-policy": "Privacy Policy, Disclaimer & Imprint"
+    "privacy-policy": "Privacy Policy, Disclaimer & Imprint.",
+    "terms": "Terms."
```
christianbrb, line 442: „Terms of Use" instead of just Terms

christianbrb (changes requested):
Needs some changes. See comments 

- [x] This line needs to get changed: https://github.com/raiden-network/light-client/blob/fef0a7d56d94a0c97674618e3ec729391f75df23/raiden-cli/src/index.ts#L56
- [x] ~~This picture needs an update: https://github.com/raiden-network/light-client/blob/fc5f16dafbeb1e370c92615c487b7b1ace01bb06/raiden-ts/docs-source/navigate-dapp.md~~ Remove picture of disclaimer from documentation.
- [x] These lines and the lines below need an update https://github.com/raiden-network/light-client/blob/17d2c7247a3787c88b58c71b247daec92828f2d5/raiden-cli/src/raiden.ts#L218 (I have read, understood and hereby agree to the above Terms of Use, Privacy Policy including the Disclaimer and Privacy Warning)

raiden-cli/src/disclaimer.json
taleldayekh: That is for the CLI, if I remove them it will just be a long string of text. :)

raiden-dapp/src/locales/en.json
taleldayekh: Ah, yes. Good catch!

taleldayekh:
> * This line needs to get changed:
 Not sure what this one (the first in the list) should be changed to?

taleldayekh:
> * This picture needs an update: https://github.com/raiden-network/light-client/blob/fc5f16dafbeb1e370c92615c487b7b1ace01bb06/raiden-ts/docs-source/navigate-dapp.md

Actually, I don't think we need to have an image of the disclaimer in the docs. I would remove it completely. :)

PR #2549, opened 2021-02-06
Wait for Capacity Update global messages to be sent before resolving deposit and openChannel requests

### LINKED ISSUE #2536: Wait for the global messages to be sent after a channel deposit before resolving the REST CLI request
## Description
On some scenarios, a transfer can be started to fast after the deposit is confirmed, before the global messages can finish being sent, which can make the transfer fail to be resolved by the PFS if it didn't get the `PFSCapacityUpdate` by the time.
It's hard to know for sure when did the PFS did get the message, but with global messages taking up to 15-30s to be delivered, we can at least ensure we've finished sending before resolving the deposit REST request and allowing the SP to move on with the next task (which can be a transfer).

## Acceptance criteria
- Deposit REST request resolves only after that deposit's global messages finish being sent.

## Tasks
- [ ]

andrevmatos:
Fixes #2536 

**Short description**
This should help in case sending these messages gets slowed down, to prevent transfers which are started right after it from being done before PFS has the opportunity to detect these updates, which should happen only on scripted environments like the SP.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/helpers.ts
+++ b/raiden-ts/src/helpers.ts
@@ -577,19 +577,16 @@ export async function getState(
  * @param state$ - Observable of RaidenStates
  * @param opts - Options
  * @param opts.meta - channelId on which to wait for a deposit
- * @param opts.config - RaidenConfig to check pfsRoom from
+ * @param opts.config - RaidenConfig to check if PFS is enabled
  * @returns Promise for undefined or messageGlobalSend.success action
  */
 export async function waitForPFSCapacityUpdate(
   action$: Observable<RaidenAction>,
   state$: Observable<RaidenState>,
-  {
-    meta,
-    config,
-  }: { meta: channelDeposit.request['meta']; config: Pick<RaidenConfig, 'pfsRoom'> },
+  { meta, config }: { meta: channelDeposit.request['meta']; config: RaidenConfig },
 ) {
   const pfsRoom = config.pfsRoom;
-  if (!pfsRoom) return;
+  if (!pfsRoom || config.pfs === null) return;
   const postMeta: messageGlobalSend.request['meta'] = { roomName: '', msgId: '' };
   return asyncActionToPromise(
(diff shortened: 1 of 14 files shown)


raiden-ts/src/helpers.ts
```diff
@@ -552,3 +568,57 @@ export async function getState(
...
+export async function waitForPFSCapacityUpdate(
+  action$: Observable<RaidenAction>,
+  state$: Observable<RaidenState>,
+  {
+    meta,
+    config,
+  }: { meta: channelDeposit.request['meta']; config: Pick<RaidenConfig, 'pfsRoom'> },
+) {
+  const pfsRoom = config.pfsRoom;
+  if (!pfsRoom) return;
```
palango, line 592: While looking into this I'm wondering whether this works as expected. In https://github.com/raiden-network/light-client/blob/master/raiden-cli/src/index.ts#L278 only `pfs` is set to `null`, but `pfsRoom` should still be set, so this would still be called.

andrevmatos: Hmm, good point. From what I can see, currently `config.pfs` only disable using the PFS, not updating it. 
This is the line which actually filters if the message should be sent or not:
https://github.com/raiden-network/light-client/blob/master/raiden-ts/src/services/epics.ts#L254-L255
And this was what I used here, so the idea was "wait for the message to be sent whenever it is going to be sent".
But I agree that it's confusing to have two config options which are able to disable pfs. Soon enough we should be removing `pfsRoom` anyway, so I'll just add both options on the condition checks, and it should be good to go for now.
Good catch!

andrevmatos: :heavy_check_mark:

palango (approved):
Nice, thanks for fixing the issue so quick.

Also nice cleanup in the tests! 🎖️

PR #2554, opened 2021-02-10
Update eslint rules to enforce import order, lodash & self-closing tags

### LINKED ISSUE #1332: Update eslint rules
## Description

Follow up to https://github.com/raiden-network/light-client/pull/1308#discussion_r408617114
We should enforce a couple more linting rules throughout the code base

## Acceptance criteria
- Self closing tags should be enforced: https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/html-self-closing.md
- Enforce import ordering on the dApp
- Added (@andrevmatos): enforce lodash import style

## Tasks
- [x] Implement new rules
- [x] Check why `order/import` does not work in SDK tests https://github.com/raiden-network/light-client/pull/1315/files#r411426482

andrevmatos:
Fixes #1332

**Short description**
Using `eslint-plugin-simple-import-sort`, since its defaults are saner than `import/order` and also more customizable.
- Import order is strictly: 
  1. side-effects + `*/{mocks/fixtures/patches}` (sdk tests mocks w/side-effects)
  2. native,external,3rd-party
  3. `raiden-*` 
  4. `@/*`
  5. parents (sorted by lvs above+below)
  6. siblings (sorted by ls below)`
- Inside the `import` destructuring, members are also sorted now.
- Enforce `import type` from TS3.8 where possible.
- SDK tests import SDK src from `@/` now.
- Enforce `import a from 'lodash/a';` in lodash imports
- Enforce self-closing tags in templates when possible (``)

**Definition of Done**

1.
2.


andrevmatos:
@thore I've skipped replacing parent imports in SDK with `@/` ones because it seems `tsserver` isn't very smart about auto-importing with it. It seems it tries instead to identify which would be shorter (relative vs aliased) and often produce inconsistent imports on my setup. Keeping all relative as is otoh always seem to work.

.eslintrc.json
```diff
@@ -28,12 +30,34 @@
...
     "import/no-extraneous-dependencies": "error",
+    "import/no-duplicates": "error",
+    "simple-import-sort/imports": [
+      "error",
+      {
+        "groups": [
+          ["^\\u0000", "/(fixtures|mocks|patches)$"],
+          ["^@?\\w"],
+          ["^raiden-\\w+"],
+          ["^"],
```
weilbith, line 56: If this should be the pattern for absolute paths, why not adding the leading slash? Because this way it looks a bit awkward/information less.

.eslintrc.json
```diff
@@ -28,12 +30,34 @@
...
+    "import/first": "error",
+    "import/newline-after-import": "error",
     "import/no-extraneous-dependencies": "error",
+    "import/no-duplicates": "error",
+    "simple-import-sort/imports": [
+      "error",
+      {
+        "groups": [
+          ["^\\u0000", "/(fixtures|mocks|patches)$"],
+          ["^@?\\w"],
```
weilbith, line 54: Isn't this mixing imports of installed packages and import via the `@` alias which is in fact an absolute path? I would rather split this and make the alias imports after all packages, as it is 1st party.

.eslintrc.json
```diff
@@ -28,12 +30,34 @@
...
+          ["^raiden-\\w+"],
```
weilbith, line 55: Basically a nice idea to have it as separate group. But in fact it will always be a single import statement, right?

raiden-ts/tests/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "import/no-unresolved": "off"
```
weilbith, line 3: Why that?

raiden-ts/tests/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../tsconfig.json",
   "compilerOptions": {
     "paths": {
-      "raiden-ts/*": ["../src/*"]
+      "@/*": ["../src/*"]
```
weilbith, line 5: While we are already on it: eventually add an alias for `<rootDir>/tests/`? Might be helpful to access shared mocks/generators/fixtures/...

.eslintrc.json
```diff
@@ -67,6 +68,15 @@
...
     "@typescript-eslint/consistent-type-imports": "warn",
+    "lodash/import-scope": ["warn", "method"],
+    "lodash/preferred-alias": "warn",
+    "lodash/callback-binding": "warn",
+    "lodash/collection-method-value": "warn",
+    "lodash/collection-return": "warn",
+    "lodash/no-double-unwrap": "warn",
+    "lodash/no-extra-args": "warn",
+    "lodash/no-unbound-this": "warn",
+    "lodash/unwrap": "warn",
```
weilbith, line 79: This is quite a lot of customization. Is there no good configuration (`recommended`, `canonical`, `v3`, ...) we could use for `extends`?

weilbith (changes requested):
Very good PR. I love strict and deterministic code formatting. Maybe I should try `Go`. xD

.eslintrc.json
andrevmatos: https://github.com/lydell/eslint-plugin-simple-import-sort#custom-grouping , this is just a customization of the `default groups`. Information-less is mostly because I had to remove the comments for JSON.

.eslintrc.json
andrevmatos: No, because this matches `@` only if it's followed by a `\w`, so interpreting this as a scoped external package. The `@` alias actually ends up in the `^` match, grouped right after any of our mono-repo imports, if any.

.eslintrc.json
andrevmatos: Yes, usually, but we may change that in the future. I just didn't want it to go with `^` e.g. `@/` packages because then they get sorted later, and it gets ugly then.

raiden-ts/tests/.eslintrc.json
andrevmatos: Even with `eslint-import-resolver-typescript`, I couldn't get `eslint` to detect and respect the `@/` alias in `tests/tsconfig.json`, so in all tests eslint was erroring because it didn't find that package, so I had to disable it.

raiden-ts/tests/tsconfig.json
andrevmatos: We don't do that many relative imports in the tests, and I'd rather have that matching the SDK pattern of using relative imports when possible. We could discuss this later in its own issue, if you wish, but I think we could leave it out of scope for this PR.

.eslintrc.json
andrevmatos: These are the [possible errors group from `eslint-plugin-lodash`](https://github.com/wix/eslint-plugin-lodash#possible-errors). Unfortunately, there's no extendable set without including several other customizations which I don't want, e.g. preferring lodash's functions over native methods.

.eslintrc.json
weilbith: Alright. According to the comments below I misunderstood the meaning/purpose of this pattern.

raiden-ts/tests/.eslintrc.json
weilbith: :expressionless:

raiden-ts/tests/tsconfig.json
weilbith: Okay. Because for the dApp it will be definitely helpful.

.eslintrc.json
weilbith: Hmm okay. Then it is what it is.

weilbith (approved):

PR #2562, opened 2021-02-15
dapp: fix action buttons for channels on mobile

### LINKED ISSUE #2420: Distorted withdraw deposit buttons on channels screen for LC on mobile
## Steps to Reproduce
 Please refer to these steps to get started with testing on the metamask mobile app
https://github.com/raiden-network/light-client/issues/2400#issuecomment-734294170

## Expected Result
- The withdraw and deposit buttons show like the web app.

## Actual Result
- 
light_client_mobile_channel_screen

weilbith:

Fixes #2420

**Short description**
This basically just removes some Vuetify magic we don't need but causes the issue.

**Definition of Done**

1. Serve the dApp locally (use PR tool?)
2. Make sure you have at least one channel open in the selected token network
3. Goto the channels view
4. Make sure the buttons to withdraw and deposit work fine on desktop
5. Switch to the mobile view in the dev tools (or Ctrl+Shift+m)
6. Make sure that the buttons still look fine (note that there is no fancy slider effect on mobile as before)

### DIFF
--- a/raiden-dapp/CHANGELOG.md
+++ b/raiden-dapp/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- [#2420] Fix withdraw and deposit button for channels on mobile
+
+[#2420]: https://github.com/raiden-network/light-client/issues/2420
+
 ## [0.15.0] - 2021-01-26
 
 ### Fixed
--- a/raiden-dapp/src/components/ErrorMessage.vue
+++ b/raiden-dapp/src/components/ErrorMessage.vue
@@ -2,7 +2,7 @@
   <v-row class="error-message">
     <v-col cols="12">
       <div class="error-message__image">
-        <v-img :src="require('../assets/error.png')" />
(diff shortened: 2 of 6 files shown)


agatsoh:
Galaxy S5/ Moto G4 shows this (checked on chromium)
image

pixel 2 / XL users get a aligned buttons 

Galaxy S9/S9+
image

Ofcourse we cannot cater to each and every mobile viewport

agatsoh (approved):
Looks good to me

weilbith:
Hmm. Do you have a good idea how we could solve this? My only suggestion would be to reduce the gap between the two buttons. Not too sure how nice this looks for larger displays. :thinking:

agatsoh:
I think if we introduce more gap between the  `withdraw/deposit` and the `close` buttons it will mess up the some other part of the template. For  now we let it remain and if some of us while testing on actual mobile or some user comes up with an issue then we could look. Thats just my view, rather than over engineer the stuff.

weilbith:
I'll now do the rebase for the fixup commit and then merge the PR as soon as the CI is done.

PR #2576, opened 2021-02-18
dApp: adding disconnect button

### LINKED ISSUE #1515: I want to disconnect the dApp
## User Story

As a user I want to disconnect the dApp e.g. if I don't want to use it anymore or to access features that are only available while disconnected (e.g. state upload)

Currently it's not obvious that the only way to disconnect is to browser reload

taleldayekh:
Fixes #1515 

**Short description**
Adds a disconnect button at the bottom of the Account screen that is available when a user is connected to the dApp. Clicking the button stops Raiden and resets the state.

Screenshot 2021-02-18 at 17 36 45

**Definition of Done**

1. Connect to the dApp and pop open the Account screen.
2. Click the Disconnect button at the bottom of the screen.

### DIFF
--- a/raiden-dapp/tests/unit/components/account/account-content.spec.ts
+++ b/raiden-dapp/tests/unit/components/account/account-content.spec.ts
@@ -200,4 +200,36 @@ describe('AccountContent.vue', () => {
       }),
     );
   });
+
+  test('displays "Disconnect" button when connected', async () => {
+    const wrapper = await createWrapper(
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      false,
+      undefined,
+    );
+    const disconnectButton = wrapper.find('.account-content__disconnect__button');
+
+    expect(disconnectButton.exists()).toBe(false);
+  });
+
+  test('does not display "Disconnect" button when not connected', async () => {
+    const wrapper = await createWrapper();
+    const disconnectButton = wrapper.find('.account-content__disconnect__button');
+
+    expect(disconnectButton.exists()).toBe(true);
+  });
+
+  test('Clicking "Disconnect" button triggers disconnect method', async () => {
+    const wrapper = await createWrapper();
+    (wrapper.vm as any).disconnect = jest.fn();
+    const disconnectButton = wrapper.find('.account-content__disconnect__button');
+
+    disconnectButton.trigger('click');
+    await wrapper.vm.$nextTick();
+
+    expect((wrapper.vm as any).disconnect).toHaveBeenCalled();
+  });
 });
--- a/raiden-dapp/tests/unit/components/account/account-content.spec.ts
+++ b/raiden-dapp/tests/unit/components/account/account-content.spec.ts
@@ -60,6 +60,10 @@ async function createWrapper(
   return wrapper;
 }
 
+const getDisconnectButton = (wrapper: Wrapper<AccountContent>): Wrapper<AccountContent> => {
+  return wrapper.find('.account-content__disconnect__button');
+};
+
 describe('AccountContent.vue', () => {
   beforeEach(() => {
     jest.clearAllMocks();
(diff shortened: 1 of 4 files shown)


raiden-dapp/tests/unit/components/account/account-content.spec.ts
```diff
@@ -200,4 +200,36 @@ describe('AccountContent.vue', () => {
...
+      undefined,
+      undefined,
+      undefined,
+      undefined,
+      false,
+      undefined,
+    );
+    const disconnectButton = wrapper.find('.account-content__disconnect__button');
+
+    expect(disconnectButton.exists()).toBe(false);
```
weilbith, line 215: Is this a mistake? I'm not sure if I need another coffee or if I'm just stupid. :laughing: 
I mean this test works because you explicitly create the wrapper with connection set to `false` but the test case title states something different.

raiden-dapp/tests/unit/components/account/account-content.spec.ts
```diff
@@ -200,4 +200,36 @@ describe('AccountContent.vue', () => {
...
+  });
+
+  test('does not display "Disconnect" button when not connected', async () => {
+    const wrapper = await createWrapper();
```
weilbith, line 219: Nitpick: also this is the default, could you explicitly set the connection status parameter to `true`?  So we have literally the condition of the title guaranteed.
I know setting this parameter is annoying. I guess that is my fault because I have written such ugly wrapper functions. The lack of named parameters annoyed. So I started a new concept for unit tests `createWrapper` functions in my recent PR. Though I don't want to ask you to change this here. Just wanted to say sorry. :see_no_evil:

raiden-dapp/tests/unit/components/account/account-content.spec.ts
```diff
@@ -200,4 +200,36 @@ describe('AccountContent.vue', () => {
...
+    const disconnectButton = wrapper.find('.account-content__disconnect__button');
+
+    expect(disconnectButton.exists()).toBe(true);
+  });
+
+  test('Clicking "Disconnect" button triggers disconnect method', async () => {
+    const wrapper = await createWrapper();
+    (wrapper.vm as any).disconnect = jest.fn();
+    const disconnectButton = wrapper.find('.account-content__disconnect__button');
```
weilbith, line 228: Nitpick: you do this three times. In other unit tests we considered to export such as a function like `getDisconnectButton(wrapper)`. Just judge yourself if you think this is already worth here to add. I guess there are arguments for both here. :shrug:

weilbith (approved):
Very clean feature addition. Nice to read. :+1: 
Just out of curiosity: why did you avoid to use a Vuetify button with the plain property? Eventually would have reduced the amount of HTML and CSS to write as you eventually don't even need the outer div. But I'm actually not too unhappy to ditch Vuetify more often as we started to distance from it. :laughing:

taleldayekh:
> Just out of curiosity: why did you avoid to use a Vuetify button with the plain property?

I wanted the "Disconnect" button to be more in the periphery and a simple text link, kind of similar to the "Privacy Policy & Imprint" that is in the footer.

raiden-dapp/tests/unit/components/account/account-content.spec.ts
taleldayekh: Yeah, I mixed up the test names, will correct it. :)

raiden-dapp/tests/unit/components/account/account-content.spec.ts
taleldayekh: Haha, sure, I can add 5 x `undefined` 😂.

raiden-dapp/tests/unit/components/account/account-content.spec.ts
taleldayekh: Done!

PR #2581, opened 2021-02-22
Adds amount-based fee safety margin

andrevmatos:
Split from #2575 

**Short description**
Adds the ability to set a `[f, a]` tuple as safety margin, where the final fee is calculated as `ceil(estimated_fee + f * estimated_fee + a * amount)`. This is needed in order to reproduce PC's default safety margin of `+3% estiamated_fee + 0.05% amount`.

This also introduces the `mergeWith` Rx operator, which should help simplify some deep-mergeMap operators callback-hell which are used simply to allow previous and current value in a context.

A new dependency, `bignumber.js`, which was already pulled in by some dependency, is now explicitly required and imported as `BN`, in order to properly perform **floating/decimal big number arithmetics**, but it should be used only for the internal operations, interfaces should still expose the simpler bn.js-based `ethers` `BigNumber` and our custom derivative types (`UInt`, `Int`, etc). When rounding, an explicit rounding mode is expected instead of setting it globally. For reference, Python's default `round()` mode is `BN.ROUND_HALF_EVEN`.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-cli/CHANGELOG.md
+++ b/raiden-cli/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
 ## [Unreleased]
+### Changed
+- [#2581] CLI now defaults to `3% * fee + 0.05% * amount` for fee safety margin, same as PC
+
+[#2581]: https://github.com/raiden-network/light-client/pull/2581
 
 ## [0.15.0] - 2021-01-26
 ### Added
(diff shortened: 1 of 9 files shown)


weilbith (approved):
I've shorten this review as I already saw the code in the originally PR where it got extracted from. :grimacing:

PR #2594, opened 2021-02-26
Improve first and non-first initial sync times on testnet and mainnet, reduce number of ETH RPC requests

### LINKED ISSUE #2566: After initial sync, startup should be faster
While testing the startup on mainnet I found that any subsequent start after the initial sync isn't considerably faster. Probably some information isn't stored in the state and re-requested on every start. From what I see the contracts that get queried from the beginning are the `MonitoringService` contract, and the `TokenNetwork` contracts.

My expectation is that starting a node after the initial sync is faster that the initial sync.

### LINKED ISSUE #2569: Fetch the PFS and MS addresses from the global registries
Part of issues for the LC side of https://github.com/raiden-network/raiden/issues/6809

Raiden switches to toDevice communication with the PFS and MS. In order to find the correct matrix user to send it to we need a source of identification. 

PFS are identified by their address in the service registration on-chain. The raiden node should fetch these addresses and store them to construct matrix user ids later on.

### Acceptance Criteria
- After initial sync just re-use what is already there
- Renaming of actions and payload (both pfs and ms) to services

andrevmatos:
Fixes #2566
Fixes #2569 

**Short description**
This PR optimizes some filters for first, initial and on-block syncs/events monitoring.
- `getLogsByChunk$` now can fetch events from arrays of `address`es, allowing a single request to be used to monitor multiple contracts at the same time
- The above is used in order to remove a long standing workaround in code: now, all TokenNetworks are scanned in search of channels with us, to decide on TNs "of interest" to be monitored
  - Besides this, we also always monitors the first 2 TNs found, in order to avoid this re-scan in cases no channels are found, e.g. on mainnet, which did slow down the non-first initial syncs in case no TN got monitored for that client.
- `fromEthersEvent` can now properly report when its first/past event scanning phase is completed, so we can better report it at `synced` promise and `deps.init$` subject
  - Also, it can take a `blockNumber$` observable to use, instead of listening `provider.on('block')`, which does start the block polling. This was done in order to delay this polling until `synced` on `initNewBlockEpic` and passing the `block/new` action-based observable whenever this function is used.
- On TokenNetwork scanning, we now rely on `fromEthersEvent` working properly, and resume `fromBlock` since latest seen block, which reduces by a lot the range of blocks on non-first initial syncs.
- `pfsList` is replaced by `servicesValid`, which is persisted on `state.services` and resumed properly on next initial sync, instead of scanning the whole deployment range again, which was one of the other bottlenecks on initial sync times.
  - a db migration is included to include this property on already initialized states
- waiting for `messageGlobalSend` responses on `init$`/`synced`, waits for init-time `PFSFeeUpdate` global messages to be sent before considering node `synced`
- **TODO on a follow up PR**: monitors all TNs by default using a single filter with all TN's addresses, possibly also using this same filter/request (once per block) to perform all events fetching on the LC, not only TN's. This will reduce as much as possible the number of requests per time-unit on the LC, which will optimize on Infura usecases. Possibly, we can do with a single getLogs request per block!

**Definition of Done**

1. Initialize a CLI & dApp client on a clean state, on testnet and mainnet, and notice the `raiden/synced`'s `payload.tookMs` value;
2. Restart it, and see this time drop to < 5s if there're channels; <1s if there's no channel (due to the global messages)

### DIFF
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -41,6 +41,7 @@ import {
   pluck,
   publishReplay,
   scan,
+  startWith,
   switchMap,
   take,
   takeUntil,
@@ -73,6 +74,7 @@ import { fromEthersEvent, getLogsByChunk$, logToContractEvent } from '../utils/e
 import {
   completeWith,
   distinctRecordValues,
+  lastMap,
   pluckDistinct,
   retryAsync$,
   retryWhile,
@@ -130,6 +132,7 @@ export function initEpic(
         ),
       );
     }),
+    completeWith(state$),
     finalize(() => init$.complete()),
   );
 }
@@ -141,15 +144,22 @@ export function initEpic(
  * @param state$ - Observable of RaidenStates
  * @param deps - RaidenEpicDeps members
  * @param deps.provider - Eth provider
+ * @param deps.init$ - Observable which completes when initial sync is done
  * @returns Observable of newBlock actions
  */
 export function initNewBlockEpic(
   action$: Observable<RaidenAction>,
   {}: Observable<RaidenState>,
-  { provider }: RaidenEpicDeps,
+  { provider, init$ }: RaidenEpicDeps,
 ): Observable<newBlock> {
   return retryAsync$(() => provider.getBlockNumber(), provider.pollingInterval).pipe(
-    mergeMap((blockNumber) => merge(of(blockNumber), fromEthersEvent<number>(provider, 'block'))),
+    // emits fetched block first, then subscribes to provider's block after synced
+    mergeMap((blockNumber) =>
+      init$.pipe(
+        lastMap(() => fromEthersEvent<number>(provider, 'block')),
+        startWith(blockNumber),
+      ),
+    ),
     map((blockNumber) => newBlock({ blockNumber })),
     completeWith(action$),
   );
@@ -592,8 +602,9 @@ function fetchNewChannelEvents$(
   [token, tokenNetwork]: [Address, Address],
   deps: RaidenEpicDeps,
 ) {
-  const { provider, getTokenNetworkContract, config$ } = deps;
+  const { provider, getTokenNetworkContract, config$, latest$ } = deps;
   const tokenNetworkContract = getTokenNetworkContract(tokenNetwork);
+  const blockNumber$ = latest$.pipe(pluckDistinct('state', 'blockNumber'));
 
   // this mapping is needed to handle channel events emitted before open is confirmed/stored
   const channelFilter: Filter = {
@@ -601,12 +612,11 @@ function fetchNewChannelEvents$(
     // set only topics[0], to get also open events (new ids); filter client-side
     topics: [Object.values(getChannelEventsTopics(tokenNetworkContract))],
   };
-
-  return config$.pipe(
-    first(),
-    mergeMap(({ confirmationBlocks }) =>
-      fromEthersEvent<Log>(provider, channelFilter, confirmationBlocks, fromBlock),
-    ),
+  return fromEthersEvent<Log>(provider, channelFilter, {
+    fromBlock,
+    blockNumber$,
+    confirmations: config$.pipe(pluck('confirmationBlocks')),
+  }).pipe(
     map(logToContractEvent<ChannelEvents>(tokenNetworkContract)),
     filter(isntNil),
     mapChannelEventsToAction([token, tokenNetwork], deps),
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -271,41 +271,53 @@ function scanRegistryTokenNetworks({
     filter(([, tokenNetwork]) => !!tokenNetwork),
     toArray(),
     mergeMap((logs) => {
-      if (!logs.length) return EMPTY;
-      const firstBlock = logs[0][2].blockNumber;
-      const tokenNetworks = new Map<string, [token: Address, event: Event]>(
-        logs.map(([token, tokenNetwork, event]) => [tokenNetwork, [token, event]]),
+      const alwaysMonitored$: Observable<tokenMonitored> = from(
+        logs
+          .splice(0, 2)
+          .map(([token, tokenNetwork, event]) =>
+            tokenMonitored({ token, tokenNetwork, fromBlock: event.blockNumber }),
+          ),
       );
-      const allTokenNetworkAddrs = Array.from(tokenNetworks.keys());
-      const aTokenNetworkContract = getTokenNetworkContract(allTokenNetworkAddrs[0] as Address);
-      const ChannelOpenedTopic = Interface.getEventTopic(
-        aTokenNetworkContract.interface.getEvent('ChannelOpened'),
-      );
-      // simultaneously query all tokenNetworks for channels from us and to us
-      return merge(
-        getLogsByChunk$(provider, {
-          address: allTokenNetworkAddrs,
-          topics: [ChannelOpenedTopic, null, encodedAddress], // channels from us
-          fromBlock: firstBlock,
-          toBlock: provider.blockNumber,
-        }),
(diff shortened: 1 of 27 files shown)


raiden-ts/src/utils/ethers.ts
```diff
@@ -46,12 +62,23 @@ export function getLogsByChunk$(
     let retry = 3;
     // every time repeatWhen re-subscribes to this defer, yield (current/retried/next) range/chunk
     return defer(async () =>
-      provider.getLogs({
-        ...filter,
-        fromBlock: start,
-        toBlock: Math.min(start + curChunk - 1, toBlock),
-      }),
+      provider.send('eth_getLogs', [
```
palango, line 65: It would be good to create a feature request for ethers to support this, so e don't have to call and handle this in our code.

raiden-ts/src/channels/epics.ts
```diff
@@ -251,9 +251,69 @@ export function blockStaleEpic(
...
+  }).pipe(
+    map(
+      logToContractEvent<[token: Address, token_network: Address, event: Event]>(registryContract),
+    ),
+    filter(isntNil),
+    filter(([, tokenNetwork]) => !!tokenNetwork),
+    toArray(),
+    mergeMap((logs) => {
+      if (!logs.length) return EMPTY;
+      const firstBlock = logs[0][2].blockNumber;
```
palango: A comment for what `[2]`is would be nice for the future.

raiden-ts/src/channels/epics.ts
```diff
@@ -271,41 +271,53 @@ function scanRegistryTokenNetworks({
...
-      const tokenNetworks = new Map<string, [token: Address, event: Event]>(
-        logs.map(([token, tokenNetwork, event]) => [tokenNetwork, [token, event]]),
-      );
-      const allTokenNetworkAddrs = Array.from(tokenNetworks.keys());
-      const aTokenNetworkContract = getTokenNetworkContract(allTokenNetworkAddrs[0] as Address);
-      const ChannelOpenedTopic = Interface.getEventTopic(
-        aTokenNetworkContract.interface.getEvent('ChannelOpened'),
+      const alwaysMonitored$: Observable<tokenMonitored> = from(
+        logs
+          .splice(0, 2)
```
palango, line 276: why two? For the purpose of remembering the sync one would be enough, so shouldn't we just do this or is this done with thought of current mainnet - so that both tokens there would be synced?

raiden-ts/src/utils/ethers.ts
```diff
@@ -78,54 +82,78 @@ export function fromEthersEvent<T>(
...
+    : typeof confirmations === 'number'
+    ? of(confirmations)
+    : confirmations;
+  const blockQueue: number[] = []; // sorted 'fromBlock' queue, at most of [confirmations * 2] size
+  let start = Date.now();
+  return confirmations$.pipe(
+    distinctUntilChanged(),
+    mergeWith((confirmations) => {
+      if (!fromBlock) {
+        // 'resetEventsBlock' is private, set at [[Raiden]] constructor, so we need 'any'
```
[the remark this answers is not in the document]
andrevmatos: `resetEventsBlock` sets `_lastBlockNumber`, and since that's the public interface, I chose to reference it here, but this comment is removed on current HEAD, and instead I did it by code, using the` resetBlock` var name.

raiden-ts/src/utils/ethers.ts
```diff
@@ -78,54 +82,78 @@ export function fromEthersEvent<T>(
...
- * @param fromBlock - Block since when to fetch events from
+ * @param opts - Options object
+ * @param opts.fromBlock - Block since when to fetch events from
+ * @param opts.confirmations - After how many blocks a tx is considered confirmed; if observable,
+ *    it should have a value at subscription time, like a ReplaySubject(1);
+ * @param opts.blockNumber$ - New blockNumber observable
+ * @param opts.onPastCompleted - Callback when first/past blocks scan completes
  * @returns Observable of target.on(event) events
  */
 export function fromEthersEvent<T>(
```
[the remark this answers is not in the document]
andrevmatos: The upper ones are function overloads, and are a way on TS to state specific function signatures to be picked by the transpiler at type-check time depending on the number and types of passed parameters. Unfortunately, there can be only one actual implementation, so it needs to be compatible (although usually broader/more relaxed) with the overloaded signatures.

raiden-ts/src/utils/ethers.ts
```diff
@@ -78,54 +82,78 @@ export function fromEthersEvent<T>(
...
-          ? resetBlock
-          : innerBlockNumber && innerBlockNumber > 0
-          ? innerBlockNumber
-          : 1;
-      fromBlock = resetBlock - confirmations;
-    }
-    // starts 'blockQueue' with subscription-time's resetEventsBlock
-    blockQueue.splice(0, blockQueue.length, fromBlock);
+  const confirmations$ = !confirmations
+    ? of(5)
```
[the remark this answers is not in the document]
andrevmatos: This is the default value. There isn't a constant, even because this is here mostly for type-safety, since I want to leave it as optional but we specify this param everywhere this function is used. I can pull it to a `DEFAULT_CONFIRMATION_BLOCKS` constant, if you feel it's relevant.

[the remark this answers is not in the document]
palango: If we define it in multiple places then I'd like if you'd define it as a constant.

raiden-ts/src/utils/ethers.ts
```diff
@@ -78,54 +82,78 @@ export function fromEthersEvent<T>(
...
+    ? of(confirmations)
+    : confirmations;
+  const blockQueue: number[] = []; // sorted 'fromBlock' queue, at most of [confirmations * 2] size
+  let start = Date.now();
+  return confirmations$.pipe(
+    distinctUntilChanged(),
+    mergeWith((confirmations) => {
+      if (!fromBlock) {
+        // 'resetEventsBlock' is private, set at [[Raiden]] constructor, so we need 'any'
+        let resetBlock: number = (target as any)._lastBlockNumber;
```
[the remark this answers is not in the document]
andrevmatos: I noticed this was made public already, but we can't affect it since it's on ethers code. I've switched to using the public access variable anyway.

raiden-ts/src/utils/ethers.ts
andrevmatos: I've talked to them in https://github.com/ethers-io/ethers.js/issues/1180 , but it seems they aren't going to upstream this feature, since the code which handles this is on `BaseProvider`, shared by all providers, but some children doesn't support this kind of request, e.g. `EtherscanProvider`. Since we instead require/use `JsonRpcProvider` and children, which does support it, I coded the functionality directly on this utility function which already wraps these `getLogs` calls anyway.

raiden-ts/src/channels/epics.ts
andrevmatos: Fortunately, TS merged recently named tuple parameters, and I've already used it here, on code, no need for comment: if you inspect the `logs` type, you'll see something like `logs: [token: Address, token_network: Address, event: Event][]`, which tells you `[0]` is the first array element, and `[2]` is the `event: Event` member, which is the `ethers`'s `Event` object.

raiden-ts/src/channels/epics.ts
andrevmatos: Yeah, this is a little usecase-centric, so 2 is just arbitrarily chosen. Mainnet is the most obvious, but on Goerli, we also make more extensive usage of the first 2 registered TNs. Since this is a temporary code, and also on a fallback case (the user can `monitorToken` explicitly and if they get a (past or new) channel on one of the other TNs, they'll be monitored anyways), I don't think it's an issue to have this default expected case hardcoded here.

raiden-ts/src/utils/ethers.ts
palango: Cool, thanks a lot!

palango (approved):

PR #2614, opened 2021-03-05
Send service messages toDevice and don't join global service rooms

### LINKED ISSUE #2571: Only join discovery room
Part of issues for the LC side of raiden-network/raiden#6809

Raiden Nodes only need to join discovery rooms once to_device messaging to the services is implemented

### Acceptance Criteria
- Not join the ms and pf rooms
- Ensure that state and events from the global rooms are not being fetch and don't join the rooms

### LINKED ISSUE #2572: Handle communication with services by toDevice messages
Part of issues for the LC side of raiden-network/raiden#6809

The matrix user_id to send to_device messages to can be deducted by the following pattern: 

`@address:localhomeserver`

All services have clients on all homeservers. To bring best performance we only send to the users on the user's homeserver

For MS and PFS there must exist different device id's to differntiate the receiving services. Suggestions can be found [here](https://github.com/raiden-network/raiden-services/issues/916)

###  to_device call content
Multiple messages to multiple users can be attached to the json content. This means that messages can be send to all PFSes with one API call. 

### Acceptance Criteria
- Adapting queue to send to the services list users from the registers (psf/ms).
- Pack all messages to one JSON
- Adjust unit/integration tests

andrevmatos:

Fixes #2571 
Fixes #2572 

**Short description**
Sends service messages to service users in the same federated server (since they're expected to have clients listening on every transport server in the federation), instead of a global room. Therefore, joining such rooms isn't needed anymore.

**Definition of Done**

1. SP pass again
2.

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -8,10 +8,12 @@
 ### Changed
 - [#2536] Wait for global messages before resolving deposits and channel open request
 - [#2566] Optimize initial sync and resume previous sync filters scans
+- [#2572] **BREAKING** Send services messages through `toDevice` instead of global rooms
 
 ### Removed
 - [#2550] **BREAKING** Remove migration of legacy state at localStorage during creation
 - [#2567] **BREAKING** Remove support for peer-to-peer communication through Matrix rooms; now supports only `toDevice` and WebRTC channels.
+- [#2571] **BREAKING** Remove ability to join and send messages to global service rooms
 
 ### Fixed
 - [#2596] Fix unlocking sent transfers even if receiving is disabled
@@ -21,6 +23,8 @@
 [#2550]: https://github.com/raiden-network/light-client/issues/2550
 [#2566]: https://github.com/raiden-network/light-client/issues/2566
 [#2567]: https://github.com/raiden-network/light-client/issues/2567
+[#2571]: https://github.com/raiden-network/light-client/issues/2571
+[#2572]: https://github.com/raiden-network/light-client/issues/2572
 [#2581]: https://github.com/raiden-network/light-client/pull/2581
 [#2596]: https://github.com/raiden-network/light-client/issues/2596
 
--- a/raiden-ts/src/config.ts
+++ b/raiden-ts/src/config.ts
@@ -30,9 +30,6 @@ const RTCIceServer = t.type({ urls: t.union([t.string, t.array(t.string)]) });
  *    transfer expiration block should be
  * - httpTimeout - Used in http fetch requests
  * - discoveryRoom - Discovery Room to auto-join, use null to disable
- * - pfsRoom - PFS Room to auto-join and send PFSCapacityUpdate to, use null to disable
- * - monitoringRoom - MS global room to auto-join and send RequestMonitoring messages;
- *    use null to disable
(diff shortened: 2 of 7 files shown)


raiden-ts/src/raiden.ts
```diff
@@ -303,6 +303,8 @@ export class Raiden {
         error: 'error',
         nextState: 'debug',
       },
+      // disable colors outside of browsers env
+      ...(globalThis?.location ? {} : { colors: false }),
```
weilbith, line 307: I think the quality can be improved by removing the comment and express it as code. So just have something like `const beingInsideBrowser = !!globalThis?.location` and then use it directly in the object assignment. Results into the exact same number of lines but with more value. :grimacing:

raiden-ts/src/helpers.ts
```diff
@@ -556,37 +557,37 @@ export async function getState(
 
 /**
  * For a given channelId (passed as opts.meta), waits for a deposit to be confirmed and for the
- * following messageGlobalSend for its PFSCapacityUpdate to succeed
+ * following messageServiceSend for its PFSCapacityUpdate to succeed
```
weilbith, line 560: This is just a representative line for the whole second commit (96670eb48255218f7d5dce622128612c956ad80d). 
Why was this replacement necessary? I can only guess and argue. But it would be cool if I as a reviewer must not make assumptions but can just follow your decisions. Especially because this commit is not just about renaming but there are some minor logic tweaks. I don't know the exact reason for both... :wink:

weilbith (approved):
Despite two small code quality comments this looks good to me. :ok_hand:

raiden-ts/src/helpers.ts
andrevmatos: Beforehand, the services messages were sent to the respective **global** rooms (and the epics which emitted those actions knew about it). Now, those rooms don't exist anymore, and the epics must instead emit actions specifying to which service they want to send the messages to, and the transport is the one who now knows how to get those messages to the services (now, through toDevice messages to the specific set of userIds from services from `ServicesRegistr`y). The renaming comes with this logic change, because the perspective also changed slightly. From the changes only, it may be a little harder to see that, but if you're reading the actions and code for the first time, hopefully this is what you'll get from the actions signature and usage.

weilbith: Perfect. Next time a version of this in the commit message. Please. :pray: :upside_down_face:

raiden-ts/src/raiden.ts
andrevmatos: :heavy_check_mark:

raiden-ts/src/helpers.ts
andrevmatos: :heavy_check_mark:

andrevmatos:
So I've changed the target branch of this PR to the just created `next`. I'll try to keep rebased on top of `master` (therefore, this new branch's history will be constantly rewritten, so do proper `git reset --hard` and `merge` back and forth on your local working copies if willing to use it). The bulk of development will stay in `master` for now, but we'll try to keep changes there backwards compatible, and I'll be responsible for rebasing `next` accordingly whenever something changes. When this becomes too hard to keep, we will make a master release and merge `next` to it then.

PR #2621, opened 2021-03-12
Support multiple custom services addresses

### LINKED ISSUE #2570: Add additional services addresses via CLI
Part of issues for the LC side of raiden-network/raiden#6809

The Raiden node will only use PFSes and MSes which are registered on-chain. For testing purpose and other reasons we should also enable adding additional addresses if a user wants to use a non-registered PFS.

andrevmatos:
Fixes #2570

**Short description**
This changes the format of `config.pfs` to support receiving an array of addresses or urls as additional/custom services to be used. By default, the first working one will be chosen when not specifying one in `findRoutes` and `transfer`.
The main reason for allowing it is to use this list as complementary Services to be notified when channel state changes (this last part is only relevant on `next`, where `toDevice` messages to specific services are used instead of global room messages), but it also serves on `master` to allow multiple services to be tried when doing an auto PFS request, instead of rejecting when the first fails.
The interface on cli now allows multiple parameters be passed to `--pathfindingServiceAddress`, which already served the purpose of specifying a single url, so this change is backwards compatible.
Most of this PR is intended for `master`, but the `next:` commit is going to be removed when merging this to `master` and then that commit will be pushed to `next`.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-cli/src/index.ts
+++ b/raiden-cli/src/index.ts
@@ -7,7 +7,7 @@ import * as path from 'path';
 import yargs from 'yargs/yargs';
 
 import type { Decodable, RaidenConfig } from 'raiden-ts';
-import { Address, assert, Capabilities, decode, Raiden, UInt } from 'raiden-ts';
+import { Address, assert, Capabilities, decode, PfsMode, Raiden, UInt } from 'raiden-ts';
 
 import { makeCli } from './cli';
 import DEFAULT_RAIDEN_CONFIG from './config.json';
@@ -121,9 +121,9 @@ function parseArguments() {
         desc: 'Anything else than "pfs" disables mediated transfers',
       },
       pathfindingServiceAddress: {
-        type: 'string',
-        default: 'auto',
-        desc: 'Force a given PFS to be used; "auto" selects cheapest registered on-chain',
+        type: 'array',
+        desc:
+          'Force given PFS URL list to be used, automatically chosing the first responding provider for transfers, instead of auto-selecting valid from "ServiceRegistry" contract',
       },
       pathfindingMaxPaths: {
         type: 'number',
@@ -296,9 +296,14 @@ function createRaidenConfig(
 
   if (argv.matrixServer !== 'auto') config = { ...config, matrixServer: argv.matrixServer };
 
-  if (argv.routingMode !== 'pfs') config = { ...config, pfs: null };
-  else if (argv.pathfindingServiceAddress !== 'auto')
-    config = { ...config, pfs: argv.pathfindingServiceAddress };
+  if (argv.routingMode !== 'pfs') config = { ...config, pfsMode: PfsMode.disabled };
+  else if (!argv.pathfindingServiceAddress) config = { ...config, pfsMode: PfsMode.auto };
+  else
+    config = {
+      ...config,
+      pfsMode: PfsMode.onlyAdditional,
+      additionalServices: argv.pathfindingServiceAddress as string[],
+    };
 
   if (!argv.enableMonitoring) config = { ...config, monitoringReward: null };
 
--- a/raiden-ts/src/config.ts
+++ b/raiden-ts/src/config.ts
@@ -33,8 +33,8 @@ const RTCIceServer = t.type({ urls: t.union([t.string, t.array(t.string)]) });
  * - pfsRoom - PFS Room to auto-join and send PFSCapacityUpdate to, use null to disable
  * - monitoringRoom - MS global room to auto-join and send RequestMonitoring messages;
  *    use null to disable
- * - pfs - Path Finding Service URL or Address. Set to null to disable, or empty string to enable
- *    automatic fetching from ServiceRegistry.
+ * - pfs - Array of Path Finding Service Addresses (require PFS to be registered) or URLs.
+ *    Set to false to disable, or true to enable automatic fetching from ServiceRegistry.
  * - pfsSafetyMargin - Safety margin to be added to fees received from PFS. Either a fee
  *    multiplier, or a [fee, amount] pair ofmultipliers. Use `1.1` to add a 10% over estimated fee
  *    margin, or `[0.03, 0.0005]` to add a 3% over fee plus 0.05% over amount.
@@ -70,7 +70,7 @@ export const RaidenConfig = t.readonly(
       discoveryRoom: t.union([t.string, t.null]),
       pfsRoom: t.union([t.string, t.null]),
       monitoringRoom: t.union([t.string, t.null]),
-      pfs: t.union([Address, t.string, t.null]),
+      pfs: t.union([t.readonlyArray(t.union([Address, t.string])), t.boolean]),
       pfsSafetyMargin: t.union([t.number, t.tuple([t.number, t.number])]),
       pfsMaxPaths: t.number,
       pfsMaxFee: UInt(32),
@@ -146,7 +146,7 @@ export function makeDefaultConfig(
     discoveryRoom: `raiden_${networkName}_discovery`,
     pfsRoom: `raiden_${networkName}_path_finding`,
     monitoringRoom: `raiden_${networkName}_monitoring`,
-    pfs: '', // empty string = auto mode
+    pfs: true, // true = auto mode
     pfsSafetyMargin: 1.0, // multiplier
     pfsMaxPaths: 3,
     pfsMaxFee: parseEther('0.05') as UInt<32>, // in SVT/RDN, 18 decimals
--- a/raiden-ts/src/config.ts
+++ b/raiden-ts/src/config.ts
@@ -6,6 +6,7 @@ import type { Observable } from 'rxjs';
 import { first } from 'rxjs/operators';
 
 import { Capabilities, DEFAULT_CONFIRMATIONS } from './constants';
+import { PfsMode, PfsModeC } from './services/types';
 import { exponentialBackoff } from './transfers/epics/utils';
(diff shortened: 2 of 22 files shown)


weilbith:
> By default, the first working one will be chosen when not specifying one in findRoutes and transfer

Quick question: how does this work for the interface function to query manually a nodes presence state as we use it in the dApp? 
Maybe I'll find out in the code.

raiden-ts/src/config.ts
```diff
@@ -65,7 +65,7 @@ export const RaidenConfig = t.readonly(
       expiryFactor: t.number, // must be > 1.0
       httpTimeout: t.number,
       discoveryRoom: t.union([t.string, t.null]),
-      pfs: t.union([Address, t.string, t.null]),
+      pfs: t.union([t.readonlyArray(t.union([Address, t.string])), t.boolean]),
```
weilbith, line 68: First of all I like you general approach for this configuration option. But I would argue it could be improved and simplified at the same time. The name `pfs` is pretty unspecific, but the type behind it has a not trivial documentation. Without loosing any functionality, what do you think about making this just of type `t.readonlyArray(t.union([Address, t.string])`? Then we can somehow (sorry that I fail to come with a suggestion) rename the option that expresses that this list are manual/additional/exceptional services. And if this list is empty we just use the automated procedure.
What this approach does not configure is the case what to do if none of the configured services services is working. I would argue to just fail as the user explicitly wants to use these services (because of reasons?) and do not continue to auto-detect services. Having a second option to specify this behavior would be more precise or rather let the user decide instead of us. The downside here is just that we need to explain the co-interaction between these two options and which is has priority in which case. But I would argue that isn't worse what we have now for this single very complex option.
This are my thoughts/suggestions. What do you think?

PS: I think the following code somehow shows that the handling of such a complex single configuration options is quite cumbersome. Especially with such a flat name as the other parts of the code have no such easy access to the documentation of it.

raiden-ts/src/services/utils.ts
```diff
@@ -103,42 +91,64 @@ export function pfsInfo(
...
+ * @param pfsAddrOrUrl - PFS account/address or URL
+ * @param deps - RaidenEpicDeps needed for various parameters
+ * @param deps.serviceRegistryContract - ServiceRegistry contract instance
+ * @param deps.network - Current Network
+ * @param deps.contractsInfo - ContractsInfo mapping
+ * @param deps.config$ - Config observable
+ * @param deps.provider - Eth provider
+ * @returns Observable containing PFS server info
+ */
+export const pfsInfo: typeof pfsInfo_ = memoize(
```
weilbith, line 146: The info object for a PFS does contain some variable data like. I'm lacking knowledge, but how would we get updated about such properties like the prize?

raiden-cli/src/index.ts
```diff
@@ -121,9 +121,9 @@ function parseArguments() {
         desc: 'Anything else than "pfs" disables mediated transfers',
       },
       pathfindingServiceAddress: {
-        type: 'string',
-        default: 'auto',
-        desc: 'Force a given PFS to be used; "auto" selects cheapest registered on-chain',
+        type: 'array',
+        desc:
+          'Force given PFS URL list to be used, instead of using best registered from "ServiceRegistry" contract',
```
weilbith, line 126: This might get quite long, but I would add that they are sorted by priority and the first available gets used. Right?

weilbith (commented):
I basically just want to discuss the configuration parameters. Else it looks very good.

I would like to thank you for your recent improvements. In my eyes your commit size and messages have become very good and it helps me a lot to review your PRs! I'm very thankful that you took the feedback. :hugs: I think it will help everyone and the whole team gets better. 
 
We have some serious issues in regards of diffs. It cost so much time to compare these big diffs where only a single line differs. Can we somehow harden this in our formatters or is this because of our tubed reactive functions.

raiden-ts/src/config.ts
andrevmatos: How would it be if the user wants to disable PFS?
I'm not a big fan of putting all these behaviors on this same config, but since they're somehow exclusive (false=disabled, true=auto, `(Address | string)[]`), I did it like that. An alternative would be to have an explicit list of `additionalServices`, then have the current option be an enum or a value which would go like `disabled`, `auto` (use any/best between additional+auto, additional are also notified) and `onlyAdditional` (notify everyone, use only one of the explicitly set value). It'd be more explicit, but I'm not sure the additional verbosity and complexity of two options are worth it instead of just allowing it be `false|true`.

raiden-ts/src/services/utils.ts
andrevmatos: Currently, it isn't, at least on a single session, we just assume PFS isn't changing that often, but we could have some ttl cache.

raiden-ts/src/config.ts
weilbith: Hmm okay. I'm not 100% happy with how it is, but I like your suggestion. Though I don't see an obvious path to go for here. So if you don't feel the need to change then close this conversation and keep it as is. :+1:

raiden-ts/src/services/utils.ts
weilbith: So this actually means the user could confirm on the dApp to use a PFS with a prize that is not up to date anymore?

raiden-ts/src/config.ts
```diff
@@ -65,7 +68,8 @@ export const RaidenConfig = t.readonly(
       expiryFactor: t.number, // must be > 1.0
       httpTimeout: t.number,
       discoveryRoom: t.union([t.string, t.null]),
-      pfs: t.union([t.readonlyArray(t.union([Address, t.string])), t.boolean]),
+      additionalServices: t.readonlyArray(t.union([Address, t.string])),
```
weilbith, line 71: Can this also include monitoring services? :thinking: This makes no sense doesn't it? But eventually when ditch Matrix completely, we also wanna configure other services? So maybe make it explicit now and name it `additionalPfsServices` so no matter what comes later we are more stable for future extensions.

raiden-ts/src/config.ts
```diff
@@ -65,7 +68,8 @@ export const RaidenConfig = t.readonly(
...
+      pfsMode: PfsModeC,
```
weilbith, line 72: Nitpick & voluntary: `pfsSelectMode` Being more precise now might help us later if we need to extend the configuration for reasons we can't know yet. (ofc this goes also for the type name if you decide to do so)

weilbith (commented):
I basically just reviewed 4e438511579d247329de884fb4432f98f474f636 as it includes the changes for the configuration as we have discussed. l'm really happy about this decision. Just naming nitpicks to be more open for future changes.

raiden-ts/src/config.ts
weilbith: We discussed in a personal call and agreed on a solution which got now implemented in 4e438511579d247329de884fb4432f98f474f636

raiden-cli/src/index.ts
weilbith: @andrevmatos please close this conversation independently when it is done. I don't see a reason to very this once done. :grimacing:

raiden-ts/src/config.ts
andrevmatos: it can include monitoring services. While MS doesn't expose an explicit API, it can use the same account as the PFS (it actually is a bundle) and we need to fetch the `address` for which to send `messageServiceSend` to `service: MS`. That's why it was named generically, because it can serve both, the registry is for both and the URL provides the address of the account for both (the difference is in the `deviceId`)

weilbith: Ah okay. Interesting. So then this configuration split change is actually even more beneficial. :+1:

raiden-ts/src/services/utils.ts
andrevmatos: It could be higher than you wanted, but you would have reviewed and agreed with that higher value (either interactively on dApp or on CLI through `--pathfindingMaxFee` option), so I'd consider this a race which is not common and reasonable to be accepted. In case it's too small, it may fail until the cache expires, but next PFS should be tried, I need to check this though or implement retrying with clean cache in case of this specific error, which can also be considered a race (although a up to 1h long one, currently).

andrevmatos: Alternatively, we could implement a dedicated function to request this endpoint (instead of reusing `pfsInfo`) which would cache only `address`, which is for what we care about this (for sending messages through toDevice on `next`).

weilbith: Maybe we could discuss this in person again? Your last comment sounds quite interesting. Else I would say this a too big and critical feature addition which deserves its own PR and should not just be added without a note.

andrevmatos: I've implemented it. Again, had to rebase due to the required strict order of the `next:` commit. `pfsInfo` isn't cached anymore, but sets the cache for `address`. `pfsInfoAddress` is the provided cached util, which uses `pfsInfo` and extracts only the `address` (which doesn't change) but doesn't prevent any following `pfsInfo` call to perform a new request. This should take the best of both cases.

weilbith: Sounds reasonable for me. Good job. 👍🏾

weilbith (approved):
**Final review**
Looks good to me. Thanks for your recent adoptions to my requests. 🙏🏾

andrevmatos:
Rebased on top of current `master`, the `next:` commit will be pushed independently with a rebase of `next` when this gets merged

PR #2622, opened 2021-03-15
devops: bump python client version for e2e image

weilbith:
Hoping to test the dApp's end-to-end tests. :crossed_fingers: 
@taleldayekh will know.


weilbith:
I fixed the issue of @taleldayekh which was preventing the e2e tests to run if the user does not already has the image locally. This was a stupid bug introduced by the recent changes.
Furthermore I added bunch of improvements to harden the whole image version usage, for building and running in combination. This should help us to never run tests wrongly and be confused by the result or overwrite an image unexpected or forget to update the version number.

weilbith:
I promise one day I'll find a way to finally improve this Docker image and make the build fast **and** deterministic (!!).

andrevmatos:
Ok, I found the culprit: https://github.com/raiden-network/raiden/issues/6767
So, [this commit](https://github.com/raiden-network/raiden/commit/ab76088ff854f6eb2d90e92460745b42d175f7ed#diff-c1109378458c932815eca2a9bc2f7d28f5bbd3fac73921fd853dc0edb8391a45) sets the PC's `deviceId` to `RAIDEN` (from `raiden`), and went on **after** `1.2.0` release, and is required for the current LC implementation which sends `toDevice` messages specifically to that device_id.

The fix is to either bump the PC version in the image to some commit after that, or to go back to sending `toDevice` messages to `'*'` in the SDK (which got set to the stricter deviceId on a5c4194dc2d69f1d89b0edecb31b0bedfa84ce76 as part of #2579)

weilbith:
@andrevmatos Could you do me a quick favor and amend the last commit with a better commit message. Basically just a short version of what you told us in the meeting. And if you quickly mention the verification by the Matrix team this would be superb. Afterwards I would continue to just merge this PR.

weilbith:
As this is moly devops only plus a minor SDK part, I would argue we don't need to extend the CHANGELOGS and can ignore this failing hook.

PR #2628, opened 2021-03-18
Allow SDK to detect and work without WebRTC API

### LINKED ISSUE #2600: dApp mobile: Make sure WebRTC works as intended
## User Story

As Chris I want to be able to rely on WebRTC working for both iOS and Android and if not I want make sure I can fallback on the toDevice.

## Acceptance criteria

WebRTC and toDevice works the same way as in the current dApp.

## Tasks

- [ ] Determine how to with certainty verify that WebRTC and toDevice works as intended.
- [ ] See what needs to be done to include WebRTC in capacitor.

andrevmatos:
Fixes #2600

**Short description**
The SDK now doesn't auto-polyfill `wrtc` for RTC support, since it doesn't wor on all environments.
Projects running the SDK on environments without native WebRTC support (e.g. NodeJS) now are expected to provide a compatible API, e.g. by dropping this line early in some script:
```js
Object.assign(globalThis, require('wrtc'));
```
In exchange, the SDK should work fine on environments which don't support this framework.
Nothing changes for environments natively supporting RTC, e.g. modern browsers.

**Definition of Done**

1. Comment out CLI's polyfill expression in the beginning of `index.ts` and re-build it
2. Run the CLI and check `webRTC` capability is turned off and CLI still can perform payments over `toDevice` messages only

### DIFF
--- a/raiden-cli/src/index.ts
+++ b/raiden-cli/src/index.ts
@@ -15,6 +15,11 @@ import DISCLAIMER from './disclaimer.json';
 import type { Cli } from './types';
 import { setupLoglevel } from './utils/logging';
 
+if (!('RTCPeerConnection' in globalThis)) {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  Object.assign(globalThis, require('wrtc'));
+}
+
 function parseFeeOption(args: readonly string[]) {
   assert(args.length && args.length % 2 === 0, 'fees must have the format [address, number]');
   const res: { [addr: string]: string } = {};
--- a/raiden-cli/package.json
+++ b/raiden-cli/package.json
@@ -41,6 +41,7 @@
     "node-localstorage": "^2.1.6",
     "raiden-ts": "*",
     "rxjs": "^6.6.6",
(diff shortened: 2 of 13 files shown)


raiden-cli/src/index.ts
```diff
@@ -15,6 +15,11 @@ import DISCLAIMER from './disclaimer.json';
 import type { Cli } from './types';
 import { setupLoglevel } from './utils/logging';
 
+if (!('RTCPeerConnection' in globalThis)) {
```
weilbith, line 18: Do we actually need this condition? This should never exist in a Node environment, shouldn't it?

weilbith (approved):
Very nice little commits to make this small, but impact-full change easy to follow. 👍🏾

raiden-cli/src/index.ts
andrevmatos: It's improbable, but possible to use this script in a browser runtime, or imported from another script, so I wanted to avoid the double polyfill in case that happens, just in case, but normally, Node will never have WebRTC ootb

PR #2635, opened 2021-03-24
Interactive UDC withdraw

### LINKED ISSUE #2629: SDK - split user deposit withdrawal into plan and actual withdraw actions
## Description
Related to: #2612

The Python client has recently also added an implementation to handle user deposit interactions. Unfortunately their behavior differs from our already existing one. One part is the automated withdrawal of planned withdraws. We always do this automated and this can't be disabled or done differently. To make the CLI strictly follow the API specification we need to do the following:

## Acceptance criteria
- configuration option to toggle automatic withdrawals (true per default, false for the CLI)
- split withdraw action into two actions and have two epics for them
- add interface function to make a final withdraw (fails if plan is not "ready")

## Tasks
- [ ]

andrevmatos:
Fixes #2629 

**Short description**
Refactor: split services/epics into separate modules for UDC handling, MS and PFS.
`udcWithdraw` is split into 2 sets of independent async actions: `udc/withdraw/plan`, which is responsible only for the planning transaction, per user request, and triggered from `Raiden.planUdcWithdraw` public method, and `udc/withdraw`, which triggers the final step, after 100 blocks of the planned tx, which got exposed through `Raiden.udcWithdraw` method.
Additionally, the current plan (if any) can be checked from the new `Raiden.getUdcWithdrawPlan` method, which will resolve to `undefined` if there's none, or to an object containing `amount`, `block` (block at which it'll become ready) and `ready` (boolean to tell if it's already ready).
A new `config.autoUdcWithdraw`, which defaults to true, enables an epic which will watch the blockchain and automatically `udc/withdraw/request` when a plan is ready; it can be turned off, and only then `Raiden.udcWithdraw` is made available for the interactive 2nd step udcWithdraw.

**Definition of Done**

1.
2.


raiden-ts/src/raiden.ts
```diff
@@ -1388,13 +1388,74 @@ export class Raiden {
...
   }
 
+  /**
+   * Fetches our current UDC withdraw plan
+   *
+   * @returns Promise to object containing maximum 'amount' planned for withdraw and 'block' at
+   *    which withdraw will become available, and 'ready' after it can be withdrawn with
+   *    [[udcWithdraw]]; resolves to undefined if there's no current plan
+   */
+  public async getUdcWithdrawPlan(): Promise<
```
palango, line 1398: We should use consistent spelling here.

We have `getUDCCapacity` and `depositToUDC` with all-caps UDC, and then `planUdcWithdraw` and `getUdcWithdrawPlan`.

andrevmatos: I agree, and maybe also either rename `depositToUdc` to `udcDeposit` or name `udcWithdraw` to `withdrawFromUdc`; we just need to be careful, since those are already part of the public API, that's why I just played along here. Let's discuss this.

palango: I like  `udcWithdraw` -> `withdrawFromUdc` more.

And I guess renaming it now before any stable release it better than doing it later.

raiden-ts/src/services/epics/udc.ts
```diff
@@ -280,42 +281,60 @@ export function udcWithdrawPlanRequestEpic(
...
-      }),
-    ),
-    filter((value) => value.withdraw_block.gt(Zero)),
-    map(({ amount, withdraw_block }) =>
-      udcWithdrawPlan.success(
-        { block: withdraw_block.toNumber(), confirmed: true },
-        { amount: amount as UInt<32> },
+  { userDepositContract, address, config$, log }: RaidenEpicDeps,
+): Observable<udcWithdraw.request> {
+  const WITHDRAW_TIMEOUT = 100;
```
weilbith, line 292: Random line of this function...
I'm not too satisfied by this approach with the blocks and the timeout. Why can't we simply listen for the withdraw plan success actions as trigger? There we can also read-out the blocks to wait. And if I'm not mistaken these actions are re-emitted after each restart, aren't they?

raiden-ts/src/services/epics/udc.ts
```diff
@@ -324,54 +343,40 @@ export function udcCheckWithdrawPlannedEpic(
...
-        ]);
-        return contract.withdraw(action.meta.amount);
-      }).pipe(
+      let balance: UInt<32>;
+      return retryAsync$(
+        async () => userDepositContract.callStatic.balances(address),
+        intervalFromConfig(config$),
+        { onErrors: networkErrors, log: log.info },
+      ).pipe(
+        mergeMap(async (balance_) => {
```
weilbith, line 354: This appears to be a more often used pattern. Is it worth to export this? 🤔

raiden-ts/src/services/epics/udc.ts
```diff
@@ -324,54 +343,40 @@ export function udcCheckWithdrawPlannedEpic(
...
+          assert(balance_.gt(Zero), [
```
weilbith, line 355: Why not just extending this check with `balance._ge(action.meta.amount)`? 🤔

raiden-ts/src/raiden.ts
weilbith: I totally agree with Paul. The `depositTo` and `withdrawFrom` sound better for function names. The others read more like getters or so. They do not include a verb.

raiden-ts/src/raiden.ts
```diff
@@ -1388,6 +1388,36 @@ export class Raiden {
...
+      amount: plan.amount as UInt<32>,
+      block: plan.withdraw_block.toNumber(),
+      ready: plan.withdraw_block.lte(this.state.blockNumber),
+    };
+  }
+
+  /**
+   * Records a UDC withdraw plan for our UDC deposit
+   *
+   * The plan will be ready for withdraw after 100 blocks.
```
weilbith, line 1413: I would remove the exact number here. Depending on which deployment we use this could theoretically differ. It would be better to keep the SDK independent. For example in future we could have test network deployments with a way shorter plan period time than on production. Or just imagine e2e test deployments.

raiden-ts/src/raiden.ts
```diff
@@ -1399,6 +1429,37 @@ export class Raiden {
     return promise;
   }
 
+  /**
+   * Complete a planned UDC withdraw and get the deposit to account.
+   *
+   * Maximum 'value' is the one from current plan, attempting to withdraw a larger value will throw
```
weilbith, line 1435: The value of a plan can actually not shrink, can't it? So how would we handle a follow up withdraw? Would the transaction just fail for wrong parameters as our code can't catch this early?

weilbith (changes requested):
Thanks for taking the time and implementing this feature. What sound to easy in the beginning actually became quite big.
I'm pretty happy with most of it. Just the epic to auto-trigger withdrawals is not so nice as I think. We should discuss it together. Especially the hard binding to the exact contract deployment constructor value is not so nice. I would like to get rid of it.

raiden-ts/src/raiden.ts
andrevmatos: This value is not configurable by deployment options, it's hardcoded, it is exactly 100 blocks, and must be, or else it'd not be this contract, but a different one. It's not a parameter, it's a constant.

raiden-ts/src/raiden.ts
andrevmatos: The plan amount can't shrink, but you request to withdraw anything in `0 < amount <= plan.amount` range, that's what I've tried to describe on the docs and conditions.

raiden-ts/src/services/epics/udc.ts
```diff
@@ -88,27 +88,36 @@ function makeUdcDeposit$(
...
+      retryAsync$(
+        async () => userDepositContract.callStatic.total_deposit(address),
+        pollingInterval,
+        { onErrors: networkErrors },
+      ).pipe(
+        mergeMap(async (deposited) => {
+          assert(deposited.add(deposit).eq(totalDeposit), [
+            ErrorCodes.UDC_DEPOSIT_OUTDATED,
+            { requested: totalDeposit.toString(), current: deposited.toString() },
+          ]);
```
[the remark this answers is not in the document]
andrevmatos: Codeclimate was failing, complaining about this because it was skipped before, and since I've split and moved this code around, it triggered again. It also used some outdated patterns on the retry, so it was reasonable to adapt and improve those while making Codeclimate happy.

raiden-ts/src/utils/error.ts
```diff
@@ -48,6 +51,42 @@ export function matchError(match: ErrorMatch | ErrorMatches, error?: any) {
...
+ * @param opts - Options object
+ * @param opts.maxRetries - maximum number of retries before rejecting
+ * @param opts.onErrors - retry only on these errors
+ * @param opts.neverOnErrors - Never retry on these errors
+ * @param opts.predicate - Retry if this predicate matches
+ * @param opts.stopPredicate - Don't retry if this predicate match
+ * @param opts.log - Log using this function
+ * @returns Function to test if errors should be retried or not
+ */
+export function shouldRetryError(opts: {
```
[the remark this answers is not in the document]
weilbith: Andre don't want to have this. Therefore closing.

raiden-ts/src/services/epics/udc.ts
weilbith: We had a call to this topic. With his good explanation this algorithm satisfies me.

raiden-ts/src/services/epics/udc.ts
weilbith: Will not be handled.

raiden-ts/src/services/epics/udc.ts
weilbith: Will not be handled.

weilbith: It will then just fail at the node.

raiden-ts/src/raiden.ts
weilbith: We discussed that. Keep it if you like it.

raiden-ts/src/raiden.ts
weilbith: Okay, Andre explained me that the plan will get deleted by each withdraw. So this is perfect fine.

raiden-ts/src/utils/error.ts
```diff
@@ -48,6 +51,42 @@ export function matchError(match: ErrorMatch | ErrorMatches, error?: any) {
...
+  readonly maxRetries?: number;
+  readonly onErrors?: ErrorMatches;
+  readonly neverOnErrors?: ErrorMatches;
+  readonly predicate?: PredicateFunc;
+  readonly stopPredicate?: PredicateFunc;
+  readonly log?: LogFunc;
+}) {
+  let count = -1;
```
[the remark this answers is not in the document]
weilbith: Alright. Andre told me how this works with contexts here. I'm too much into low level stuff atm...

weilbith (approved):
Thanks for the long call! 🙏🏾 
Here the awaited approval. Good work.

raiden-ts/src/utils/error.ts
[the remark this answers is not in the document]
andrevmatos: Just for reference, JS function calls create contexts, which are objects bound to any function returned. It works like an "instance", and each time you call the returned function, it'll refer to that specific context when it got called. If this gets called multiple times, reassigned to the same or a different variable, those contexts are independent and initialized at the parent's function call time.

raiden-ts/src/utils/error.ts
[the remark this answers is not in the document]
andrevmatos: Yeah, I couldn't come up with a better name; tried to express in the docs that the decision function is the returned one, but I'd be open for a better name.

raiden-ts/src/services/epics/udc.ts
andrevmatos: I'll move this constant to the `constants.ts` file and remove the docstrings references.

raiden-ts/src/services/epics/udc.ts
andrevmatos: Thanks for the headsup.. actually, the first `retryAsync$` is not needed since it's also covered by the `retryWhile` there, so I'll simplify this code a little bit more. About the pattern, it can be not very practical to split, since this references the context variable (`balance`), and also usually have specific error messages and contexts which would be more verbose to pass all in a custom operator call than to do it like that. But we can re-evaluate this if the pattern spreads more.

raiden-ts/src/raiden.ts
andrevmatos: :heavy_check_mark:

raiden-ts/src/raiden.ts
andrevmatos: Removed from the docstring, moved the constant to `constants.ts` :heavy_check_mark:

raiden-ts/src/services/epics/udc.ts
andrevmatos: :heavy_check_mark:

raiden-ts/src/services/epics/udc.ts
andrevmatos: :heavy_check_mark:

PR #2642, opened 2021-03-29
Feature/cli add udc endpoints

### LINKED ISSUE #2612: Add `user_deposit` endpoint to CLI
## Description
For interacting with the UDC, I added a `/user_deposit` endpoint recently to the python client's API. To follow up the light client should implement it as well.

From the original PR (https://github.com/raiden-network/raiden/pull/6821):
> It has three possibilities for POST requests depending on the body parameter:
> - deposit: `{total_deposit: <amount>}`
> - plan a withdraw: `{planned_withdraw_amount: <amount>}`
> - withdraw: `{withdraw_amount: <amount>}`

All return a `transaction_hash` after the tx was minted. The planning of a withdrawal additionally returns the number of blocks until the funds can be withdrawn (`planned_withdraw_block_number`).

For more details see the docs: https://raiden-network.readthedocs.io/en/latest/rest_api.html#user-deposit-for-services

## Acceptance criteria
- Implement the endpoint as in the documentation.

## Tasks
- [ ] Add new endpoint: `/user_deposit`
- [ ] Add deposit method
- [ ] Add plan a withdrawal method
- [ ] Add withdraw method

weilbith:
Fixes #2612

**Short description**
This PR has to deal a lot with the incredibly ugly API specification we have to follow. I tried my best to keep the code simple, flat and easy to read. Though I'm really not proud of the constant error instances to handle those new special cases where the CLI has to bridge the different approaches between SDK and PC for the deposit handling. I'm open for any better solution that is still super simple. At the moment it is a single route local issue. I hope this does not start to metastasize on future API extensions.

**Definition of Done**

- [ ] Change has been manually tested by the reviewer

**Steps to manually test the change**

Checkout the [API spec](https://raiden-network.readthedocs.io/en/latest/rest_api.html#user-deposit-for-services) to get example request and see the different error cases the server needs to handle. Try at least one complete cycle of deposit, increase deposit, increase deposit too low, plan too high withdraw, plan withdraw, withdraw to early, withdraw to high, withdraw finally, withdraw again, ...

### DIFF
--- a/raiden-cli/src/routes/userDeposit.ts
+++ b/raiden-cli/src/routes/userDeposit.ts
@@ -106,6 +106,16 @@ async function determineAndExecuteRequestedInteraction(
   }
 }
 
+async function getUDCInfo(this: Cli, _request: Request, response: Response): Promise<void> {
+  const balance = await this.raiden.getUDCCapacity();
+  const totalDeposit = await this.raiden.getUDCTotalDeposit();
+
+  response.json({
+    balance: balance.toString(),
+    total_deposit: totalDeposit.toString(),
+  });
+}
+
 async function getUDCWithdrawPlan(
   this: Cli,
   _request: Request,
@@ -148,8 +158,9 @@ async function interactWithUDC(this: Cli, request: Request, response: Response):
 export function makeUserDepositRouter(this: Cli): Router {
   const router = Router();
 
-  router.get('/withdraw_plan', getUDCWithdrawPlan.bind(this));
+  router.get('/', getUDCInfo.bind(this));
   router.post('/', interactWithUDC.bind(this));
+  router.get('/withdraw_plan', getUDCWithdrawPlan.bind(this));
 
   return router;
 }
--- a/raiden-cli/src/routes/userDeposit.ts
+++ b/raiden-cli/src/routes/userDeposit.ts
@@ -57,10 +57,10 @@ async function deposit(this: Cli, totalDeposit: string): Promise<DepositResponse
 
 async function planWithdraw(this: Cli, amount: string): Promise<PlanWithdrawResponse> {
   const transactionHash = await this.raiden.planUDCWithdraw(amount);
-  const plannedWithdrawBlockNumber = (await this.raiden.getBlockNumber()) + 95; // TODO: !!!
+  const withdrawPlan = await this.raiden.getUDCWithdrawPlan();
   return {
     transaction_hash: transactionHash,
-    planned_withdraw_block_number: plannedWithdrawBlockNumber,
+    planned_withdraw_block_number: withdrawPlan!.block, // The plan must exists here.
   };
 }
 
--- a/raiden-cli/src/routes/userDeposit.ts
+++ b/raiden-cli/src/routes/userDeposit.ts
@@ -1,15 +1,31 @@
-import { BigNumber } from 'ethers';
+import { BigNumber, constants } from 'ethers';
 import type { Request, Response } from 'express';
 import { Router } from 'express';
 
-import { ErrorCodes } from 'raiden-ts';
-
 import type { Cli } from '../types';
-import { isInsuficientFundsError, isInvalidParameterError } from '../utils/validation';
+import {
+  isConflictError,
+  isInsuficientFundsError,
+  isInvalidParameterError,
+} from '../utils/validation';
 
 type DepositResponse = { transaction_hash: string };
 type PlanWithdrawResponse = { transaction_hash: string; planned_withdraw_block_number: number };
 type WithdrawResponse = { transaction_hash: string };
+
+const MalformedNumberValue = new Error('A provided number value was not decodable');
+const MalformedRequestFormat = new Error('The provided JSON is in some way malformed');
+const TotalDepositTooLowError = new Error(
+  'The provided total_deposit is not higher than the previous total_deposit',
+);
+
+function parseAsBigNumber(input: unknown): BigNumber {
+  try {
+    return BigNumber.from(input);
+  } catch {
+    throw MalformedNumberValue;
+  }
+}
 
 function getTotalDepositAmount(request: Request): string | undefined {
   return request.body.total_deposit?.toString();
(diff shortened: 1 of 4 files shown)


raiden-cli/package.json
```diff
@@ -31,6 +31,7 @@
     "typescript": "^4.2.3"
   },
   "dependencies": {
+    "@ethersproject/constants": "^5.0.10",
```
andrevmatos, line 34: If possible, I think it´d be good to not add the internal projects as dependencies for the user-facing projects; you can access `constants` from `ethers` import directly

raiden-cli/src/routes/userDeposit.ts
```diff
@@ -0,0 +1,151 @@
...
+import type { Cli } from '../types';
+import {
+  isConflictError,
+  isInsuficientFundsError,
+  isInvalidParameterError,
+} from '../utils/validation';
+
+type DepositResponse = { transaction_hash: string };
+type PlanWithdrawResponse = { transaction_hash: string; planned_withdraw_block_number: number };
+type WithdrawResponse = { transaction_hash: string };
```
andrevmatos, line 15: Please, for interfaces we control (i.e. are not enforced by external APIs), let´s keep consistent variable and members naming and use camelCase for those; we may re-introduce the linting rule later on, but for now we just need to keep an eye on this

raiden-cli/src/routes/userDeposit.ts
```diff
@@ -0,0 +1,151 @@
...
+  return (
+    getTotalDepositAmount(request) === undefined &&
+    getPlannedWithdrawAmount(request) === undefined &&
+    getWithdrawAmount(request) !== undefined
+  );
+}
+
+async function deposit(this: Cli, totalDeposit: string): Promise<DepositResponse> {
+  const totalDepositTarget = parseAsBigNumber(totalDeposit);
+  const currentTotalDeposit = await this.raiden.getUDCCapacity();
```
andrevmatos, line 69: Do we expose the current UDC capacity/deposit somewhere? If not, even if not specified by the requirements, I think we could/should, maybe just a `GET` to the main endpoint returning a simple JSON, otherwise it may be hard for the user to calculate how much they should send to this API. I think the WebUI access this directly, but we don´t want to depend on that.

Besides that, I just noticed this may be a problem, the deposit is expected to receive a `totalDeposit`, but we don´t expose that value on the SDK methods, only the capacity, which may be misleading when dealing with total deposit. We may need to expose and adjust to that if we want to be correct and consistent with the current reference implementation/usage (webUI).

andrevmatos (commented):
it looks good to me; there´s just some points we need to sync to have this the best way it can be (as far as the current api requirements allow)

raiden-cli/src/routes/userDeposit.ts
andrevmatos: Scratch that, should have finished reading the file before submitting this one, it´s obviously enforced by the external reference.

raiden-cli/src/routes/userDeposit.ts
```diff
@@ -0,0 +1,151 @@
...
+}
+
+/**
+ * @param this - Cli object
+ * @returns Router instance
+ */
+export function makeUserDepositRouter(this: Cli): Router {
+  const router = Router();
+
+  router.get('/withdrawPlan', getUDCWithdrawPlan.bind(this));
```
andrevmatos, line 147: Is this endpoint also provided like that by the PC API? If not, I think we could/should keep consistency on the naming and use snake_case for the endpoint paths

raiden-cli/src/routes/userDeposit.ts
```diff
@@ -0,0 +1,151 @@
...
+    throw MalformedRequestFormat;
+  }
+}
+
+async function getUDCWithdrawPlan(
+  this: Cli,
+  _request: Request,
+  response: Response,
+): Promise<void> {
+  const withdrawPlan = await this.raiden.getUDCWithdrawPlan();
```
andrevmatos, line 115: You need to call `toString` on the `amount`, it´s serializing the BigNumber as an object

raiden-cli/src/routes/userDeposit.ts
andrevmatos: #2644 exposes a `Raiden.getUDCTotalDeposit` method, which should allow you to expose a `GET` endpoint (possibly informing both `balance` and `total_deposit`) and base `deposit` amount on `total_deposit` instead of capacity, which should be more correct.

raiden-cli/package.json
weilbith: ✔️

raiden-cli/src/routes/userDeposit.ts
weilbith: ✔️

raiden-cli/src/routes/userDeposit.ts
weilbith: ✔️

weilbith:
I'll no rebase on `master` to get the recent SDK feature that enables us to add the total deposit endpoint as requested [here](https://github.com/raiden-network/light-client/pull/2642#discussion_r603651570). Thereby I will already integrate the fixup commits. But all adoptions to request remain visible in the history.

raiden-cli/src/routes/userDeposit.ts
weilbith: ✔️ 
I also added a get capacity endpoint because why not...

weilbith:
We still some issues with very long pending transactions. We need to check that.

raiden-cli/src/routes/userDeposit.ts
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * @param this - Cli object
+ * @returns Router instance
+ */
+export function makeUserDepositRouter(this: Cli): Router {
+  const router = Router();
+
+  router.get('/withdraw_plan', getUDCWithdrawPlan.bind(this));
+  router.get('/total_deposit', getUDCTotalDeposit.bind(this));
+  router.get('/capacity', getUDCCapacity.bind(this));
```
andrevmatos, line 167: Nitpick, as a personal taste, feel free to ignore if you don´t agree: do we need 2 endpoints for those? I think `router.get('/')` returning a `{ total_deposit, balance }` object would be better.
It´s better to call it `balance` here, since this is the naming used in the webUI and PC REST API where we usually call it capacity. Also, most/all APIs return objects instead of plain numbers directly, so returning an object with the two values would be better to keep consistency. Finally, I think this is simple enough to be a single endpoint. Even though those are async requests, they´re fast as they fetch it from stored value in the state machine.

andrevmatos (approved):
lgtm, feel free to implement or dismiss the nitpick comment as you wish. thanks

raiden-cli/src/routes/userDeposit.ts
weilbith: Hmm. I thought about it but I rejected it. But if you know also have the same idea it is maybe not so bad. I just do it as I have no strong opinion here.

weilbith: ✔️

weilbith:
I replaced the last two commits with the separate endpoints with a single one as requested [here](https://github.com/raiden-network/light-client/pull/2642#discussion_r604883110)

PR #2646, opened 2021-03-30
Wait for condition to be ready on settleChannel and withdrawFromUDC

### LINKED ISSUE #2645: Wait for condition to be ready on settleChannel and withdrawFromUDC
@andrevmatos [This API](https://github.com/raiden-network/light-client/blob/62ffe722cd9143953f9a24e8c7dcab425b6796b5/raiden-ts/src/raiden.ts#L1431-L1459) is a bit counter-intuitive for me. I expected it to wait until it can to the withdraw and execute it, but it fails when the withdraw plan is not "ready" yet.
Addition: It's kind of the same for the `settle` call. I think we should either provide a `waitUntilSettleable` method or make `settleChannel` itself wait.

_Originally posted by @palango in https://github.com/raiden-network/light-client/pull/2643#discussion_r604013175_

These are the requests which have an `auto` config: channel settlement and 2nd phase of UDC withdraw. Currently, if they're not in auto-mode, and the user request the action before it's ready (channel is settleable, udc withdraw plan timeout passed), they fail. We should instead wait for these inevitable (block timeout based) conditions to be met instead and then proceed with requesting.

## Acceptance criteria
- [x] `settleChannel` waits for channel to become settleable before requesting channel settlement instead of erroring
- [x] `withdrawFromUDC` waits for UDC withdraw plan to be ready before requesting 2nd phase of UDC withdraw

andrevmatos:
Fixes #2645 

**Short description**
These methods are only available if the `autoSettle` and `autoUDCWithdraw` configs aren't set. In these cases, calling it before it's ready (timeout blocks have passed) would error. This PR makes an early call wait until ready instead of erroring.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/helpers.ts
+++ b/raiden-ts/src/helpers.ts
@@ -15,6 +15,7 @@ import {
   filter,
   first,
   map,
+  mergeMap,
   pluck,
   take,
   takeWhile,
@@ -23,8 +24,10 @@ import {
 } from 'rxjs/operators';
 
 import type { RaidenAction } from './actions';
+import type { channelSettle } from './channels/actions';
 import { channelDeposit } from './channels/actions';
-import type { RaidenChannel, RaidenChannels } from './channels/state';
+import type { Channel, RaidenChannel, RaidenChannels } from './channels/state';
+import { ChannelState } from './channels/state';
 import { channelAmounts, channelKey } from './channels/utils';
 import type { RaidenConfig } from './config';
 import {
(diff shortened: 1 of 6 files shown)


raiden-ts/src/helpers.ts
```diff
@@ -590,3 +593,36 @@ export async function waitForPFSCapacityUpdate(
...
   );
 }
+
+const settleableStates = [ChannelState.settleable, ChannelState.settling] as const;
+const preSettleableStates = [ChannelState.closed, ...settleableStates] as const;
+
+/**
+ * Waits for channel to become settleable
+ *
+ * Errors if channel doesn't exist or isn't closed, settleable ot settling (states which precede
```
palango, line 603: s/ot/or

palango (approved):
Great! Thanks for implementing this quickly. One small typo to fix though.

raiden-ts/src/helpers.ts
andrevmatos:
```suggestion
 * Errors if channel doesn't exist or isn't closed, settleable or settling (states which precede
```

PR #2655, opened 2021-04-07
dapp: upgrade Capacitor to version 3

### LINKED ISSUE #2654: dApp - upgrade Capacitor to version 3
This upgrade is necessary for the following issue. When we uploaded the first internal test releases to the Google Play Store, we could not. The reason was that the app requires sensible permissions to get camera and microphone access. But the dApp is actually only using the camera. Though the origin issue comes from the Capacitor core package. It does not allow for treeshaking and comes with many components we don't use. But these component ask the user for certain permissions when used. When Google scan the application bundle it detects these refernces. And for sensible permissions the application must come with a data security/data protection document bundled with it. We don't have such yet. But we also only need the camera as a nice to have feature at a later point in time.

For first experiments (#2648) it was enough to change the generated Capacitor code to remove the unwanted permissions. That made a first internal test release of the app possible. But it is still the goal to prevent us having any manual adoptions of the generated code. We don't want to get into this hell. Therefore I was searching for alternatives. And the alternative is Capacitor version 3. It has purified the core module and outsourced everything else to official plugins that can be added when necessary. Thereby the only permission that gets requested is the necessary internet connection.

It must be noted that version 3 is still in an early stage on not stable. As their is no obvious other solution to the problem and first tests have shown that it works, this is still the solution we go for. It is eventually not a bad idea to take the most recent version in this early stage of the mobile app bundling. So we don't set up everything for the "old" version and then have to update later with possibly much.

weilbith:
Fixes #2654

First manual tests look good. We need end-to-end test... 😑 
When we discover that any permission is missing while we use the app we just add it then.
Important is to check that the build (remove `./android` folder and run `yarn capacitor:init:android` again) in the manifest `./android/app/src/AndroidManifest.xml` does not include any permissions than internet (at the bottom of the file, highlighted with comment).

We need to check how this conflicts with @palango his changes to use SQLite. My hope is that because his plugin targets Cordova directly that it does not care about this upgrade...

### DIFF
--- a/raiden-dapp/.gitignore
+++ b/raiden-dapp/.gitignore
@@ -1,7 +1,7 @@
 .DS_Store
 node_modules
 /dist
-
+.gradle/
 /tests/e2e/results/
 
 # local env files
--- a/raiden-dapp/capacitor.config.json
+++ b/raiden-dapp/capacitor.config.json
@@ -1,7 +1,1 @@
-{
-  "appId": "com.brainbot.raiden.dapp",
-  "appName": "Raiden dApp",
-  "bundledWebRuntime": false,
-  "webDir":"dist",
-  "linuxAndroidStudioPath": "/usr/bin/android-studio"
-}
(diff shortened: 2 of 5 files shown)


palango (approved):
Looks good, and permissions only list internet currently.

weilbith:
Perfect. 👌🏾 
@palango are you now testing your SQLite stuff on top of this? 🤔

palango:
> Perfect. 👌🏾
> @palango are you now testing your SQLite stuff on top of this? 🤔

Yes

PR #2657, opened 2021-04-07
dapp: init Fastlane for android beta release

### LINKED ISSUE #2656: dapp - intialize Fastlane for first Android beta track release
## Description

For the moment we will focus on the Android platform. Reasons are that it is machine independent, plus we don't have yet access to the new Apple appstore account. As this includes the full initial setup this should be as small as possible. 
Note that @taleldayekh and I did already some first experiments in #2654 These include a manual (via web-browser) first release (internal test track). This was necessary as for Android, Fastlane can't do the first release ever. 

Unfortunately we are currently stuck. The keystore used to sign the first application bundle was meant to be used for testing only. Therefore it was a weak one and got deleted in the process. Unfortunately does the Google Play Console expect us to use the same key for ever. Jacob and I are in contact with the customer support and they will add our new key. This can take up to 48h.

This setup should not include any CI stuff yet. While automated with Fastlane, the releases would be still triggered manually on the console yet. But as they should target the beta release track, this doesn't matter anyway.

weilbith:
Fixes #2656

The upload key is not yet authorized. Thereby Fastlane will fail on the last step, complaining about the wrong signature.
Documentation, especially for the secrets/keyfiles will follow soon.


weilbith:
Please note that this also requires #2655 as else the app gets denied due to sensitive permission usage without data protection document.

weilbith:
Okay, the review process with @taleldayekh took part offline. He is now able to run it locally. The occurred issues during this process has led to documentation improvements in the last commits. 
@taleldayekh could you have a final look and give your approval?

taleldayekh (approved):
Thank you for adding the missing documentation and your final comment summarises it all well. Prefect! 👍🏼

weilbith:
I'll resolve the conflicts.

PR #2666, opened 2021-04-09
Add native sqlite backend for mobile

### LINKED ISSUE #2601: Mobile dApp: Introduce the SQL lite plugin
## User Story

As Chris I want to be able to rely on IndexdDB and local storage working as intended on iOS and Android.

## Acceptance criteria

Saving to IndexedDB or local storage should never be an issue on mobile and behave as in the current dApp.

## Tasks (split each task into separate issues)

- [x] Do a deeper research on why we should not rely on it working in its current state.
- [ ] Introduce the SQL lite plugin.

palango:

Fixes #2601

**Short description**

Adds the `cordova-sqlite-storage` plugin to be able to use native sqlite storage on mobile.

**TODOs**
- [x] Only tested on iOS emulator so far
- [ ] Only enable this on mobile builds

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-dapp/package.json
+++ b/raiden-dapp/package.json
@@ -24,6 +24,7 @@
   },
   "dependencies": {
     "compare-versions": "^3.6.0",
+    "cordova-sqlite-storage": "^6.0.0",
     "core-js": "^3.8.3",
     "ethereum-blockies-base64": "^1.0.2",
     "ethers": "^5.0.31",
@@ -31,6 +32,8 @@
     "idb": "^6.0.0",
     "lodash": "^4.17.20",
     "loglevel": "^1.7.1",
+    "pouchdb": "^7.2.2",
+    "pouchdb-adapter-cordova-sqlite": "^2.0.8",
     "raiden-ts": "*",
     "rxjs": "^6.6.3",
     "tiny-async-pool": "^1.2.0",
--- a/raiden-dapp/package.json
+++ b/raiden-dapp/package.json
@@ -23,6 +23,7 @@
     "serve:no-logging": "export NODE_ENV=nologging && vue-cli-service serve"
   },
   "dependencies": {
+    "@capacitor/core": "^3.0.0-rc.0",
     "compare-versions": "^3.6.0",
     "cordova-sqlite-storage": "^6.0.0",
     "core-js": "^3.8.3",
@@ -57,7 +58,6 @@
     "@babel/preset-typescript": "^7.12.16",
     "@capacitor/android": "^3.0.0-rc.0",
     "@capacitor/cli": "^3.0.0-rc.0",
-    "@capacitor/core": "^3.0.0-rc.0",
     "@capacitor/ios": "^3.0.0-rc.0",
     "@cypress/code-coverage": "^3.9.2",
     "@cypress/webpack-preprocessor": "^5.6.0",
--- a/raiden-dapp/src/services/raiden-service.ts
+++ b/raiden-dapp/src/services/raiden-service.ts
@@ -50,14 +50,22 @@ export default class RaidenService {
     stateBackup?: string,
     subkey?: true,
   ): Promise<Raiden> {
-    const { default: adapterPlugin } = await import('pouchdb-adapter-cordova-sqlite');
-    PouchDB.plugin(adapterPlugin);
-
-    const storageOpts = {
+    let storageOpts:
+      | { state: string | undefined }
+      | { state: string | undefined; adapter: string; iosDatabaseLocation: string } = {
       state: stateBackup,
-      adapter: 'cordova-sqlite',
-      iosDatabaseLocation: 'Library',
     };
+    // When running in capacitor, enable native sqlite plugin
+    if (process.env.VUE_APP_DB_ADAPTER === 'cordova-sqlite') {
+      const { default: adapterPlugin } = await import('pouchdb-adapter-cordova-sqlite');
+      PouchDB.plugin(adapterPlugin);
+
+      storageOpts = {
+        state: stateBackup,
+        adapter: 'cordova-sqlite',
+        iosDatabaseLocation: 'Library',
+      };
+    }
 
     try {
       const contracts = await ConfigProvider.contracts();
--- a/raiden-dapp/src/services/raiden-service.ts
+++ b/raiden-dapp/src/services/raiden-service.ts
@@ -1,3 +1,4 @@
+import { Capacitor } from '@capacitor/core';
 import type { BigNumber, BigNumberish, providers } from 'ethers';
 import { constants, utils } from 'ethers';
 import PouchDB from 'pouchdb';
@@ -52,7 +53,12 @@ export default class RaidenService {
   ): Promise<Raiden> {
     let storageOpts:
       | { state: string | undefined }
-      | { state: string | undefined; adapter: string; iosDatabaseLocation: string } = {
+      | {
+          state: string | undefined;
+          adapter: string;
+          location: string;
+          iosDatabaseLocation: string;
+        } = {
       state: stateBackup,
     };
     // When running in capacitor, enable native sqlite plugin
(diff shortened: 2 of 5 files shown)


raiden-dapp/package.json
```diff
@@ -24,13 +24,16 @@
...
     "compare-versions": "^3.6.0",
+    "cordova-sqlite-storage": "^6.0.0",
     "core-js": "^3.8.3",
     "ethereum-blockies-base64": "^1.0.2",
     "ethers": "^5.0.31",
     "fontsource-roboto": "^4.0.0",
     "idb": "^6.0.0",
     "lodash": "^4.17.20",
     "loglevel": "^1.7.1",
+    "pouchdb": "^7.2.2",
```
palango, line 35: Any help why this is necessary? Shouldn't this be included from the SDK?

But without this I got:
```
error: 'pouchdb' should be listed in the project's dependencies. Run 'npm i -S pouchdb' to add it (import/no-extraneous-dependencies) at src/services/raiden-service.ts:3:21:
  1 | import type { BigNumber, BigNumberish, providers } from 'ethers';
  2 | import { constants, utils } from 'ethers';
> 3 | import PouchDB from 'pouchdb';
```

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -49,14 +50,21 @@ export default class RaidenService {
     stateBackup?: string,
     subkey?: true,
   ): Promise<Raiden> {
+    const { default: adapterPlugin } = await import('pouchdb-adapter-cordova-sqlite');
```
palango, line 53: @weilbith Do we already have a way to decide whether or not we're building for mobile?

raiden-dapp/package.json
weilbith: If you import this directly in the dApp, then yes you should follow this. You should not rely on sub-dependencies of you dependencies. They can change and then your package fails suddenly. It is a dependency, so that's it.

raiden-dapp/src/services/raiden-service.ts
weilbith: We have a dedicated environment for it. We could make the adapter a configuration. And then handle it conditionally. This would theoretically allow to even set other handlers later without effort.

palango: > We could make the adapter a configuration. And then handle it conditionally.

With that you mean enabling is based on some environment variable?

weilbith: I would not make it a boolean but rather just put the adapter name into a variable. If this variable is defined, do all the stuff you added with this variable as "generic". If not (default) just skip it and let the SDK do its default stuff.

palango:
@weilbith Could you try this on Android please?

On iOS I get logged output whenever something is written to the DB over the plugin. That makes it easy to see if it works:
```
To Native Cordova ->  SQLitePlugin backgroundExecuteSqlBatch SQLitePlugin163522230 ["options": [{
    dbargs =     {
        dbname = "raiden_goerli_0x5a5CF4A63022F61F1506D1A2398490c2e8dfbb98_0xFb7790eD9fa3231Fc9AdD4EDd292C65De7BfC4e4_0";
    };
    executes =     (
                {
            params =             (
                "{\"id\":\"state.blockNumber\",\"rev_tree\":[{\"pos\":1,\"ids\":[\"3eca986c474cae8f8c08177d80465edc\",{\"status\":\"available\"},[[\"631aa25294b3e3eeb0780700e2d2a602\",{\"status\":\"available\"},[[\"97fde76d899d0f6110619d2936e8e5c9\",{\"status\":\"available\"},[[\"0c006868e7617d7f63b779da02cc7b10\",{\"status\":\"available\"},[[\"619bf1b63a9d2ecd4ec13f3a35f8011f\",{\"status\":\"available
...\"},[[\"69c55a1bd43fc86ba274c66ebaa54dc5\",{\"status\":\"available\"},[[\"85e7c16ec6850c4b407e7fcd8f4e98aa\",{\"status\":\"available\"},[[\"7a82af1f5f08accf1ee85172b5d681eb\",{\"status\":\"available\"},[[\"9b58412373f0830b9e6d708831ef6c01\",{\"status\":\"available\"},[[\"7cbdf5ec6ae33845d334d8439082acf1\",{\"status\":\"available\"},[[\"d0726327cabeebe42b899aee44fbcd71\",{\"status\":\"available\"},[[\"80d9da42eb3a90489d18c13a426c2f23\",{\"status\":\"available\"},[[\"3a9280913ee1035246eb18936c4fb0ec\",{\"status\":\"available\"},[[\"744636f165874a4af5787444284d2fce\",{\"status\":\"available\"},[[\"e1e514d58b166e681c59fb2460b9fb44\",{\"status\":\"available\"},[]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]]}],\"seq\":58}",
                58,
                "42-e1e514d58b166e681c59fb2460b9fb44",
                "state.blockNumber"
            );
            sql = "UPDATE 'document-store' SET json=?, max_seq=?, winningseq=(SELECT seq FROM 'by-sequence' WHERE doc_id='document-store'.id AND rev=?) WHERE id=?";
        }
    );
}]]
```

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -49,14 +50,29 @@ export default class RaidenService {
...
     subkey?: true,
   ): Promise<Raiden> {
+    let storageOpts:
+      | { state: string | undefined }
+      | { state: string | undefined; adapter: string; iosDatabaseLocation: string } = {
+      state: stateBackup,
+    };
+    // When running in capacitor, enable native sqlite plugin
+    if (process.env.VUE_APP_DB_ADAPTER === 'cordova-sqlite') {
+      const { default: adapterPlugin } = await import('pouchdb-adapter-cordova-sqlite');
```
andrevmatos, line 60: Just a headsup: this may cause webpack or other bundlers to always include this dependency, because they don´t know this codepath isn´t reachable. There should be some way to tell them to tree-shake this out depending on the platofmr/env vars at build time, etc. I got this issue in the past with the SDK, where an `await import` triggered some devDependency to be required as peer-dependency of projects using it, so may be worth it to check if there´s an easy way to `#ifdef` this out

palango: Is there an example for this? I cannot find anything in the codebase...

andrevmatos: No, this is up for research.

weilbith:
For documentation purpose:
I tried it and it does not work for me. I contacted Paul to make sure I made no mistake. We even tried a quick fix in attempt to solve it. But unfortunately it still does not work.

raiden-dapp/src/services/raiden-service.ts
```diff
@@ -52,19 +53,33 @@ export default class RaidenService {
   ): Promise<Raiden> {
     let storageOpts:
       | { state: string | undefined }
-      | { state: string | undefined; adapter: string; iosDatabaseLocation: string } = {
+      | {
+          state: string | undefined;
+          adapter: string;
+          location: string;
+          iosDatabaseLocation: string;
```
andrevmatos, line 60:
```suggestion
          location?: string;
          iosDatabaseLocation?: string;
```

raiden-dapp/src/services/raiden-service.ts
palango: I checked the `dist` sizes on master and this branch for a normal `yarn build´. They are equal, so I guess we don't need to handle this now:
```
➜  raiden-dapp git:(master) ✗ du -h dist
868K	dist/css
3.6M	dist/js
 60K	dist/img/icons
280K	dist/img
3.4M	dist/fonts
8.2M	dist




➜  raiden-dapp git:(sqlite-plugin) ✗ du -h dist
868K	dist/css
3.4M	dist/js
 60K	dist/img/icons
280K	dist/img
3.4M	dist/fonts
8.1M	dist
```

palango:
As mentioned [here](https://github.com/raiden-network/light-client/pull/2666#discussion_r616875620) this doesn't increase code size for the web target.

However, there is some suboptimal configuration regarding the capacitor build. This is tracked in https://github.com/raiden-network/light-client/issues/2678

raiden-ts/src/helpers.ts
```diff
@@ -525,11 +525,11 @@ export async function getState(
   ].join('_');
 
   let db;
-  let { state: dump } = storage;
-  const { adapter, prefix } = storage;
+  const { state: stateDump, ...opts } = storage;
```
andrevmatos, line 528: Weren't these changes already committed? Maybe a rebase is needed.

andrevmatos (approved):
lgtm, thank you

raiden-ts/src/helpers.ts
palango: This branch branched from master way before your change. I just cherry-picked it in.

Issue #2669, opened 2021-04-09
Adapt unstable contracts


taleldayekh:
## Description

Related to https://github.com/raiden-network/light-client/issues/2514

We now have `unstable` and `development` contracts.

## Acceptance criteria
- 

## Tasks
- [ ] Make `unstable` contracts selectable in LC
- [ ] Add `development-environment` option in CLI
- [ ] Update deployments
  - [ ] `stable` uses *demo* contracts
  - [ ] `unstable` uses *unstable* contracts

Issue #2514, opened 2021-01-20
Update demo env


taleldayekh:
## Description
This needs to be done for the PFS suggested partners to work. @eorituz can you help here? Is it important to have this feature in the demo env for now?

Needs more discussions and not only in the LC team.

We are not ditching it. Split it by contracts or link by transport servers are the alternatives.
Will the RSB be released?

## Acceptance criteria
- Conceptualise the problem

## Tasks
- [ ] Bump PFS to latest version
- [ ] Wait for @fredo PFS changes to be done before this upgrade

taleldayekh:
@eorituz, would you be ok with Andrés last comment in the "ditch demo env" issue? #2278

eorituz:
Sounds like a good idea to me. 
@ezdac and the TransCore team took over the maintenance of the demoenvs. @ezdac What do you think?

palango:
We now have a separated demo environment. The usage is documented here: https://hackmd.io/q5oULgo6Qzuj5jBTC_E5zw

The LC by default uses the test environment.
- https://lightclient.raiden.network/staging/ uses the test environment
- https://lightclient.raiden.network/ uses the demo environment

manuelwedler:
So we have a second deployment of the contracts on goerli, is this correct? This makes it necessary to update the Explorer and maybe the Wizard.

Is the PFS of the demo environment updated?

palango:
> Is the PFS of the demo environment updated?

No, the demo env kept the current version. But there is a [new issue](https://github.com/raiden-network/raiden/issues/6964) for updating the versions used in the demo env.

PR #2670, opened 2021-04-12
Unstable contracts

palango:

Fixes https://github.com/raiden-network/light-client/issues/2669
Closes https://github.com/raiden-network/light-client/issues/2514

**Short description**

Makes it possible to use the unstable/demo contracts from the raiden contracts package.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-cli/src/index.ts
+++ b/raiden-cli/src/index.ts
@@ -118,8 +118,6 @@ function parseArguments() {
         default: '127.0.0.1:5001',
         desc: 'host:port to bind to and listen for API requests',
       },
-    })
-    .options({
       routingMode: {
         choices: ['pfs', 'local', 'private'] as const,
         default: 'pfs',
--- a/raiden-dapp/.env.production
+++ b/raiden-dapp/.env.production
@@ -1,5 +1,6 @@
 VUE_APP_PFS=https://pfs.demo001.env.raiden.network
 VUE_APP_MATRIX_SERVER=https://transport.demo001.env.raiden.network
+VUE_APP_UDC_ADDRESS=0x0794F09913AA8C77C8c5bdd1Ec4Bb51759Ee0cC5
 VUE_APP_HUB=hub.raiden.eth
 VUE_APP_ALLOW_MAINNET=false
 VUE_APP_IMPRINT=https://raiden.network/privacy.html
--- a/raiden-dapp/src/services/raiden-service.ts
+++ b/raiden-dapp/src/services/raiden-service.ts
@@ -57,7 +57,7 @@ export default class RaidenService {
         {
           state: stateBackup,
         },
-        contracts,
+        process.env.VUE_APP_UDC_ADDRESS ?? contracts,
         {
           pfsSafetyMargin: 1.1,
           ...(process.env.VUE_APP_PFS
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Changed
+- [#2669] Update to Raiden contracts `v0.37.5`
+
+[#2669]: https://github.com/raiden-network/light-client/issues/2669
+
 ## [0.16.0] - 2021-04-01
 ### Added
 - [#1342] Flat (fixed) mediation fees for mediator nodes
--- a/raiden-ts/scripts/versions.js
+++ b/raiden-ts/scripts/versions.js
@@ -5,7 +5,7 @@ const cwd = path.dirname(fs.realpathSync(__filename));
 
 // Get the smart contract version from the `raiden-contracts`
 // sub module.
-const contracts_version = '0.37.1';
+const contracts_version = '0.37.5';
 // TODO: Enable when https://github.com/raiden-network/raiden-contracts/issues/1287 done
 /*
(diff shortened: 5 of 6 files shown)


raiden-ts/src/raiden.ts
```diff
@@ -288,6 +289,7 @@ export class Raiden {
     config?: Decodable<PartialRaidenConfig>,
     subkey?: true,
     subkeyOriginUrl?: string,
+    devEnvironment?: string,
```
palango, line 292: It doesn't feel nice to add this here, as this more devops related than SDK related. 

@andrevmatos Do you see any nicer way? We could parse the contract infos in the CLI and pass them, but then that code would be duplicated.

raiden-ts/src/types.ts
```diff
@@ -87,3 +87,8 @@ export enum EventTypes {
   DEPOSITED = 'DEPOSITED',
   CONFIRMED = 'CONFIRMED',
 }
+
+export enum DevEnvironment {
```
palango, line 91: Is this the right place for this?

raiden-ts/src/raiden.ts
andrevmatos: I don´t like it. Passing yet another variable just to choose between deployments seems not a good public API, specially because it´s specific to a network (goerli) and there are already an interface for arbitrary contracts to be fetched: passing UDC address.
CLI or dApp don´t need to parse everything, it can just have the unstable UDC address hardcoded in some `.env` file, env var or `config.json` (maybe conditional to some other envvar).

raiden-cli/package.json
```diff
@@ -31,6 +31,7 @@
     "typescript": "^4.2.4"
   },
   "dependencies": {
+    "@ethersproject/providers": "^5.1.0",
```
andrevmatos, line 34: Please, don´t import the subprojects from ethers on dapp/cli. This can be accessed from `ethers` instead, which is already available.

raiden-cli/src/index.ts
```diff
@@ -368,6 +373,19 @@ async function checkDisclaimer(accepted?: boolean): Promise<void> {
...
 
+async function getUdcAddressIfUnstable(
+  ethRpcEndpoint: string,
+  developmentEnvironment: string,
+): Promise<string | undefined> {
+  const provider = new JsonRpcProvider(ethRpcEndpoint);
+  const network = await provider.getNetwork();
+
+  if (network.name === 'goerli' && developmentEnvironment === 'unstable') {
+    return '0xD145252daB9323F9Acb70295d6a0589c611F225d';
```
andrevmatos, line 384: Wouldn´t this better go in `config.json`, and then you just use it if the `developmentEnvironment` is unstable?

raiden-cli/src/index.ts
```diff
@@ -368,6 +373,19 @@ async function checkDisclaimer(accepted?: boolean): Promise<void> {
   assert(accepted, 'Disclaimer not accepted!');
 }
 
+async function getUdcAddressIfUnstable(
+  ethRpcEndpoint: string,
+  developmentEnvironment: string,
+): Promise<string | undefined> {
+  const provider = new JsonRpcProvider(ethRpcEndpoint);
```
andrevmatos, line 380: I´d prefer to not have yet another provider instantiated here. Two possible alternatives: either assume unstable implies goerli, and let it crash if the user asked for unstable but isn´t connected to goerli, since this would be an user error; or we could finally mind that `--network-id` param we´ve been neglecting so far and validate it from there.

andrevmatos (approved):
lgtm! thank you for the patience, I like it =)

palango:
> lgtm! thank you for the patience, I like it =)

I agree. this is a way nicer solution than the first try.

drselump14:
Hello, In my understanding this PR replaced the `deployment_goerli.json` with `deployment_goerli_unstable.json`. But I can't find the `deployment_goerli_unstable.json`. Should I rename the deployment_goerli.json to deployment_goerli_unstable.json in my local environment? Otherwise, the build is broken rn

andrevmatos:
Hello, @drselump14 . `deployment_goerli_unstable.json` is part of the new pinned commit on the `raiden-contracts` submodule. If you can't find it, you may need to `git submodule sync && git submodule update --init` to ensure you got the submodule updated as well. No need to copy anything, these changes will work out of the box on both environments.

drselump14:
@andrevmatos Hi, thanks for the explanation. I did the submodule update and it successfully build the code. Thanks!
Now, I have another errors

```
These dependencies were not found:

* ethers/constants in ../raiden-ts/dist/transfers/epics.js
* ethers/utils in ../raiden-ts/dist/transfers/epics.js, ../raiden-ts/dist/transport/epics.js
```

Do you know  how to fix this issue?

palango:
> Do you know how to fix this issue?

Did you rebuild the sdk after the submodule update? 

```sh
 yarn workspace raiden-ts build
```

andrevmatos:
@drselump14 You may have some leftover files in your dist folders. Try a `yarn clean` then `yarn build` inside `raiden-ts` folder.

drselump14:
Ah, sorry for the unclear question. The errors come from this command `yarn workspace raiden-dapp serve` . `yarn workspace raiden-ts build` works fine

Issue #2604, opened 2021-03-03
dApp mobile: List necessary UI adjustments for making dApp look good on "all" screens


taleldayekh:
## Description

The UI for the mobile version of the dApp should not break in either functionality or layout. **Please write you name next to the title in the comment with the UI parts you are working on**.

## Acceptance criteria
- The mobile dApp works and looks good on iOS and Android.

## ToDo
- [ ] Make the dApp look good from screen size **320x568** and up

## Tasks
- [x] Test on different screen sizes and browsers (look up css break points for mobile)
- [ ] Safe zone margins
- [ ] Create new issues if bigger layout changes needs to be made.

palango:
## iOS emulator

Results looks quite well to me

![Simulator Screen Shot - iPhone 12 - 2021-03-03 at 15 56 11](https://user-images.githubusercontent.com/90851/109834561-9dad5400-7c42-11eb-9e01-685b3b552b81.png)
![Simulator Screen Shot - iPhone 12 - 2021-03-03 at 15 56 54](https://user-images.githubusercontent.com/90851/109834575-a1d97180-7c42-11eb-84ad-75415345d270.png)
![Simulator Screen Shot - iPhone 12 - 2021-03-03 at 15 57 35](https://user-images.githubusercontent.com/90851/109834578-a2720800-7c42-11eb-8322-2b7b698b8573.png)
![Simulator Screen Shot - iPhone 12 - 2021-03-03 at 15 55 34](https://user-images.githubusercontent.com/90851/109834583-a3a33500-7c42-11eb-8028-33a3ab1f3867.png)
![Simulator Screen Shot - iPhone 12 - 2021-03-03 at 15 54 36](https://user-images.githubusercontent.com/90851/109834586-a4d46200-7c42-11eb-8065-9c16e2cc71be.png)

taleldayekh:
Looks the same for me, so wonder how come the layout was overlapping on the initial run.

agatsoh:
I tested the dapp on a bunch of different mobiles and tabs viewport the only real screen where I encountered some problem is the screen which you see below (and this was also due to low resolution low viewport mobile). My laptop is fortunately touch screen I didnt realise that earlier but I have tried this using my laptop in touch screen mode and the buttons are mostly responsive.
In lower resolution or lower viewports there is a scroll which is introduced which does not break or introduce any wierd CSS glitches when the keyboard is brought up which is a pretty good sign.


[...]

PR #2679, opened 2021-04-21
dApp: Fixing broken info overlays on mobile

taleldayekh:
**Short description**

This PR takes care of all broken info overlays on mobile. See issue #2604. The changes has been made based on a mobile device with screen size 320x568 in mind and have also been tested on screen size 414x736 (iPhone Plus).

**Definition of Done**

1. Open the various info overlays on mobile and desktop and check that they look fine.

### DIFF
--- a/raiden-dapp/src/components/overlays/InfoOverlay.vue
+++ b/raiden-dapp/src/components/overlays/InfoOverlay.vue
@@ -1,18 +1,20 @@
 <template>
   <div class="info-overlay">
-    <div class="info-overlay__header">
-      <header-content disable-back-button disable-info-button />
-    </div>
-    <div class="info-overlay__close-icon">
-      <v-icon icon @click="closeOverlay()"> mdi-close </v-icon>
-    </div>
-    <div class="info-overlay__image">
-      <img :src="require(`@/assets/info-overlay/${$route.meta.infoOverlay.headerImage}`)" />
-    </div>
-    <span class="info-overlay__title">{{ $t($route.meta.infoOverlay.header) }}</span>
-    <span class="info-overlay__body">{{ $t($route.meta.infoOverlay.body) }}</span>
-    <div class="info-overlay__button">
-      <action-button enabled :text="$t('info-overlay.done')" @click="closeOverlay()" />
+    <div class="info-overlay__content">
+      <div class="info-overlay__content__header">
+        <header-content disable-back-button disable-info-button />
+      </div>
+      <div class="info-overlay__content__close-icon">
+        <v-icon icon @click="closeOverlay()">mdi-close</v-icon>
+      </div>
+      <div class="info-overlay__content__image">
+        <img :src="require(`@/assets/info-overlay/${$route.meta.infoOverlay.headerImage}`)" />
+      </div>
+      <span class="info-overlay__content__title">{{ $t($route.meta.infoOverlay.header) }}</span>
+      <span class="info-overlay__content__body">{{ $t($route.meta.infoOverlay.body) }}</span>
+      <div class="info-overlay__content__button">
+        <action-button enabled :text="$t('info-overlay.done')" @click="closeOverlay()" />
+      </div>
     </div>
   </div>
 </template>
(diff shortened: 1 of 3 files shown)


weilbith:
I haven't reviewed the code yet, but tried it out. Do you think we can fix the close button at the top right?

Screenshot 2021-04-23 at 11 30 56

raiden-dapp/src/components/overlays/InfoOverlay.vue
```diff
@@ -38,60 +40,77 @@ export default class InfoOverlay extends Vue {
...
 
 .info-overlay {
-  align-items: center;
   background-color: $transfer-screen-bg-color;
   border-radius: 10px;
-  display: flex;
-  flex-direction: column;
-  height: calc(844px);
-  padding: 0 46px;
+  height: 844px;
```
weilbith, line 50: Why such an absolute value here and not relative to the parent (i.e. 100% like the width below)? I think this would even allow to unify it with the mobile view. Not too sure. But it look more complicated than it is...

raiden-dapp/src/scss/dimensions.scss
```diff
@@ -1,3 +1,4 @@
 $first-child-padding: 80px;
 $list-header-padding-top: 30px;
 $dialog-button-height: 35px;
+$ios-safe-area: 44px;
```
weilbith, line 4:
```suggestion
$ios-safe-area-height: 44px;
```

raiden-dapp/src/components/overlays/InfoOverlay.vue
```diff
@@ -38,60 +40,77 @@ export default class InfoOverlay extends Vue {
...
+  &__content {
+    align-items: center;
     display: flex;
-    height: 80px;
-    width: 100%;
-  }
+    flex-direction: column;
+    height: 100%;
+    @include respond-to(handhelds) {
+      margin-top: $ios-safe-area;
```
weilbith, line 67: I think this is wrong. You set the height to 100% and now you push it down by a margin. I think `padding-top` is correct. Else you push the content down. If you go into the dev-tools and inspect the layout you can see this.

raiden-dapp/src/components/overlays/InfoOverlay.vue
```diff
@@ -38,60 +40,77 @@ export default class InfoOverlay extends Vue {
...
-  padding: 0 46px;
+  height: 844px;
   position: absolute;
   width: 100%;
   z-index: 102;
+
   @include respond-to(handhelds) {
     border-radius: 0;
     height: 100vh;
+    overflow-y: auto;
```
weilbith, line 58: Can't this be device inspecific? So it just applies when necessary. And as it should be only become necessary for very small devices this is implicit.

raiden-dapp/src/components/overlays/InfoOverlay.vue
```diff
@@ -38,60 +40,77 @@ export default class InfoOverlay extends Vue {
...
-  height: calc(844px);
-  padding: 0 46px;
+  height: 844px;
   position: absolute;
   width: 100%;
   z-index: 102;
+
   @include respond-to(handhelds) {
     border-radius: 0;
     height: 100vh;
```
weilbith, line 57: I think this can be removed if you change the above to 100%. But I'm not familiar with the details of the bug.

taleldayekh:
> I haven't reviewed the code yet, but tried it out. Do you think we can fix the close button at the top right?

It's there by design. The title is meant to appear as part of the header with nothing else in it or above it. But perhaps it looks better if it is not like that.

raiden-dapp/src/components/overlays/InfoOverlay.vue
```diff
@@ -38,60 +40,77 @@ export default class InfoOverlay extends Vue {
...
+      }
+    }
 
-  &__image {
-    margin-top: 46px;
-    height: auto;
-    width: 300px;
-  }
+    &__close-icon {
+      display: flex;
```
weilbith, line 80: This does not work. The content div has flex column. Thereby it floats below the header due to the listing. It needs to be `absolute` and then set `right` and `top`.

raiden-dapp/src/components/overlays/InfoOverlay.vue
```diff
@@ -38,60 +40,77 @@ export default class InfoOverlay extends Vue {
...
-  &__button {
-    align-items: flex-end;
-    display: flex;
-    flex: 1;
-    margin-bottom: 98px;
-    width: 100%;
+    &__body {
+      font-size: 19px;
+      text-align: center;
+      margin: 0 36px;
```
weilbith, line 101: Maybe you want to define a left/right padding for the whole content div?

taleldayekh:
> PS: I used Firefox for this. And it doesn't change if I increase the height to your suggested value.

Ah, I did not meant for it to adjust perfectly to the screen, I just fixed it so that it does not break.

weilbith (changes requested):
May nitpicks. Sorry. But as the styling/css of this dApp is often very unnecessary chaotic I tried to point out a couple of things here.
I must admit I could not determine what of the changes made the actual fix. I don't really see why this new nested content div is useful for. 🤷🏾

raiden-dapp/src/components/overlays/InfoOverlay.vue
taleldayekh: Since it is part of the header it would be relative to the header, the header is not the size of the application and it would therefore look like this:

Screenshot 2021-04-23 at 11 57 37

weilbith: Ah okay. Damn. We need to have this as a variable somewhere...

raiden-dapp/src/scss/dimensions.scss
taleldayekh: This will actually be removed completely. We tested the dApp on an new iPhone yesterday and the safe area is not a problem when you run it as a progressive web app. @palango suggested to check if this can be adjusted in Capacitor somehow. But for now that's not an issue.

raiden-dapp/src/components/overlays/InfoOverlay.vue
weilbith: Close because of [this comment](https://github.com/raiden-network/light-client/pull/2679#discussion_r619096640)

raiden-dapp/src/components/overlays/InfoOverlay.vue
taleldayekh: You are right, padding would make more sense. But same comment as above. I will remove this completely anyways.

raiden-dapp/src/components/overlays/InfoOverlay.vue
taleldayekh: Yes, I'll fix that 👍🏽

raiden-dapp/src/scss/dimensions.scss
weilbith: Okay. So you will delete it in this PR? If so feel free to close the conversation when this happened. Else I would still rename it. It is code that gets merged, no matter how long we think **now** this will live.

raiden-dapp/src/components/overlays/InfoOverlay.vue
taleldayekh: It is not floating below the header because of the flex. It is below the header regardless and by design. The flex is simply there to be able to use `justify-content: flex-end;` to push the x-icon to the far left.

raiden-dapp/src/components/overlays/InfoOverlay.vue
weilbith: Same as in the conversation above. 👍🏾

raiden-dapp/src/components/overlays/InfoOverlay.vue
taleldayekh: Yeah, that is good, then I can also remove the padding from the icon. 👍🏽

raiden-dapp/src/scss/dimensions.scss
taleldayekh: Yes, exactly, I will delete it.

raiden-dapp/src/components/overlays/InfoOverlay.vue
weilbith: But... I don't understand. I mean this can't work. Maybe invite me to a call to discuss this quickly.

weilbith: Okay in the call in turned out that this is the intended design. For me it looked broken. But if that is how it should look like, it is correctly implemented. 👍🏾

weilbith (approved):
Thanks for working on all the comments. 🙃

PR #2683, opened 2021-04-25
Remove dependency on reactive notifications of peer's presences changes and update WebRTC signaling algorithm

### LINKED ISSUE #2677: Reduce presence dependency by allowing fetching presence from PFS when not available
## Description
Currently, presence information for a given node's `address` is required for some operations:
- sending a message, to know to which `userId` to send it
- evaluating if a target or partner in a mediated transfer is available

We must reduce the dependencies on these information as much as possible, and when required, have a sane fallback, e.g. fetching it directly from the specific PFS endpoint, and then optimize that by using contextual information when possible, e.g. by evaluating and using information coming through a `LockedTransfer.metadata`.

## Acceptance criteria
- Mediated transfers can work even when presence information **coming from the Matrix server** isn't reliable.

## Tasks
- [ ]

andrevmatos:
Fixes #2677 

**Short description**
This doesn't actually remove the presence completely, but instead removes `latest.presences` _cache_ usage, replacing it by explicitly requesting presence and waiting for the success. The epic handling it then can decide if it'll use a cached presence (e.g. latest seen, as previously) or actively re-request the current user presence, e.g. from matrix (to not be supported on `next` transport) or from PFS (through a new endpoint).

Also, WebRTC now has a new signaling protocol, backwards compatible with past protocol:
- All nodes now acts as both caller and callee: i.e. removed the fixed ordering based on sorted addresses
- Callee starts listening peer's messages as soon as peer is whitelisted, and will try to answer as soon as offer is received; it uses/respects `callId` (`channel.label`) sent by caller, and don't mind caller's `webRTC` capability
- Caller start to try to call peer upon certain events of interest
  - Currently:
    - `channelMonitored` (startup partners, new channel opened partners)
    - `transferSigned` (transfer's initiator, target; can be removed/narrowed later if we can exchange secrets without a direct RTC channel)
    - `messageSend.request` (to retry when trying to message a peer which we don't have an RTC channel with yet)
    - `rtcChannel` with `payload=undefined` (previous/last RTC channel closed)
  - Caller always retrieve presence from PFS, errors if peer don't have `webRTC` capability set
  - Caller retries a *limited* number of times, expo backoff 5s-60s delays, ~6 retries, then give up and retry again iff a new event from above is emitted
- At all times, for each peer, `callee` and `caller` channels race, the winner cancels the other's attempt to establish the channel.
- Since callee is always listening, peer's caller could retry and would still be able to get a new channel through
- If a node is online for some time, its caller should have given up; a partner coming online will then get their caller handler calling, so it always get almost instantly (except for matrix's `toDevice` message delays) connected
- If both nodes are coming online at the same time, and they get each other's presence as offline at first, caller's retry loop will get them connected after ~5s at most, so races are always gracefully handled
- This protocol also work for a matrixless transport, where one side can't receive calls (maybe a web dApp which can't expose a REST endpoint for such) but also is always expected to connect to full node's partners which are always online and  have such open endpoints
- This is backwards compatible with previous WebRTC algorithm, because since it acts as both caller & callee, the previous transport requiring/assuming a fixed role will always succeed.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/transport/epics/webrtc.ts
+++ b/raiden-ts/src/transport/epics/webrtc.ts
@@ -19,7 +19,6 @@ import {
   catchError,
   delayWhen,
   distinct,
-  distinctUntilChanged,
   endWith,
   filter,
   finalize,
@@ -55,11 +54,10 @@ import type { RaidenEpicDeps } from '../../types';
 import { isActionOf, isResponseOf } from '../../utils/actions';
 import { jsonParse, jsonStringify } from '../../utils/data';
 import { matchError } from '../../utils/error';
-import { completeWith, takeIf, timeoutFirst } from '../../utils/rx';
+import { completeWith, dispatchRequestAndGetResponse, takeIf, timeoutFirst } from '../../utils/rx';
 import type { Address } from '../../utils/types';
 import { decode, isntNil } from '../../utils/types';
-import type { matrixPresence } from '../actions';
-import { rtcChannel } from '../actions';
+import { matrixPresence, rtcChannel } from '../actions';
 import { getCap, getSortedAddresses } from '../utils';
 import { parseMessage } from './helpers';
 
@@ -554,8 +552,8 @@ export function rtcConnectEpic(
   action$: Observable<RaidenAction>,
   {}: Observable<RaidenState>,
   deps: RaidenEpicDeps,
-): Observable<rtcChannel | messageSend.request | messageReceived> {
-  const { latest$, config$, address } = deps;
+): Observable<rtcChannel | messageSend.request | messageReceived | matrixPresence.request> {
+  const { config$, address } = deps;
   return action$.pipe(
     // allow RTC connect to neighbors, initiators for received and targets for sent transfers
     filter(isActionOf([transferSigned, channelMonitored])),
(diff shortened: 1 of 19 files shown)


raiden-ts/src/transport/epics/webrtc.ts
```diff
@@ -217,377 +238,453 @@ function handleCandidates$(
...
+                  open$.next(true);
+                  open$.complete();
+                  return rtcChannel(dataChannel, { address: info.peerAddress });
                 }),
               ),
-              // send invite with offer, complete when response goes through
-              dispatchAndWait$(action$, request, isResponseOf(messageSend, meta)).pipe(
-                endWith([connection, dataChannel] as const),
+              fromEvent<MessageEvent>(dataChannel, 'message').pipe(
+                observeOn(asapScheduler),
```
palango, line 298: What does this do?

raiden-ts/src/config.ts
```diff
@@ -76,8 +71,6 @@ export const RaidenConfig = t.readonly(
       discoveryRoom: t.union([t.string, t.null]),
       additionalServices: t.readonlyArray(t.union([Address, t.string])),
       pfsMode: PfsModeC,
-      pfsRoom: t.union([t.string, t.null]),
```
palango, line 79: Do we need to do something for compatibility with old DBs?

palango (commented):
Looks good in general. The walk-through yesterday helped a lot understanding this.

Some codeclimate issues left to fix.

raiden-ts/src/config.ts
andrevmatos: No, it's safe to have extra properties on the DB. The problem would be to change the format/type of something, or to add a required property which would not be there on already initialized DBs, only then we need a migration.

palango (approved):
Good to go after a rebase, the removal of the commits from `next` and the required fixes fro code climate.

Issue #2588, opened 2021-02-24
Refactor unit tests


taleldayekh:
## Description
This can be difficult/challenging to do.

## Acceptance criteria
Write test for those functions:
- [x] `synced: Promise<raidenSynced['payload'] | undefined> = makeSyncedPromise(this.action$);` (@andrevmatos)
- [x] `getTokenInfo = makeTokenInfoGetter(this.deps);` (@andrevmatos)
- [x] `userDepositTokenAddress: () => Promise<Address> = memoize(` (@andrevmatos)
- [x] `static async create<R extends typeof Raiden>(` (@andrevmatos)
- [x] `async start(): Promise<void> {`
- [x] `get started(): boolean | undefined {`
- [x] `async stop(): Promise<void> {`
- [x] `get log() {`
- [x] `get state(): RaidenState {`
- [x] `get address(): Address {`
- [x] `get mainAddress(): Address | undefined {`
- [x] `get network(): Network {`
- [x] `async getBlockNumber(): Promise<number> {`
- [x] `updateConfig(config: PartialRaidenConfig) {`
- [x] `async getAvailability(`
- [x] `async monitorToken(token: string): Promise<Address> {`
- [x] `async dumpDatabase() {` (@andrevmatos)
- [x] `getBalance(address?: string): Promise<BigNumber> {` (@palango)
- [x] `async getTokenBalance(token: string, address?: string): Promise<BigNumber> {` (@palango)
- [x] `async getTokenList(): Promise<Address[]> {` (@andrevmatos)
- [x] `async openChannel(` (@agatsoh)
- [x] `async depositChannel(` (@agatsoh)
- [x] `async withdrawChannel(` (@agatsoh)
- [x] `async closeChannel(` (@agatsoh)
- [x] `async settleChannel(` (@agatsoh)
- [x] `async transfer(` (@palango)
- [x] `async waitTransfer(transferKey: string): Promise<RaidenTransfer> {`
- [x] `async findPFS(): Promise<PFS[]> {`
- [x] `async findRoutes(`
- [x] `async directRoute(`
- [x] `async mint(` (@andrevmatos)
- [x] `async getUDCCapacity(): Promise<BigNumber> {`
- [x] `async getUDCTotalDeposit(): Promise<BigNumber> {`
- [x] `async depositToUDC(`
- [x] `async transferOnchainBalance(` (@palango)
- [x] `async transferOnchainTokens(` (@palango)
[...]

PR #2686, opened 2021-04-28
sdk: Add unit tests for `Raiden` class

palango:
**Short description**
Part of #2588

This adds tests for
- `getAvailability`
- `getUDCCapacity`
- `getUDCTotalCapacity`

### DIFF
--- a/raiden-ts/tests/unit/raiden.spec.ts
+++ b/raiden-ts/tests/unit/raiden.spec.ts
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { getAddress } from '@ethersproject/address';
+import { BigNumber } from '@ethersproject/bignumber';
 import { hexlify } from '@ethersproject/bytes';
 import type { Network } from '@ethersproject/providers';
 import { JsonRpcProvider } from '@ethersproject/providers';
@@ -10,7 +11,7 @@ import logging from 'loglevel';
 import type { MatrixClient } from 'matrix-js-sdk';
 import type { Observable } from 'rxjs';
 import { AsyncSubject, BehaviorSubject, of, ReplaySubject } from 'rxjs';
-import { filter, first, ignoreElements, mapTo } from 'rxjs/operators';
+import { filter, first, ignoreElements, map, mapTo } from 'rxjs/operators';
 
 import type { RaidenAction } from '@/actions';
 import { raidenConfigUpdate, raidenShutdown, raidenStarted, raidenSynced } from '@/actions';
@@ -28,12 +29,14 @@ import {
 } from '@/contracts';
 import { combineRaidenEpics } from '@/epics';
 import { Raiden } from '@/raiden';
+import { udcDeposit } from '@/services/actions';
 import type { RaidenState } from '@/state';
 import { makeInitialState } from '@/state';
 import { standardCalculator } from '@/transfers/mediate/types';
+import { matrixPresence } from '@/transport/actions';
 import type { ContractsInfo, Latest, RaidenEpicDeps } from '@/types';
 import { pluckDistinct } from '@/utils/rx';
-import type { Address } from '@/utils/types';
+import type { Address, UInt } from '@/utils/types';
 
 jest.mock('@ethersproject/providers');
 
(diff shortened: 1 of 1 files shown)


raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -120,9 +124,109 @@ function makeDummyDependencies(): RaidenEpicDeps {
...
+    const raiden = new Raiden(
+      dummyState,
+      deps,
+      combineRaidenEpics(of(initEpicMock, matrixPresenceEpicMock)),
+      dummyReducer,
+    );
+    await expect(raiden.start()).resolves.toBeUndefined();
+    expect(raiden.started).toEqual(true);
+
+    raiden.action$.pipe(filter(matrixPresence.success.is)).subscribe((action) => {
```
agatsoh, line 177: Is it possible for you to add a promise pattern here like I did in `stop` ?
https://github.com/raiden-network/light-client/blob/e6c2c46339e9bd4a45cd82c490a6e83939112c7a/raiden-ts/tests/unit/raiden.spec.ts#L162

raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -120,9 +124,109 @@ function makeDummyDependencies(): RaidenEpicDeps {
...
+    );
+  }
+
   test('address', () => {
     const deps = makeDummyDependencies();
     const raiden = new Raiden(dummyState, deps, combineRaidenEpics(of(dummyEpic)), dummyReducer);
     expect(raiden.address).toBe(dummyState.address);
   });
+
+  function matrixPresenceEpicMock(action$: Observable<RaidenAction>) {
```
agatsoh, line 148: Just a nitpick here, it would be good if you keep all the epic mocks at the top together and then start with the tests.

palango: Don't you think it gets more readable this way?

raiden-ts/tests/unit/raiden.spec.ts
palango: What's the advantage in your eyes?

agatsoh: I got a similar review from Andre and wondered the same. Apart from the thing that it looks clean the `.toPromise` function is called only when the observable completes which is the last emitted value each time from the observables

agatsoh: I kind of know why Andre is making use of this but i also kind of not like to use this since this is going to be deprecated pretty soon.
https://indepth.dev/posts/1287/rxjs-heads-up-topromise-is-being-deprecated

raiden-ts/tests/unit/raiden.spec.ts
agatsoh: Ok just that in other tests as well i observed that whatever we wanted to mock we either place it in the `beforeEach` function or just after starting `describe` and then the tests are all grouped together without anything in between each test. but this is minor and you could leave it.

raiden-ts/tests/unit/raiden.spec.ts
andrevmatos: There'll be a drop-in replacement, `lastValueFrom`, so the pattern is still valid. The main benefit of it is that the `expect` calls are performed in the same context (the test function body), instead of a callback (as with `subscribe`), so it's harder to reach the end of the `async test` before all assertions ran. Also, supposedly, it should give better traceback in case of failures, although this isn't always the case with the current node+jest default parameters.

raiden-ts/tests/unit/raiden.spec.ts
andrevmatos: I'd prefer, if some mock was specific to a single test and not reused, that this was inside the test body, in order to not pollute the outside context, but this is just a matter of taste and codestyle.

raiden-ts/tests/unit/raiden.spec.ts
agatsoh: Thanks @andrevmatos  for explaining it in a better way !!!

raiden-ts/tests/unit/raiden.spec.ts
palango: ✔️

raiden-ts/tests/unit/raiden.spec.ts
palango: ✔️

agatsoh (approved):
LGTM, Thanks

Issue #2690, opened 2021-04-29
Allow the user to cancel connection process


weilbith:
## Description

This is in first place a workaround/fix for the experimental WalletConnect provider.
In case the connection pending dialog pops up, it should include a button which allows the user to cancel the connection before it finishes. This tackles the issue where a provider does not work properly but does not throw. For example it never signs the  subkey due to an issue. A practical case where this happens is when the user had established successfully a session with his wallet via WalletConnect. Then the wallet closed the session while the dApp is offline. The next time the user runs the dApp, it will still pick up the old session and asks the wallet to sign the subkey. This will never resolve as the wallet is nomore part of the session. But the current WalletConnect library does not detect this scenario. As the user has no way out of this situation as it is right now, he would be forever stuck here if he do not know how to use the developer debug tools.
Having such a cancel functionality in place allows the user to escape the problem and establish a fresh connection, though the data of his account is still preserved.

This issue require some refactoring of how the whole connection process works right now.

## Acceptance criteria
- 

## Tasks
- [ ]

Issue #2137, opened 2020-09-15
Replace strings with enums for state actions mutations etc.


taleldayekh:
## Description
@weilbith Could you fill in details?

## Acceptance criteria
- 

## Tasks
- [ ]

weilbith:
Related to #2186

PR #2691, opened 2021-04-29
dapp: add connection pending dialog

### LINKED ISSUE #2685: dApp - show dialog for long pending connection process
## Description

When the creation of the SDK instance takes some time, the user should be shown a dialog. This dialog should inform the user that he might to take some action in his provider/wallet like the signature for a Raiden subkey.

weilbith:
Fixes #2685

**Short description**
This is just very first issue and this feature will get improved over time (e.g. #2690)
I'm open to change the time value to anything. 

**Definition of Done**

1. Run the dApp
2. Click on the connect button and wait for >7s without taking any action in the wallet
3. Make sure a new dialog pops up which informs you that you need to take action in the wallet.

### DIFF
--- a/raiden-dapp/src/components/dialogs/ConnectionPendingDialog.vue
+++ b/raiden-dapp/src/components/dialogs/ConnectionPendingDialog.vue
@@ -1,0 +1,44 @@
+<template>
+  <raiden-dialog :visible="dialogVisible" :hide-close="true">
+    <v-card-title>
+      {{ $t('home.connection-pending-dialog.title') }}
+    </v-card-title>
+
+    <v-card-text>
+      <spinner :size="60" class="my-4" />
+      {{ $t('home.connection-pending-dialog.text') }}
+    </v-card-text>
+  </raiden-dialog>
+</template>
+
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator';
+
+import RaidenDialog from '@/components/dialogs/RaidenDialog.vue';
+import Spinner from '@/components/icons/Spinner.vue';
+
+const SHOW_DIALOG_TIMEOUT = 7000;
+
+/*
+  Please note that because the underlying dialog component by Vuetify, it is not
+  possible to implement the delayed rendering with a Vue transition or pure CSS
+  effect. Therefore this alternative approach of using the life-cycle hook and
+  a timer is necessary.
+*/
+@Component({ components: { RaidenDialog, Spinner } })
+export default class ConnectionPendingDialog extends Vue {
+  dialogVisible = false;
+
+  mounted() {
+    this.delayVisibilityOfDialog();
+  }
+
+  delayVisibilityOfDialog(): void {
+    setTimeout(this.showDialog, SHOW_DIALOG_TIMEOUT);
+  }
+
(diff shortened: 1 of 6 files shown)


weilbith:
CodeClimate is complaining about our index store. We already know that this one is messed up and we need time and priority to refactor this properly. This gets handled by #2137 Therefore we will ignore it for this PR and hope it does not continue complaining on every PR not affecting the store.

raiden-dapp/src/components/dialogs/ConnectionPendingDialog.vue
```diff
@@ -0,0 +1,44 @@
...
+*/
+@Component({ components: { RaidenDialog, Spinner } })
+export default class ConnectionPendingDialog extends Vue {
+  dialogVisible = false;
+
+  mounted() {
+    this.delayVisibilityOfDialog();
+  }
+
+  delayVisibilityOfDialog(): void {
```
taleldayekh, line 36: Not sure I understand why we are waiting 7 seconds to display the dialog?

weilbith: The idea was to support the user when it might be necessary but don't annoy him. So theoretically if everything is fine the user should never see this dialog as it is now. But when after some time the connection process did not finish, it is likely that the user is missing to do something. Like we all know the case where MetaMask is not popping up and we are waiting for a long time. Or especially for WalletConnect where you manually need to open the app.
So the intention is too keep the user-flow nice and clean, but when it seems necessary support him.

taleldayekh: I will try it out first and see how it feels like with your current implementation.

taleldayekh: Hmm, so for me the QR code appeared and after I've logged in to WalletConnect and scanned the QR code the dialog and spinner was already visible. So I guess a user will always see it because in many cases it will take more than 7 seconds to pick up your phone and all that stuff?

taleldayekh (approved):
All good, we agreed on keeping this implementation as is and re-evaluate in the future when more functionality has been added.

raiden-dapp/src/components/dialogs/ConnectionPendingDialog.vue
weilbith: We had a discussion about this in a call. The result is that we are unsure about the whole flow in detail. Especially because with the upcoming connection manager it will change. We have arguments for and against different solutions here. As a conclusion we are in a pat situation. Therefore, for the moment, we keep it as what it was introduced for: a fix for the WalletConnect session issue. Once the new user flow with the connection manager got established, we can evaluate how well this fits. For example will this close Talel his comment that the dialog will be always visible due to the long session creation process. Maybe we will even merge the dialogs content into the new loading indicator as this will be target of change anyway.

As the comment might show, many things are in change and the vision is not 100% clear. Therefore we do one step after each other and improve other time.

weilbith:
I put a longer summary in the resolved conversation. Thanks for your help @taleldayekh 🙏🏾

PR #2692, opened 2021-04-29
sdk: Add test for `withdrawFromUDC`

palango:
Part of #2588 

**Short description**

Adds a test for `withdrawFromUDC`

### DIFF
--- a/raiden-ts/tests/unit/raiden.spec.ts
+++ b/raiden-ts/tests/unit/raiden.spec.ts
@@ -31,6 +31,7 @@ import {
 import { combineRaidenEpics } from '@/epics';
 import { Raiden } from '@/raiden';
 import { udcDeposit, udcWithdraw, udcWithdrawPlan } from '@/services/actions';
+import { pfsListInfo } from '@/services/utils';
 import type { RaidenState } from '@/state';
 import { makeInitialState } from '@/state';
 import { standardCalculator } from '@/transfers/mediate/types';
@@ -40,6 +41,11 @@ import { pluckDistinct } from '@/utils/rx';
 import type { Address, UInt } from '@/utils/types';
 
 jest.mock('@ethersproject/providers');
+
+jest.mock('@/services/utils', () => ({
+  ...jest.requireActual<any>('@/services/utils'),
+  pfsListInfo: jest.fn(() => of(['abc', 'def'])),
+}));
 
 // TODO: dedupe this from integrations/mocks.ts in a higher utility file
 // don't import from there to avoid pulling in all the patches there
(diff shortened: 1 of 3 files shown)


raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -1,4 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
+import { makeHash } from '../integration/mocks';
```
andrevmatos, line 2: We shouldn't import from integration mocks. That pulls in some side-effect mocks as well (`patch*` functions). Please, create a `tests/utils.ts` module or alike, and pull up that function. There're some others which need to go there as well, like `makeAddress`.

raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -418,4 +428,74 @@ describe('Raiden', () => {
...
     await expect(payload).resolves.toEqual(expect.objectContaining({ block: withdrawBlock }));
   });
+
+  test('withdrawFromUDC', async () => {
+    const withdrawAmount = 10;
+
+    function udcWithdrawEpicMock(action$: Observable<RaidenAction>) {
+      return action$.pipe(
+        filter(udcWithdraw.request.is),
+        mapTo(
```
andrevmatos, line 438: This works fine because you **know** which parameters you're going to dispatch, but I'd say it'd be simpler and more correct to use `map` and pull some parameters (like `action.meta` and maybe some values from `payload`) from it.

raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -418,4 +428,74 @@ describe('Raiden', () => {
...
+        { address, network, contractsInfo },
+        { config: { additionalServices: ['pfs1'] }, services: { pfs2: 1 } },
+      ),
+      deps,
+      combineRaidenEpics(of(initEpicMock)),
+      dummyReducer,
+    );
+    await expect(raiden.start()).resolves.toBeUndefined();
+
+    const result = await raiden.findPFS();
```
andrevmatos, line 495: If possible, avoid using `await` outside `expect().resolves` and friends, since if these hang or reject, you're gonna get some uglier tracebacks.

andrevmatos (approved):
Some nitpicks, mostly good to go. Merge after accepting or ignoring them. Thanks

raiden-ts/tests/unit/raiden.spec.ts
palango: 👍

raiden-ts/tests/unit/raiden.spec.ts
palango: ✔️

raiden-ts/tests/unit/raiden.spec.ts
palango: Done, please have a look. I'm not sure if the `from '../utils';` import is correct.

PR #2693, opened 2021-04-29
Adding openChannel test for the raiden class

agatsoh:

Part of #2588

**Short description**
`openChannel` function test

1.
2.

### DIFF
--- a/raiden-ts/src/epics.ts
+++ b/raiden-ts/src/epics.ts
@@ -2,7 +2,7 @@ import { MaxUint256 } from '@ethersproject/constants';
 import unset from 'lodash/fp/unset';
 import isEqual from 'lodash/isEqual';
 import negate from 'lodash/negate';
-import type { Observable } from 'rxjs';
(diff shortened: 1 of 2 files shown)


agatsoh:
@andrevmatos  here is my code openChannel please have a look.

andrevmatos (approved):
I reviewed and fixed the test. There were some small issues which made it more difficult to debug this because the dummy logger was not enabled (`setLevel`) by default. But the main issue was that `waitForPFSCapacityUpdate` inside `openChannel` was not resolving because the `as Signature` field of the mocked `PFSCapacityUpdate` did not match an actual signature (it was just a few bytes long, instead of the required 65b). When mocking values, it may be important to check they all actually pass any required check, like this one. Once fixing this, the test passed, and I could even remove the `scheduled` operator, since the raw one already worked.
Additionally to that, I also rebased and made a small QoL improvement, by allowing one to pass the epics list to `combineRaidenEpics` directly as an array, instead of requiring constructing an `Observable` with `of` (but this also works). Let's get this merged.

PR #2697, opened 2021-05-03
sdk: Add test for `suggestPartners`, `transferOnchainBalance`, `transferOnchainTokens`

palango:
Part of #2588

### DIFF
--- a/raiden-ts/tests/unit/raiden.spec.ts
+++ b/raiden-ts/tests/unit/raiden.spec.ts
@@ -697,4 +697,62 @@ describe('Raiden', () => {
       ]),
     );
   });
+
+  test('getBalance/transferOnchainBalance', async () => {
+    const deps = makeDummyDependencies();
+    deps.signer = {
+      sendTransaction: jest.fn(async () => ({
+        wait: jest.fn(async () => ({
+          status: 1,
+        })),
+        hash: txHash,
+      })) as any,
+    } as any;
+
+    const raiden = new Raiden(
+      makeInitialState({ address, network, contractsInfo }),
+      deps,
+      combineRaidenEpics([initEpicMock]),
+      dummyReducer,
+    );
+
+    await raiden.start();
+
+    const balance = raiden.getBalance();
+    await expect(balance).resolves.toEqual(BigNumber.from(1_000_000));
+
+    const mockedGetBalance = deps.provider.getBalance as jest.MockedFunction<
+      typeof deps.provider.getBalance
+    >;
+    expect(mockedGetBalance.mock.calls[0][0]).toEqual(raiden.address);
+
+    // Test transfering the complete balance
+    const tx = raiden.transferOnchainBalance(partner);
+    await expect(tx).resolves.toEqual(txHash);
+
+    const mockedSendTransaction = deps.signer.sendTransaction as jest.MockedFunction<
+      typeof deps.signer.sendTransaction
+    >;
+    expect(mockedSendTransaction.mock.calls[0][0]).toEqual(
+      expect.objectContaining({
+        to: partner,
+        value: BigNumber.from(1_000_000 - 5 * 21_000),
+        gasPrice: BigNumber.from(5),
+        gasLimit: BigNumber.from(21_000),
+      }),
+    );
+
+    // Test transfering a specified amount
+    const tx2 = raiden.transferOnchainBalance(partner, 1, { gasPrice: 1 });
+    await expect(tx2).resolves.toEqual(txHash);
+
+    expect(mockedSendTransaction.mock.calls[1][0]).toEqual(
+      expect.objectContaining({
+        to: partner,
+        value: BigNumber.from(1),
+        gasPrice: BigNumber.from(1),
+        gasLimit: BigNumber.from(21_000),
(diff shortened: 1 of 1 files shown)


raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -49,6 +49,24 @@ import { makeAddress, makeHash } from '../utils';
 
 jest.mock('@ethersproject/providers');
 
+jest.mock('rxjs/fetch', () => ({
```
palango, line 52: @andrevmatos Is there a way to do this in the test locally? It's not nice to have to do this in the global namespace.

andrevmatos: Unfortunately, not; these mocks need to go in the global namespace, because they need to be mocked at `import` time (of the modules using the mocked values). Despite that, you should not mock `rxjs/fetch` directly, but instead the `globalThis.fetch` API, which is used by `fromFetch` under the hood. Give a look at `integration/mocks.ts` on how that can be done.

agatsoh: @palango if you are done with these changes then please let me know everything looks fine, so i could approve it.

palango: What's the advantage of overwriting `globalThis.fetch`?

palango: Anyhow, fix is done.

andrevmatos: The advantage is to be able to interact/mock a more streamlined API (fetch), and also not depend if it's used through the `fromFetch` helper or directly (e.g. we use both in the SDK codebase)

raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -82,6 +100,7 @@ function dummyReducer(state: RaidenState = dummyState) {
 
 function makeDummyDependencies(): RaidenEpicDeps {
   const provider = new JsonRpcProvider();
+  provider.getNetwork = jest.fn(async () => network);
```
andrevmatos, line 103: Nitpick, you can put `getNetwork` definition/mock inside the `Object.assign` below

raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -654,4 +685,68 @@ describe('Raiden', () => {
...
+    const raiden = new Raiden(
+      makeInitialState({ address, network, contractsInfo }),
+      deps,
+      combineRaidenEpics([initEpicMock]),
+      dummyReducer,
+    );
+    raiden.getBalance = jest.fn(async () => BigNumber.from(1_000_000)) as any;
+
+    await raiden.start();
+    const tx = raiden.transferOnchainBalance(partner, 1, { gasPrice: 1 });
```
andrevmatos, line 734: We need also to test the provider actually was called to transfer the requested amount, and it'd be very interesting to also test the case where we want to transfer the whole balance, with either specifying or auto gasPrice

raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -654,4 +685,68 @@ describe('Raiden', () => {
...
+    const raiden = new Raiden(
+      makeInitialState({ address, network, contractsInfo }),
+      deps,
+      combineRaidenEpics([initEpicMock]),
+      dummyReducer,
+    );
+    raiden.getTokenBalance = jest.fn(async () => BigNumber.from(1_000_000)) as any;
+
+    await raiden.start();
+    const tx = raiden.transferOnchainTokens(token, partner, 1);
```
andrevmatos, line 749: Same as above, but for tokens, no need to mind gasPrice then, but still check we're able to transfer the whole balance.

raiden-ts/tests/unit/raiden.spec.ts
palango: ✔️

raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -654,4 +690,138 @@ describe('Raiden', () => {
...
+    await raiden.start();
+
+    const balance = raiden.getTokenBalance(token);
+    await expect(balance).resolves.toEqual(BigNumber.from(1_000_000));
+
+    const tokenContract = deps.getTokenContract(token);
+    const mockedBalanceOf = tokenContract.callStatic.balanceOf as jest.MockedFunction<
+      typeof tokenContract.callStatic.balanceOf
+    >;
+    expect(mockedBalanceOf.mock.calls[0][0]).toEqual(raiden.address);
```
andrevmatos, line 811: Nitpick: usually, it's better to have these checks with `toHaveBeenCalledWith` and friends than accessing the call data directly, but no need to fix it here, they're kind of equivalent, but just a headsup for future tests

andrevmatos (approved):
lgtm, ty!

raiden-ts/tests/unit/raiden.spec.ts
palango: Ah cool. Didn't know about that one.

PR #2698, opened 2021-05-03
Adding unit test for closeChannel, depositChannel public function of the raiden class

agatsoh:

Part of  #2588 

**Short description**

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/tests/unit/raiden.spec.ts
+++ b/raiden-ts/tests/unit/raiden.spec.ts
@@ -15,7 +15,7 @@ import { filter, first, ignoreElements, map, mapTo, mergeMap } from 'rxjs/operat
 
 import type { RaidenAction } from '@/actions';
 import { raidenConfigUpdate, raidenShutdown, raidenStarted, raidenSynced } from '@/actions';
-import { channelDeposit, channelOpen, tokenMonitored } from '@/channels/actions';
+import { channelClose, channelDeposit, channelOpen, tokenMonitored } from '@/channels/actions';
 import type { Channel, ChannelEnd } from '@/channels/state';
 import { ChannelState } from '@/channels/state';
 import { BalanceProofZero } from '@/channels/types';
@@ -191,6 +191,12 @@ describe('Raiden', () => {
   const partner = makeAddress();
   const txBlock = 42;
   const transferAmount = 42;
+  const channelId = 10;
+  const settleTimeout = 500;
+  const isFirstParticipant = true;
+  const channelOpenHash = makeHash();
+  const meta = { tokenNetwork, partner };
+  const key = channelKey(meta);
 
   function initEpicMock(action$: Observable<RaidenAction>): Observable<raidenSynced> {
     return action$.pipe(
@@ -219,6 +225,37 @@ describe('Raiden', () => {
         ),
       ),
     );
+  }
+
+  const emptyChannelEnd: ChannelEnd = {
+    address: AddressZero as Address,
+    deposit: Zero as UInt<32>,
+    withdraw: Zero as UInt<32>,
+    locks: [],
+    balanceProof: BalanceProofZero,
+    pendingWithdraws: [],
+    nextNonce: One as UInt<8>,
+  };
+
+  function getChannel(): Channel {
+    return {
+      _id: channelUniqueKey({ ...meta, id: channelId }),
+      id: channelId,
+      state: ChannelState.open,
+      token: token,
+      tokenNetwork: tokenNetwork,
+      settleTimeout: settleTimeout,
+      isFirstParticipant: isFirstParticipant,
+      openBlock: txBlock,
+      own: {
+        ...emptyChannelEnd,
+        address,
+      },
+      partner: {
+        ...emptyChannelEnd,
+        address: partner,
+      },
+    } as Channel;
   }
 
   test('address', () => {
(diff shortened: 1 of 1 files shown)


raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -185,6 +189,54 @@ describe('Raiden', () => {
     );
   }
 
+  function tokensReducer(state: any, action: RaidenAction): RaidenState {
```
palango, line 192: This can me merged in the `makeInitialState` function like here: https://github.com/raiden-network/light-client/blob/ac2982ff85ffb2b6efe22b7ad0af8a3b7d5189cc/raiden-ts/tests/unit/raiden.spec.ts#L552

agatsoh: I would not like to make it there because this is in response to dispatching `tokenMonitored` action so its better we do it inside the test. `tokenMonitored` action is dispatched in the `monitorToken` function.

palango: But you already have a test for `monitorToken`, so no need to test this here again

agatsoh: Ok misunderstanding I meant I will use `makeInitialState` inside the test, because there is a `makeinitialState` outside the test where we construct `dummyState`

palango: Sure, but this allows you to get rid of the `tokensReducer` completely...

raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -155,6 +155,10 @@ describe('Raiden', () => {
   const txBlock = 42;
   const txHash = makeHash();
   const transferAmount = 42;
+  const id = 10;
```
palango, line 158: is this the channel id? if so please rename it.

raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -155,6 +155,10 @@ describe('Raiden', () => {
...
+  const settleTimeout = 500;
+  const isFirstParticipant = true;
+  const channelOpenHash = makeHash();
```
palango, line 161: Can you use `txHash` here instead of creating a new one?

raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -654,4 +665,56 @@ describe('Raiden', () => {
       }),
     ).resolves.toEqual(channelOpenHash);
   });
+
+  test('closeChannel', async () => {
+    const closeChannelHash = makeHash();
+    const closeBlock = 20;
+    const channelOpenedState: RaidenState = channelOpenSuccessReducer(
```
palango, line 672: Same here: if you just want to fill the `channels` field, why not do it directly?

andrevmatos: Yes, @agatsoh , I agree with Paul. You may need to use the open reducer and epic in case you're testing channel opening. After that, you should use/implement the bare minimum of the logic needed to test your specific test, and assign/mock directly anything that was already tested before and is *known* to work

agatsoh: I am just re-using the code that I used in `openChannel` and its going to be re-used in `depositChannel` as well, it makes the our job simpler

raiden-ts/tests/unit/raiden.spec.ts
agatsoh: I have named all of them differently as well declared them differently so that there is no ambiguity to anyone reading the tests later.

raiden-ts/tests/unit/raiden.spec.ts
agatsoh: yes this is the channel Id I will rename it.

raiden-ts/tests/unit/raiden.spec.ts
agatsoh: Its the same  what you would do to fill `channels` field just that `channelOpenSuccessReducer` makes it all the more easier to fill this channels field and we re-use code across the three tests.

agatsoh: @palango @andrevmatos I would anyhow without using the `channelOpenSuccessReducer` need to fill up the `Channel` type, so thats why I use the `channelOpenSuccessReducer` as a utility or helper function to help me fill it out. But if you still want me to not use it then I will make necessary changes. Waiting for your thoughts again on this.

agatsoh:
@andrevmatos may be you could give a look when you have time. I have rebased it.

raiden-ts/tests/unit/raiden.spec.ts
andrevmatos: @agatsoh you can create a utility to return you a filled state, or to create a Channel state for you which you can assign to your state. I know this can look as some cluttering, but this is actually how unit testing is supposed to work, or else you'd end up reimplementing the whole set of reducers and using it in place.

agatsoh: @andrevmatos ok i have made the necessary changes so you could give another look.

agatsoh:
@andrevmatos  Since they are very closely related I also added the `depositChannel` test in the same PR

raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -840,4 +899,70 @@ describe('Raiden', () => {
...
+                id: channelId,
+                participant: address as Address,
+                totalDeposit: deposit as UInt<32>,
+                txHash,
+                txBlock: 11,
+                confirmed: true,
+              },
+              { tokenNetwork, partner },
+            ),
+            messageServiceSend.request(
```
andrevmatos, line 920: Can't this mocked action be deduplicated from the `openChannel` test?

raiden-ts/tests/unit/raiden.spec.ts
```diff
@@ -825,6 +848,42 @@ describe('Raiden', () => {
...
+      makeInitialState(
+        { address, network, contractsInfo },
+        { tokens: { [token]: tokenNetwork }, channels: { [key]: getChannel() } },
+      ),
+      deps,
+      combineRaidenEpics([initEpicMock, channelCloseEpicMock]),
+      dummyReducer,
+    );
+    await expect(raiden.start()).resolves.toBeUndefined();
+    await expect(raiden.closeChannel(token, partner, { subkey: false })).resolves.toEqual(txHash);
```
andrevmatos, line 884: Please, also verify that a `promise` with `channelClose.request` action got emitted with the desired payload

agatsoh: Ok will do

raiden-ts/tests/unit/raiden.spec.ts
agatsoh: Done I have taken these actions out so that they could be shared between `openChannel ` and `depositChannel`

agatsoh:
@andrevmatos yes please i have made the requested changes.

andrevmatos (approved):
lgtm

PR #2701, opened 2021-05-05
sdk: Clean up helper imports in integration tests

palango:
The new utils module was introduced in 857952e67d1d177426f0f60b53c39e72c3775601
but some imports were not adjusted right away.


andrevmatos (approved):
lgtm; soon, we may also need to move the `integration/epics/` suites to `integration/`, since all integration tests include epics by definition, but let's do this one step after the other; thank you

PR #2712, opened 2021-05-10
dapp: adopt naming scheme for connection process

weilbith:
While working on the connection manager feature it has shown that the
term 'connect' gets used to much for multiple steps in the same process.
So originally the first step the user has to do to start the dApp is
called connecting. This term is used for the state in the Vuex store,
the button labels, the RaidenService class method, the Home view
component etc. pp. **and** now also for the ConnectionManager. Therefore
we will stick with this.
The trouble comes now with the new `EthereumConnection`s, plus their
static `connect` function. Which leads us to having an Ethereum
connection object which gets instaniated with a connect function which
will then be handed over to the actual connect procress.
To simplifiy this, we now name them `EthereumProvider`s. This is
also closer to the original naming of the `Web3Provider`. And the
factory function of them is called `link`. Thereby the user first links
to an Ethereum provider of his choice and use that to connect the dApp.


weilbith:
No CHANGELOG entry necessary. Only internal code changes.

taleldayekh (approved):
A bunch of renaming, go for it ;)

PR #2715, opened 2021-05-12
Feature/dapp add injected provider to connetion manager

weilbith:
Based on #2714

**Short description**
Next provider for the connection manager. No refactoring for the boilerplate code yet.
No CHANGELOG entry as it is no real new functionality and we don't have an issue for it. 😆 

**Definition of Done**

1. Open the dApp
2. Connect with MetaMask as usually
3. Disconnect and disable your MetaMask plugin
4. Reload the dApp
5. Make sure the injected provider button is disabled

### DIFF
--- a/raiden-dapp/src/components/ConnectionManager.vue
+++ b/raiden-dapp/src/components/ConnectionManager.vue
@@ -11,17 +11,31 @@
 
     <template v-if="showProviderButtons">
       <action-button
+        class="connection-manager__provider-dialog-button"
         :enabled="true"
-        text="Wallet Connect"
+        :text="$t('connection-manager.dialogs.wallet-connect-provider.header')"
         width="280px"
-        @click="openWalletConnectDialog"
+        @click="openWalletConnectProviderDialog"
       />
-      <wallet-connect-provider-dialog
-        v-if="walletConnectDialogVisible"
-        @linkEstablished="onProviderLinkEstablished"
-        @cancel="closeWalletConnectDialog"
+      <action-button
+        class="connection-manager__provider-dialog-button"
+        :enabled="true"
+        :text="$t('connection-manager.dialogs.injected-provider.header')"
+        width="280px"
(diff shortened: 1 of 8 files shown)


weilbith:
The fundamental PR with the connection manager got merged to `master`. Therefore I cherry-picked the changes on top of it and pushed them here. Thereby convert this PR and ready to review.

raiden-dapp/src/components/ConnectionManager.vue
```diff
@@ -93,22 +110,37 @@ export default class ConnectionManager extends Vue {
...
+
     if (rpcUrl && privateKey) {
       const provider = await DirectRpcProvider.link({ rpcUrl, privateKey });
       this.connect(provider);
     }
   }
 
-  async onProviderLinkEstablished(linkedProvider: EthereumProvider): Promise<void> {
-    await this.connect(linkedProvider);
+  openWalletConnectProviderDialog(): void {
```
taleldayekh, line 132: Question, why not have one function for both displaying and hiding the dialog? `this.walletConnectProviderDialogVisible = !this.walletConnectProviderDialogVisible`? Same for the injected provider? Is it because you want to use the `closeAllProviderDialogs` function?

taleldayekh (approved):
Looks good and works like a charm. 👍🏽

raiden-dapp/src/components/ConnectionManager.vue
weilbith: Yes, also. Besides its kinda not correct to say that the action button toggles the visibility and if the cancel event comes it also toggles. This only works when the state is always perfect.
But I agree, I dislike in general that this is so much overhead for each new dialog. I would love to simplify it. I was not able to come up with something that is not ugly in a different manner. For example I was considering to put into the dialog components themselves and there into the base dialog. But its no good practice I think. As long as you don't think this is a genius idea, I would like to stick with the current solution until we have something better.

PR #2723, opened 2021-05-18
dapp: remove obsolete cypress retry plugin

weilbith:
Cypress was complaining that this plugin is still installed. This functionality is now part of the core package. Apparently this plugin wasn't even used anymore.

If the CI runs though with its end-to-end tests that should be it.


palango (approved):
Fine to merge when tests pass.

PR #2750, opened 2021-05-24
Next: use address_metadata for optimizing sending and mediating transfers

andrevmatos:
Fixes # 

**Short description**
PC is expected to make this behavior optional, as we already implement, but currently they're requiring the `address_metadata` mapping be passed on transfer's `metadata`. Since we have it, we also implemented using it instead if requiring to actively request peer's presence upon `messageSend.request`, which could slowdown transfers in case WebRTC channel was not available (which is also the case for transfers with/through PC, since they've disabled their RTC transport).
This is still an **optimization**, and the clients can as always do without it.
For us, this is a **non-breaking change**, since we've implemented first the fallback of requesting peer's presence from service if needed.

**Definition of Done**

1.
2.


palango (approved):
Changes look good, but tests need to be fixed.

andrevmatos:
Since this is approved, I'll merge as is, and fix tests directly on top of master later. Thanks

PR #2761, opened 2021-05-27
dapp: Use demo env on staging temporarily

palango:
Otherwise the dApp cannot be used at all until `next` is merged

Once merged, I'll revert this commit on `next`.


taleldayekh (approved):
Thanks for fixing @palango 👍🏽

Issue #2730, opened 2021-05-19
Allow encrypted secret to be sent on LockedTransfer's metadata


andrevmatos:
## Description
An improvement to the transfer workflow is for the node to be able to send the transfer's secret ECIES encrypted to the target, so they don't need to request it from the initiator off-band, making the transfer faster, reducing the number of messages, removing the requirement for non-channel-connected nodes to be able to communicate.
The drawback is that the initiator can't know the transfer failed until it expires, due to they losing the ability to chose to reveal the secret or not, which now lays with the target.

## Protocol
- Metadata must be passed through **mediators** as is forward
  - Unknown fields must still be included in the canonicalJSON `additionalHash` signature in order for it to be valid, but the actual object must not be modified, including routes; they should be exclusively set by initiator, while mediators may **read-only** some if they need to, e.g. to get next hop in route
- **Initiator** encrypts the transfer's secret and include it in `metadata.secret`, together with the Request parameters, in order to ensure the target will be able to decide if they should reveal it and the mediators didn't take more fees than intended
  - Initiator always knows target's public key, due to knowing their presence, which allows the pubkey to be derived from displayName/userId signature and used for ECIES encryption
  - `metadata.secret` is a hex-encoded `bytestring` containing the encrypted version of a CanonicalJSON-serialized object
  - pseudo-code: `metadata.secret = hex_encode(encryptEcies(JSON.stringify(secret_payload), target.pubkey));`
  - `secret_payload` schema:
    - `payment_identifier`: string decimal encoded number up to 64bits
    - `secret`: hex-string 32bytes
    - `amount`: string decimal encoded number up to 256bits
    - `expiration`: block, string decimal encoded number up to 256bits
  - If the initiator needs to retain its ability to cancel the transfer by not revealing the secret at SecretRequest time, they may simply not do it, and go with the current schema. 
  - Even if sending it, the initiator should still be able to keep current behavior and reveal the secret to target in case they ask and SecretRequest parameters match, ensuring backwards compatibility.
- Upon receiving a transfer, the **target**:
  - Should check if `secret` is present in metadata, and try to decrypt and parse the resulting string as a JSON.
  - If it can, it then checks `secret` is present, and hashes to the transfer's `secrethash`
  - If any of above fails (also for backwards compatibility), target can go the usual path and `SecretRequest` to the initiator
[...]

PR #2766, opened 2021-05-31
Passthrough metadata unchanged

andrevmatos:
**Proposal** on how protocol regarding metadata should be.
1st bulletpoint of #2730
Implements https://github.com/raiden-network/spec/issues/343

**Short description**
- A new capability is added: `immutableMetadata`
  - Falls back to falsy, which is current (PC) behavior
    - Route must be cleared/prunned: partner expects to be first address in route
    - `LockedTransfer`'s `additionalHash` is generated from the JCS serialization **of the metadata transformed to checksummed addresses**, not necessarily what was received
  - Defaults to truthy (opt-in) on LC
  - When true (new behavior, below), tells partners client is able to receive routes where they're not first hop in `metadata.routes[*].route` array and passthrough metadata unchanged if next hop supports it
- `TransferLocked`'s `metadata` on mediation is passed **as is**, completely unchanged; this includes `route` being passed as is, without prunning
- It's signature (`additionalHash`) is validated as the **exact** signature of the JCS serialized received metadata (no transform in any way).
- Only when/where needed **internally**, we decode `metadata` **known fields** and change it to match a given schema: e.g. that addresses are in the checksum format; these mappings are only used for internal logic, and doesn't reach the created/relayed message or signature validation.
  - We try our best to support any format received here, and if we can't for something critical, instead of failing validation (which would put us out-of-sync), we fail the corresponding effect, e.g. fail to mediate the transfer, but still accept the received nonce (which will expire later, obviously).
 - Anything unknown in `metadata` is forwarded as is, and we sign it as is.
   - This could be a security risk (arbitrary hashes), but since this is signed strictly as part of a larger message (`LockedTransfer`), which is EIP191-prefixed, and also is guaranteed to contain at least some known fields (or else we couldn't route the transfer), there's no risk of it being used elsewhere/replayed.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -1,9 +1,13 @@
 # Changelog
 
 ## [Unreleased]
+### Added
+- [#2766] Add `Capabilities.IMMUTABLE_METADATA` (true on LC, fallback to falsy for backwards compatibility) to allow opting in of not prunning metadata.route and allowing to pass it through mediators unchanged
+
 ### Fixed
 - [#2831] Force PFS to acknowledge our capabilities updates
 
+[#2766]: https://github.com/raiden-network/light-client/pull/2766
 [#2831]: https://github.com/raiden-network/light-client/issues/2831
 
 ## [1.0.0] - 2021-06-16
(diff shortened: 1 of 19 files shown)


andrevmatos:
Rebased and changed opt-out `Capabilities.CLEAR_ROUTE` to opt-in `Capabilities.IMMUTABLE_METADATA`, as agreed with PC team.

palango (approved):
Nice job!

PR #2777, opened 2021-06-03
dapp: fix flaky end-to-end tests

### LINKED ISSUE #2774: dApp - end-to-end tests are flaky on the second connection
The dApp's end-to-end test sometimes fail after it was downloading a backup and then trying to re-connect again. The dialog for the private key opens and closes immediately closes again. As consequence the input elements are missing and the tests fail.

weilbith:
Fixes #2774

In fact it fixes the duplicated navigation error after a backup. Apparently this solves the end-to-end tests which fail right after doing a backup. The CI has succeed many times in a row. So fingers crossed this was the awkward cause for the bug. After all it does fix the navigation error.

### DIFF
--- a/raiden-dapp/CHANGELOG.md
+++ b/raiden-dapp/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [#2671] Fix preventing transfers to partners with closed channel
 - [#2675] Fix insecure behavior on navigation without connected token
 - [#2754] Fix empty logs in production mode
+- [#2774] Fix duplicated navigation error after doing a backup
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


weilbith:
Alright, the CI succeed 5 times in a row. It looks like that was the issue. I'll write a commit message and PR body and convert into a real PR.

weilbith:
You might see some cancelled CI workflows for this PR. This is because I have amended the commit to write the actual message, add the missing CHANGELOG entry and also rebase on `master` to resolve the conflict. So don't get confused by it.

weilbith:
We are now on 10 consecutive successful runs. Looks good.

taleldayekh (approved):
I trust the 10 consecutive successful runs :)

PR #2779, opened 2021-06-04
Feature/dapp unit tests should fail on console error

### LINKED ISSUE #2725: dApp - make unit tests fail on console error or warning output
## Description

While the unit tests for the dApp do succeed, they print tons of error and warning outputs by Vue. This is caused by incomplete mocking. We fixed that once, but it got out of control again. We are used to it again and don't see actual issues. Furthermore it is very annoying as it is super verbose and you  have to scroll for ever to see the relevant output if something fails.
Therefore I would like to add a small Jest plugin which makes a test case fail if there is some output on the console warning or error channel. This will detect wrong mockings early and forces the developer to fix it right away.

## Acceptance criteria
- dApp unit test fail on console errors/warnings 

## Tasks
- [ ]

weilbith:
Fixes #2725

### DIFF
--- a/raiden-dapp/jest.setup-after.ts
+++ b/raiden-dapp/jest.setup-after.ts
@@ -1,1 +1,7 @@
 import '@testing-library/jest-dom';
+import failOnConsole from 'jest-fail-on-console';
+
+failOnConsole({
+  shouldFailOnWarn: true,
+  shouldFailOnError: true,
+});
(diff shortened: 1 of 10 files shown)


weilbith:
No CHANGELOG entry necessary for this.

palango (approved):
Looks good, and there are no warnings nor errors in the dapp unit tests any more.

PR #2803, opened 2021-06-10
cli: add default mediation fee configuration

### LINKED ISSUE #2790: Set mainnet mediation fee defaults
We need to set mediation fee defaults for mainnet in the CLI.

Ref https://github.com/raiden-network/team/issues/881

## Tasks
- [x] Set default fixed fee: https://github.com/raiden-network/raiden/blob/develop/raiden/utils/mediation_fees.py#L34-L35
- [x] Set default proportional fee: https://github.com/raiden-network/raiden/blob/develop/raiden/settings.py#L78
- [x] Set default imbalance fee: https://github.com/raiden-network/raiden/blob/develop/raiden/settings.py#L79-L81
- [x] Set default cap fees setting: https://github.com/raiden-network/raiden/blob/develop/raiden/settings.py#L113

weilbith:
Fixes #2790

### DIFF
--- a/raiden-cli/src/config.json
+++ b/raiden-cli/src/config.json
@@ -8,6 +8,7 @@
     "Mediate": 1
   },
   "mediationFees": {
+    "cap": true,
     "0x0000000000000000000000000000000000000000": {
       "cap": true,
       "flat": 0, 
--- a/raiden-cli/src/config.json
+++ b/raiden-cli/src/config.json
@@ -4,6 +4,27 @@
   "autoSettle": true,
   "autoUDCWithdraw": false,
   "caps": {
-    "Receive": 1
+    "Receive": 1,
+    "Mediate": 1
+  },
+  "mediationFees": {
+    "0x0000000000000000000000000000000000000000": {
+      "cap": true,
+      "flat": 0, 
+      "proportional": 4000,
+      "imbalance": 3000
+    },
+    "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2": {
+      "cap": true,
+      "flat": 10000000000,
+      "proportional": 4000,
+      "imbalance": 3000
+    },
+    "0x6B175474E89094C44Da98b954EedeAC495271d0F": {
+      "cap": true,
+      "flat": 1000000000000, 
+      "proportional": 4000,
+      "imbalance": 3000
+    }
   }
 }
--- a/raiden-cli/CHANGELOG.md
+++ b/raiden-cli/CHANGELOG.md
@@ -6,10 +6,12 @@
 - [#1576] Add functionality to deploy token networks
 - [#2577] Add `--proportional-imbalance-fee` and `--cap-mediation-fees` options for imbalance fees
 - [#2795] Add `--gas-price` option to control gasPrice strategy as a multiplier of provider-returned `eth_gasPrice`
+- [#2590] Add default configuration for mediation fees
 
 [#1576]: https://github.com/raiden-network/light-client/issues/1576
 [#2577]: https://github.com/raiden-network/light-client/issues/2577
 [#2795]: https://github.com/raiden-network/light-client/issues/2795
+[#2790]: https://github.com/raiden-network/light-client/issues/2790
 
 ## [0.16.0] - 2021-04-01
 ### Changed
--- a/raiden-cli/src/index.ts
+++ b/raiden-cli/src/index.ts
@@ -148,11 +148,6 @@ async function parseArguments() {
         default: false,
         desc: "Enables monitoring if there's a UDC deposit",
       },
-      enableMediation: {
-        type: 'boolean',
-        default: true,
-        desc: 'Enables support for mediated payments.',
-      },
       flatFee: {
         type: 'string',
         nargs: 2,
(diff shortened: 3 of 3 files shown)


raiden-cli/src/config.json
```diff
@@ -4,6 +4,20 @@
   "autoSettle": true,
   "autoUDCWithdraw": false,
   "caps": {
-    "Receive": 1
+    "Receive": 1,
+    "Mediate": 1,
```
andrevmatos, line 8: Please, remove trailing commas in objects in JSON; Some JSON parsers complain about this. Ours doesn't, but only for json, it's better to keep compatibility here.

raiden-cli/src/config.json
```diff
@@ -4,6 +4,20 @@
...
+  },
+  "mediationFees": {
+    "0x0000000000000000000000000000000000000000": {
+      "flat": 0, 
+      "prop": 4000,
+      "inbalance": 3000,
+    },
+    "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2": {
+      "flat": 10000000000,
```
palango, line 17: Doesn't Adnre say here that this will reset the proportional and imbalance fees to zero? https://github.com/raiden-network/light-client/issues/2790#issuecomment-856891739

andrevmatos, line 17: Notice that setting an object containing only e.g. `flat` will shadow the whole default `AddressZero` config, meaning these tokens will have no config for the other models. Is this intended?

raiden-cli/src/config.json
```diff
@@ -4,6 +4,20 @@
...
   "caps": {
-    "Receive": 1
+    "Receive": 1,
+    "Mediate": 1,
+  },
+  "mediationFees": {
+    "0x0000000000000000000000000000000000000000": {
+      "flat": 0, 
+      "prop": 4000,
+      "inbalance": 3000,
```
andrevmatos, line 14: Not sure if my `imbalance`'s spelling is correct, but that's how it's currently in code:
```suggestion
      "proportional": 4000,
      "imbalance": 3000,
```

andrevmatos: Also, for the default config, I'd add the explicit `"cap": true`, even if the default is hardcoded.

raiden-cli/src/config.json
weilbith: Ah and the PC doesn't? Sorry then I misunderstood you when I asked you. Sure then this needs adoption to set them to zero.

weilbith: See the discussion in the review comment conversation above.

raiden-cli/src/config.json
weilbith: Fu** I'm stupid. Sorry. :see_no_evil:

weilbith:
Wow. One of the smallest, but yet worst PRs I ever opened. :expressionless:

raiden-cli/src/config.json
weilbith: ✔️

raiden-cli/src/config.json
weilbith: Thanks for the clarifying talk. Sorry that I messed this up. 😑

weilbith: ✔️

raiden-cli/src/config.json
weilbith: ✔️

weilbith:
In regards of [this conversation](https://github.com/raiden-network/light-client/issues/2790#issuecomment-858753518) from the related issue, @palango found out that the enable mediation command line argument does not exist on the Python client. Therefore we agreed to removed it here, so it does not clash with the configuration file. Addressed in efd7f4f.

weilbith:
What is still open is, that if the user sets fees via command line arguments, we need to set the missing fee model properties for this token to the default ones. This is because the Python Client and the Light Client do behave differently here. See [this comment](https://github.com/raiden-network/light-client/issues/2790#issuecomment-856891739)

weilbith:
Alright so a manual test shows that this call:
```sh
node build/index.js --accept-disclaimer --flat-fee 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 4 --proportionalFee 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 400 --flat-fee 0x6B175474E89094C44Da98b954EedeAC495271d0F 500 --proportionalImbalanceFee 0x6B175474E89094C44Da98b954EedeAC495271d0F 50000 --no-cap-mediation-fees
```

produces this configuration:

```json
{
  '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2': { flat: '4', proportional: '400', imbalance: 3000, cap: false },
  '0x6B175474E89094C44Da98b954EedeAC495271d0F': { flat: '500', proportional: 4000, imbalance: '50000', cap: false }
}
```

Note the default values which are actual numbers instead of the parsed command line option values. Is that correct this way?

weilbith:
I'll resolve the merge conflict when I squash the fixup commits.

andrevmatos:
@weilbith it's not a problem that the config merge produce mixed numbers and numeric strings. They both are decodable as numbers in the fee model's decoders, so it should be fine.

palango (approved):

PR #2804, opened 2021-06-10
Support customising gasPrice strategy

### LINKED ISSUE #2795: Implement gas price scaling in order to ensure transactions are mined in due time
## Description
Some onchain transactions in SDK need to be guaranteed to be mined before a given block. With scenarios of high gas price fluctuations, like mainnet and goerli currently, transactions may take several hours to get mined with the _naive_ approach currently set if simply using return of `eth_gasPrice` RPC method (default strategy of `ethers` when no gas price is specified). In dApp with e.g. Metamask, this is not a big problem, as the user needs to interactively confirm the transaction prompt, and therefore is in charge of guaranteeing a reasonable gas price, but CLI needs to perform this on its own.

Currently, the main transactions which have a "deadline" are:
- channel withdraw (withdrawal expiration block)
- channel update ( settleable block)
- secret register (lock expiration)

The proposed strategy is to start with the value returned by `eth_gasPrice`, and bump it by a given multiplier (e.g. 50%) when achieving **half way** through the deadline.
To do this from a central place, we should have contracts transactions requests be emitted as actions, and then a central epic to serialise and submit them, controlling `nonce` and `gasPrice` given strategy above.

## Acceptance criteria
- 

## Tasks
- [ ]

andrevmatos:
Fixes #2795 

**Short description**
This allows some control of `gasPrice` with which transactions are sent. On dApp/Metamask, this controls only the default presented `gasPrice` in MM's tx prompt. On CLI, `--gas-price` also can receive some aliases, like `fast` (1.2) or `medium` (1.05). When not set, the default is 1.0. This is a multiplier for the current ETH-provider-returned `eth_gasPrice`.

A better fix, presented on the issue, is to instead be able to even use lower gasPrices by default, but be able to bump them for priority transactions (txs with deadlines, like `register secret` or `channel update`, but this should be good enough for now. Eventually, those strategies could also be exposed through the API and let the user set it per operation.

**Definition of Done**

1. Check CLI with `--gas-price fast` option makes scenarios send transactions with +20% the current avg `gasPrice`

### DIFF
--- a/raiden-cli/src/index.ts
+++ b/raiden-cli/src/index.ts
@@ -176,6 +176,24 @@ async function parseArguments() {
         default: true,
         desc: 'Enables capping mediation fees to not allow them to be negative (output transfers amount always less than or equal input transfers)',
       },
+      gasPrice: {
+        desc: 'Set gasPrice strategy for transactions, as a multiplier of ETH node returned "eth_gasPrice"; some aliases: medium=1.05, fast=1.2, rapid=1.5',
+        coerce(val): number | undefined {
+          if (!val) return;
+          switch (val) {
+            case 'medium':
+              return 1.05;
+            case 'fast':
+              return 1.2;
+            case 'rapid':
+              return 1.5;
+            default:
+              val = +val;
+              assert(val && 0.1 <= val && val <= 10, 'invalid gasPrice');
+              return val;
+          }
+        },
+      },
     })
     .help()
     .alias('h', 'help')
@@ -311,6 +329,7 @@ function createRaidenConfig(
     pfsMaxPaths: argv.pathfindingMaxPaths,
     pfsMaxFee: argv.pathfindingMaxFee,
     pfsIouTimeout: argv.pathfindingIouTimeout,
+    ...(argv.gasPrice ? { gasPriceFactor: argv.gasPrice } : undefined),
   };
 
   if (argv.matrixServer !== 'auto') config = { ...config, matrixServer: argv.matrixServer };
--- a/raiden-ts/src/channels/epics.ts
+++ b/raiden-ts/src/channels/epics.ts
@@ -79,9 +79,10 @@ import {
   retryWhile,
   takeIf,
 } from '../utils/rx';
-import type { Address, Hash, HexString, Signature, UInt } from '../utils/types';
-import { isntNil, last } from '../utils/types';
+import type { Address, Hash, HexString, Signature } from '../utils/types';
+import { decode, isntNil, last, UInt } from '../utils/types';
 import {
+  blockGasprice,
   blockStale,
   blockTime,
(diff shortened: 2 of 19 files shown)


raiden-cli/src/index.ts
```diff
@@ -311,6 +329,7 @@ function createRaidenConfig(
     pfsMaxPaths: argv.pathfindingMaxPaths,
     pfsMaxFee: argv.pathfindingMaxFee,
     pfsIouTimeout: argv.pathfindingIouTimeout,
+    ...(argv.gasPrice ? { gasPriceFactor: argv.gasPrice } : undefined),
```
palango, line 332: Shouldn't we set a factopr of 1.0 instead of `undefined` if nothing is given?

raiden-ts/src/channels/epics.ts
```diff
@@ -250,6 +251,39 @@ export function blockStaleEpic(
...
+  { provider, config$ }: RaidenEpicDeps,
+): Observable<blockGasprice> {
+  return state$.pipe(
+    pluckDistinct('blockNumber'),
+    withLatestFrom(config$),
+    // switchMap will "reset" timer every block, restarting the timeout
+    exhaustMap(([, { gasPriceFactor }]) =>
+      defer(async () => provider.getGasPrice()).pipe(
+        map((price) => price.mul(Math.round(gasPriceFactor * 1e6)).div(1e6)),
+        map((price) => (price.gte(1e9) ? price : 1e9)), // min 1Gwei
```
palango, line 276: Why provide this minimum?

palango:
I guess I'd also keep the linked issue open, but push it to post mainnet.

raiden-cli/src/index.ts
andrevmatos: I wouldn't say so, let's leave the unset case to the SDK, and there the default is 1.0, so we don't have to keep track of the default in multiple places

raiden-ts/src/channels/epics.ts
andrevmatos: I've seen ETH node tripping balls sometimes and giving very low gasPrices (like 1wei) on testnets, which even there made txs stuck for several hours; this just ensures a reasonable minimum which is compatible with most networks.

raiden-cli/src/index.ts
palango: Sounds good, I missed the spread operator there.

raiden-ts/src/channels/epics.ts
palango: Thanks for the explanation. I was a bit scared that it's to high for testnets, but there it also doesn't matter if spending a bit more.

palango (approved):
Great work!

PR #2811, opened 2021-06-14
dapp: Lower codecov target

palango:
We break this since a while anyways, let's set it lower and fix it in the future.


andrevmatos (approved):
lgtm

PR #2814, opened 2021-06-14
Confirm confirmation on open/depositChannel methods

### LINKED ISSUE #2813: Make SDK confirm confirmation of open/deposit operations
While debugging kronk's `bf3_multi_directional_payment/node_4909*` failed SP run, I noticed this weird issue:
- `node004` deposit got mined at block `#8759000` and confirmed  it `21:55:50 @ #8759005` as expected.
- Meanwhile, probably due to ETH node's slowness/overload, `node003` got the blocks (i.e. was not stale) but for some reason only saw and immediately confirmed that deposit `21:56:45 @ #8759008`, i.e. `55s` after the actual confirmation.
- At this point, node004 had already responded to the deposit REST API call in CLI, and the SP moved on to perform the scenario's transfer, but node004's transfer failed with `No route between nodes found` at `21:56:25` because node003 hadn't yet confirmed that deposit and updated PFS on its side of the capacity of channel

So, the problem above happens because `open`/`deposit` operations on CLI resolve as soon as the node confirmed the transaction, but partners may or may not see it at the same time, and we **should** allow waiting at least another `confirmationBlocks` **on top of** minimal tx confirmation before resolving this request and allowing anyone depending on it (e.g. SP) to proceed with the transfers which need to assume that this request was completed and picked up by anyone else interested (partners, PFS).

Of course, this won't make us fail proof, if the ETH node keeps lagging behind, eventually not even `confirmationBlocks` will be enough for partner to see/confirm our deposit, but we should at least give some time, since the way it is now, it's implicitly assumed (and required) that partners must _less than 1 blocktime_ behind, which is a bold assumption for a distributed system. `confirmationBlocks` is already our margin of confidence for this kind of thing, we must just ensure that, in this case, it's applied also to confirmation of these time-critical operations.

_Confirm confirmation_ may sound weird, but this is what happens here: confirmation is when the tx is confirmed on-chain, and _confirming it_ is waiting a given _confidence threshold_ after it before resolving and proceeding with operations which may fail if partner didn't also saw that **first** confirmation.

## Steps to Reproduce
1. Have one one lag 2-4 blocks behind the other
2. Check that `openChannel` call resolves before partner could see/confirm these operations and react to them (e.g. by sending the respective `PFSCapacityUpdate`

## Expected Result
- Nodes are in sync and already reacted/confirmed when any of them resolve its request, while they're at most a reasonable confirmationBlocks=5blocks (default) lagging behind one another

## Actual Result
- Requester resolves before partner had time to pick up and react to this open/deposit tx.

andrevmatos:
Fixes #2813 

**Short description**
Check issue's description for a breakdown of the race. This PR doesn't touch state machine, and just wait at the public method level, so the only significant impact should be a +5 blocks (~1min15) increase in `open`/`depositChannel` requests, which is awaited before allowing API to resolve in order to ensure partners and services have enoguh time to react to those events.
On dApp, the user is able to close the dialog and get the notification/see the deposit got confirmed and proceed to perform the transfers, if they want, so this is more of a safeguard for the race on these on-chain operations.

**Definition of Done**

1. Check +5 blocks are awaited after **confirmed** `channel/deposit/success` before the deposit dialog is auto-closed

### DIFF
--- a/raiden-ts/src/raiden.ts
+++ b/raiden-ts/src/raiden.ts
@@ -602,13 +602,20 @@ export class Raiden {
    * @param options.settleTimeout - Custom, one-time settle timeout
    * @param options.subkey - Whether to use the subkey for on-chain tx or main account (default)
    * @param options.deposit - Deposit to perform in parallel with channel opening
+   * @param options.confirmConfirmation - Whether to wait `confirmationBlocks` after last
+   *    transaction confirmation; default=true
    * @param onChange - Optional callback for status change notification
    * @returns txHash of channelOpen call, iff it succeeded
    */
   public async openChannel(
     token: string,
     partner: string,
-    options: { settleTimeout?: number; subkey?: boolean; deposit?: BigNumberish } = {},
+    options: {
+      settleTimeout?: number;
+      subkey?: boolean;
+      deposit?: BigNumberish;
+      confirmConfirmation?: boolean;
+    } = {},
     onChange?: OnChange<EventTypes, { txHash: string }>,
   ): Promise<Hash> {
     assert(Address.is(token), [ErrorCodes.DTA_INVALID_ADDRESS, { token }], this.log.info);
@@ -630,14 +637,17 @@ export class Raiden {
     const deposit = !options.deposit
       ? undefined
       : decode(UInt(32), options.deposit, ErrorCodes.DTA_INVALID_DEPOSIT, this.log.info);
+    const confirmConfirmation = options.confirmConfirmation ?? true;
 
     const meta = { tokenNetwork, partner };
     // wait for confirmation
-    const openPromise = asyncActionToPromise(channelOpen, meta, this.action$, true).then(
+    const openPromise = asyncActionToPromise(channelOpen, meta, this.action$).then(
       ({ txHash }) => txHash, // pluck txHash
     );
-    let depositPromise;
-    let postPromise;
+    const openConfirmedPromise = asyncActionToPromise(channelOpen, meta, this.action$, true);
+
+    let depositPromise: Promise<Hash> | undefined;
+    let postPromise: Promise<unknown> | undefined;
     if (deposit?.gt(0)) {
       depositPromise = asyncActionToPromise(channelDeposit, meta, this.action$, true).then(
         ({ txHash }) => txHash, // pluck txHash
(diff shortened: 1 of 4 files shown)


raiden-ts/src/raiden.ts
```diff
@@ -652,21 +662,25 @@ export class Raiden {
...
+      onChange?.({ type: EventTypes.DEPOSITED, payload: { txHash: depositTxHash } });
       await postPromise;
     }
 
+    if (confirmConfirmation) {
+      // wait twice confirmationBlocks for deposit or open tx
+      await waitConfirmation(
+        await this.deps.provider.getTransactionReceipt(depositTxHash ?? openTxHash),
+        this.deps,
+        this.config.confirmationBlocks * 2,
```
weilbith, line 680: Just a nitpick, because that confused me, and I needed to inspect the
`waitConfirmation` function to understand it. Becaus we did already await the
`openPromise` which already includes 1x confirmation block numbers on top of the
tx-block. And now we wait again for 2x confirmation blocks. This works becuase
the function adds this number to the block number of the transaction receipt and
waits for this block. So for example if you call this function very late when
this block has been already mined it would resolve immediately. Maybe you could
put this detail into the function description? So a "hover" in the IDE already
reveals this secret.

weilbith (approved):
I think the implementation is good.
I'm not too sure if we should include this in the SDK directly. As you said does
the dApp for example not care about it. So maybe this is better to be done in
the tools on top of the SDK? In fact in the CLI wait for the open channel method
to resolve and then wait another `confirmation number` of blocks? Is there any
additional advantage to have it in the SDK? My point is that the SDK is the core
and is already very big and handles tons of issues for the protocol. I would
like to avoid including not necessary elements which could be handled
externally.

palango:
Mh, I wonder if this is the right approach for race conditions in the scenario player. Wouldn't it be easier to add some waits after opens/deposits there?

weilbith:
Hmm. I kinda agree with @palango. But to be honest I don't like both. I think
the SP will have a hard time to resolve this issue and probably shouldn't. Don't
add too much complexity to the test tool. But I'm also not too happy with having
it within the SDK. :shrug:

andrevmatos:
I thought thoroughly about the 3 approaches:
- implement it in the SDK
- implement it in the interface (e.g. cli)
- just add some wait in the SP

I ended up implementing it in the SDK, but on the public interface instead of the state machine, exactly because I understand as being a user's assumption that, once this method resolves, they should be **ready** to perform a transfer, and the race mentioned actually would prevent that also for dApp's users. If the user closes the dialog and try a transfer which can't succeed yet anyway, they'll know it failed because deposit was not ready. Nonetheless, we do our best to have the user _informed_ in due time that they now can proceed, by resolving it in due time.

raiden-ts/src/raiden.ts
andrevmatos: :heavy_check_mark:

PR #2816, opened 2021-06-15
chore: Add NOTICE file

palango:
Note and provide dependencies licenses.


taleldayekh (approved):
Great Paul, thank you for taking care of this! ❤️

palango:
I wonder if we could/should remove the version numbers

PR #2819, opened 2021-06-15
dapp: fix caching of too big vendor chunk

weilbith:
The workbox plugin for webpack does filters per default too big assets so they
don't get cached. Because our dependencies have grown, the vendor script chunk
hit this limit. As result the dApp was not working offline anymore as this asset
was missing.
The limit got increased quite high to prevent this issue from happening again.
Especially because it is so unexpected and hard to notice.


andrevmatos (approved):
lgtm

PR #2868, opened 2021-07-24
Check mediators for 'Mediate' capability when initiating transfer

andrevmatos:
Based on top of #2853

**Short description**
PFS doesn't handle capabilities, so it's up to us as initiators to validate that all mediators are able to mediate transfers, by checking they've truthy `Capabilities.Mediate`. Fortunately, PFS sends all hops metadata in `address_metadata`, which allows us to inspect and validate received route's capabilities. This should prevent us from even starting a transfer where some mediator doesn't want to mediate, but the best fix is to have this handled also in the PFS side.

**Definition of Done**

1. Trying to mediate through a node which doesn't want to (e.g. another dApp) should fail before starting transfer

### DIFF
--- a/raiden-ts/CHANGELOG.md
+++ b/raiden-ts/CHANGELOG.md
@@ -7,10 +7,12 @@
 
 ### Fixed
 - [#2831] Force PFS to acknowledge our capabilities updates
+- [#2868] Invalidate routes where some mediator have `capabilities.Mediate=0` and error if this is the only route received from PFS.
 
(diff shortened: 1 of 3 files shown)


palango:
Created the PFS issue: https://github.com/raiden-network/raiden-services/issues/1044

weilbith (approved):
As told, I only reviewed the last commit here.
Looks good to me. I'm optimistic that is works in practice. 😬

PR #2871, opened 2021-07-27
chore: extend CodeClimate ignore rules

weilbith:
Unfortunately we need to commit the build directory for the GitHub actions into
the repository. This obviously comes with the issue that CodeClimate has tons of
reports for these files. As there are no actual code we can safely ignore them
and clean up the CodeClimate report.


palango (approved):
👍

PR #2883, opened 2021-08-06
Bugfix/fix ci staging workflow for dapp e2e tests

### LINKED ISSUE #2877: Fix CI publish staging workflow
The CI pipeline works for the default workflow. But it fails for the `publish_staging` workflow when executing the dApp end-to-end tests. The only obvious difference for the tests between both pipelines is that one runs with an `e2e` mode build of the dApp, while the other uses the `staging` mode. That might have an impact. But a first analysis reveals an issue that pops up way earlier in the test flow. In fact Cypress fails to load the dApp at all. It can't even access the very landing page.
This is especially awkward for two reasons. First does Cypress initially check if there is a server running under the configured URL. If there is no process listening, Cypress will fail even earlier and reports this specific error. Second, I re-run the job in CI with an SSH connection to the machine. I turned on the logging (`set -x`) for the `run-cypress.sh` script. Everything looks fine. And if I `curl` the dApp while the Cypress tests are running, I successfully get the `index.html` file. So after all the web-server is working and it serves the build `staging` version of the dApp. Though something in the setup phase of the VueJS application loaded in the `index.html` does not seem to work. Unfortunately we are still unable to see the `console` of the dApp within the videos of the tests. Therefore we need to check if we can reproduce the issue locally with a `staging` build and then debug it. We might need to add some missing environment variables to the mode.

weilbith:
Fixes #2877

Includes time optimization of the whole CI workflows.
@palango you know the dirty details of my first approach to fix it. 😜

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -277,8 +277,23 @@ workflows:
       - build_sdk:
           requires:
             - install
+      - test_sdk_unit:
+          requires:
+            - install
+      - test_sdk_integration:
+          requires:
+            - install
+      - test_sdk_e2e:
+          requires:
+            - install
       - build_dapp:
-          mode: e2e
+          mode: development
+          requires:
+            - build_sdk
+      - test_dapp_unit:
+          requires:
+            - build_sdk
+      - test_dapp_e2e:
           requires:
             - build_sdk
       - build_cli:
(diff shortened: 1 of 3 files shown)


andrevmatos (approved):
lgtm, thanks for these huge improvements. would it be possible to build the sdk as part of `yarn post-install` script or something like that, so we could merge 3rd stage to 2nd? notice that a lot of the time between these two (install and build_sdk jobs) is actually overhead, loading the whole persisted workspce/etc. Just a suggestion, you can merge it as is

weilbith:
Thanks for the review. I'm actually right now working on further improvements for our pipeline. I'm especially checking out improvements for monorepos. 
To be honest I don't see really where improvement from your suggestions comes. Or just for saving having it in a separate job with an environment spinup?

As there will be more PR(s) coming for my other improvements, I would merge that one here now and see how we can work on your suggestion in the upcoming ones.

PR #2885, opened 2021-08-06
chore: fix staging CI workflow

weilbith:
That must be it...
The only difference to before and because the job failed due to the missing `dist` folder.


andrevmatos (approved):
lgtm

weilbith:
I force merged as this is all about the staging workflow

PR #2901, opened 2021-08-24
Delay non-closing auto-settle to prevent wasted gas on channelSettle race

### LINKED ISSUE #2798: Offset `settle` calls from channel partners
When closing a channel and reaching the end of the settlement window, both channel partners will send a settlement transaction to the blockchain.

As those coincide, only one will go through, while one will be rejected. This could be observed while running the BF1 scenario.

- Successful tx: https://etherscan.io/tx/0x08d7125341f12b114d128271f9b37cc4645d825656bee8dc284655e88c083e8b
- Failed tx: https://etherscan.io/tx/0x55ac81644439c97447b8bd10c67f6c616e1fdfe6f2ee8c8e46e80739e171eaff

To limit failing transaction we could offset the transactions ending by the partner with the higher address by a number of blocks.

andrevmatos:
Fixes #2798 

**Short description**
Optimization: auto-settling peers have shifted timeouts after settleable block; closing side is expected to auto-settle only `confirmationBlocks=5` after channel becomes settleable; if it doesn't, non-closing side auto-settle is delayed to `revealTimeout=50` blocks; unless partner is a PC, which always settle early, so in this case LC always wait longer before attempting auto-settling.
This tries to prevent both clients calling `settleChannel` at the same time and one of them wasting the gas, as only one of the txs will succeed.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/channels/epics/settle.ts
+++ b/raiden-ts/src/channels/epics/settle.ts
@@ -31,7 +31,7 @@ import type { Hash, HexString } from '../../utils/types';
 import { channelSettle, channelSettleable, newBlock } from '../actions';
 import type { Channel } from '../state';
 import { ChannelState } from '../state';
-import { assertTx, channelKey, groupChannel$ } from '../utils';
+import { assertTx, channelKey, groupChannel } from '../utils';
 
 /**
  * Process newBlocks, emits ChannelSettleableAction if any closed channel is now settleable
@@ -81,7 +81,7 @@ export function channelAutoSettleEpic(
   { config$ }: RaidenEpicDeps,
 ): Observable<channelSettle.request> {
   return state$.pipe(
-    groupChannel$,
+    groupChannel(),
     mergeMap((grouped$) =>
       grouped$.pipe(
         filter(
--- a/raiden-ts/src/channels/epics/settle.ts
+++ b/raiden-ts/src/channels/epics/settle.ts
@@ -16,13 +16,12 @@ import {
 
 import type { RaidenAction } from '../../actions';
 import { intervalFromConfig } from '../../config';
-import { Capabilities } from '../../constants';
 import { chooseOnchainAccount, getContractWithSigner } from '../../helpers';
 import { createBalanceHash } from '../../messages/utils';
 import type { RaidenState } from '../../state';
 import { Direction } from '../../transfers/state';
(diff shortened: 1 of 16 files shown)


raiden-ts/src/channels/epics/settle.ts
```diff
@@ -66,56 +67,56 @@ export function channelSettleableEpic(
...
+      { blockNumber, channels },
+      presences,
+      { confirmationBlocks, revealTimeout },
+    ]) {
+      for (const channel of Object.values(channels)) {
+        if (channel.state !== ChannelState.settleable) continue;
+
+        const partnerIsOnlineLC =
+          channel.partner.address in presences &&
+          !getCap(presences[channel.partner.address]?.payload.caps, Capabilities.DELIVERY);
```
palango, line 101: Do we have such checks (having delivery cap on signals python client) in other places? If so, we should factor them out into one place.

andrevmatos: Yes, I think in one other place. I'll make a new commit to split this out in a separate function.

raiden-ts/src/channels/epics/settle.ts
```diff
@@ -66,56 +67,56 @@ export function channelSettleableEpic(
...
+
+        /* iff we are *sure* partner will wait longer (i.e. they're a LC and *we* are the closing
+         * side), then we autoSettle early; otherwise (it's a PC or they're the closing side),
+         * we can wait longer before autoSettling */
+        let waitConfirmations: number;
+        if (partnerIsOnlineLC && channel.closeParticipant === address)
+          waitConfirmations = confirmationBlocks;
+        else waitConfirmations = revealTimeout;
```
palango, line 109: Shouldn't this be `revealTimeout + confirmationBlocks`?

palango (approved):
Great, simple solution. 

LGTM after the small refactoring.

raiden-ts/src/channels/epics/settle.ts
andrevmatos: No need.. `revealTimeout` is just a bigger value than `confirmationBlocks`, to shift the delay; no need to _confirm_ it

raiden-ts/src/channels/epics/settle.ts
andrevmatos: :heavy_check_mark:

PR #2931, opened 2021-09-08
Enable stricter TS rules for indexed properties access

andrevmatos:
This would have prevented the issue fixed by #2925 

**Short description**
This enables TS@4.1's `--noUncheckedIndexedAccess` and TS@4.2's `--noPropertyAccessFromIndexSignature` on src only, which should allow us to more strictly type undefined property access on indexed type, requiring us either to type-guard it or to consciously mark it as defined.
There should be no logic change on this PR, just type-strictness fixes.

**Definition of Done**

1.
2.


palango (approved):
Nice, more checks did never hurt anybody :)

PR #2937, opened 2021-09-09
Update contracts to 0.40.0rc

palango:
@andrevmatos Should I update the e2e image right away?

I also didn't update the code to automatically parse the version because I didn't tag the contracts release.


andrevmatos (approved):
@palango yes, it'd be nice to update the e2e image, if we want to have our e2e tests covering these new contracts (and since they're mostly the only inter-client compliance tests, I think that may be desirable). Thank you!

andrevmatos (approved):
lgtm!

PR #2942, opened 2021-09-13
Change the filter tag regex so that publish_production workflow also works for pre-releases

### LINKED ISSUE #2899: CI didn't publish pre-release to npm
CI skipped task to build and publish the `v2.0.0-rc.1` pre-release to npm. Not sure if because it was a pre-release version or it stopped publishing to npm completelyt since #2885 . I did the publishing manually, but we need to investigate and fix the automated workflow.
image

agatsoh:
Change the filter tag regex so that publish_production workflow also works for pre-releases

Fixes #2899 

**Short description**

**Definition of Done**

1.
2.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ anchor_1: &attach_options
 anchor_2: &filter_release_tag
   filters:
     tags:
-      only: /^v\d+\.\d+\.\d+$/
+      only: /^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/
 
 anchor_3: &executor_shared_options
   working_directory: << pipeline.parameters.working_directory >>


.circleci/config.yml
```diff
@@ -23,7 +23,7 @@ anchor_1: &attach_options
 anchor_2: &filter_release_tag
   filters:
     tags:
-      only: /^v\d+\.\d+\.\d+$/
+      only: /^v\d+\.\d+\.\d+(?:-rc\.\d)?$/
```
weilbith, line 26:
```suggestion
      only: /^v\d+\.\d+\.\d+(?:-rc\d+)?$/
```

weilbith (changes requested):
Hey, thanks for taking your time and investigating here. 🙏🏾 
I have to comment on your regex:
  - why the `\.` between the `-rc` and the digit of the release candidate?
  - nothing prevents to have us more than 10 release candidates for a single version, so `\d+` here is better.

.circleci/config.yml
andrevmatos:
```suggestion
      only: /^v\d+\.\d+\.\d+(?:-rc\.?\d+)?$/
```

weilbith: Why that? 🤔

andrevmatos: the PC uses the dot between rc and digit; although our latest pre-release doesn't, it was the first one, so I thought just allowing that on this regex wouldn't hurt. To be more precise, actually, I'd prefer a regex which would match ANY valid [pre-release tag from semver](https://semver.org/#spec-item-9)

andrevmatos: From [semver's docs](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string)
```suggestion
      only: /^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$/
```

andrevmatos: Amend: our pre-release has the dot, so we should comply

weilbith: Alright. So it should be your last expression from `semvers` plus the optional dot?

agatsoh: @weilbith The above regex change that you see is directly coming from the [semver docs](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string) and satisfies all the semver release, pre-release criteria. Hence I will go with it.

weilbith (approved):

andrevmatos (approved):
lgtm; tested the regex with our most common tag patterns, and they seem fine, so it should work

PR #2957, opened 2021-09-23
Update user message to deposit the required service tokens by keeping a 10 percent margin

### LINKED ISSUE #2956: Change message on Manage UDC tokens tab for RDN deposit
## Steps to Reproduce
1. 
2. 
Please refer to this issue for the backdrop 
https://github.com/raiden-network/raiden-services/issues/1076#issuecomment-925819693

## Expected Result
- 
With a deposit of  only 80 RDN the MS did not kick in. This is not a bug per say but a gap in conveying to the user the exact amount of RDN that needs to be deposited.
## Actual Result
- Deposit of 80 RDN is now insufficient in the UDC for the MS to fire. It should be minimum 88 RDN or such suitable number after which user would not face issues with MS.

agatsoh:
For a hindsight into this issue please check this issue out 
https://github.com/raiden-network/raiden-services/issues/1076

Fixes #2956  

**Short description**

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-dapp/src/views/account/UDC.vue
+++ b/raiden-dapp/src/views/account/UDC.vue
@@ -81,6 +81,8 @@ import type { PlannedUdcWithdrawal } from '@/store/user-deposit-contract';
 const { mapState: mapStateUserSettings } = createNamespacedHelpers('userSettings');
 const { mapState: mapStateUserDepositContract } = createNamespacedHelpers('userDepositContract');
 
+const UDC_SECURITY_MARGIN_PERCENTAGE = 10;
+
 @Component({
   components: {
     ActionButton,
@@ -103,7 +105,9 @@ const { mapState: mapStateUserDepositContract } = createNamespacedHelpers('userD
 })
 export default class UDC extends Vue {
   amount = utils.formatEther(DEFAULT_MS_REWARD.mul(2));
-  minAmount = +utils.formatEther(DEFAULT_MS_REWARD);
+  minAmount = +utils.formatEther(
+    DEFAULT_MS_REWARD.add(DEFAULT_MS_REWARD.mul(UDC_SECURITY_MARGIN_PERCENTAGE).div(100)),
+  );
   udcCapacity = constants.Zero;
   hasEnoughServiceTokens = false;
   blockNumber!: number;


palango (changes requested):
This is wrong. We don't want to increase the reward amount, but need to take the 10% margin into account.

agatsoh:
@palango ok I got will change this

raiden-ts/src/constants.ts
```diff
@@ -38,3 +38,4 @@ export const RAIDEN_DEVICE_ID = 'RAIDEN';
 export const UDC_WITHDRAW_TIMEOUT = 100;
 export const DEFAULT_CONFIRMATIONS = 5;
 export const DEFAULT_MS_REWARD = parseEther('80') as UInt<32>;
+export const UDC_SECURITY_MARGIN_PERCENTAGE = 10;
```
andrevmatos, line 41: If this constant isn't used in SDK, please, don't add it here. SDK already has all information it needs. If you want it on the dApp, add it there

raiden-dapp/src/views/account/UDC.vue
```diff
@@ -103,7 +103,9 @@ const { mapState: mapStateUserDepositContract } = createNamespacedHelpers('userD
 })
 export default class UDC extends Vue {
   amount = utils.formatEther(DEFAULT_MS_REWARD.mul(2));
-  minAmount = +utils.formatEther(DEFAULT_MS_REWARD);
+  minAmount = +utils.formatEther(
+    DEFAULT_MS_REWARD.add(DEFAULT_MS_REWARD.div(UDC_SECURITY_MARGIN_PERCENTAGE)),
```
andrevmatos, line 107: This is wrong. You're adding the original value divided by a variable called `PERCENTAGE`. It only works here because `10%` also happens to be `1/10` of `100%`. Either you use some logic like `orig.add(orig.mul(percentage).div(100))`, or rename it to something like `DIV_FACTOR` or so.

andrevmatos (changes requested):

raiden-dapp/src/views/account/UDC.vue
agatsoh: Done

raiden-ts/src/constants.ts
agatsoh: Removed from the SDK

agatsoh:
@andrevmatos I have made the required changes please have a look.

andrevmatos (approved):

PR #2967, opened 2021-10-09
Fix parallel open + deposit and disable channel `closing` if Metamask prompt is cancelled

### LINKED ISSUE #2963: dApp/SDK: Prevent Light Client from entering a closing state when user closes a channel but rejects MetaMask prompt
## User Story

As Chris I want my wallet confirmation, e.g. MetaMask to be the final say in whether a channel closing should happen or not.

While doing user testing it became obvious that in a real world use-case a user familiar with rejecting transactions in their wallet expects to be able to do so when using the LC as well despite the confirmation dialogs in the dApp.

We should make adjustments to better meet the users expectations in this case and not put the channels in a unusable state when a channel close is not actually happening.

Since MetaMask in this case is the last confirmation step it makes sense that it should be the final say on whether a channel enters a closing state or not.

andrevmatos:
Fixes #2963
Also fixes an issue seen in highly-concurrent scenarios like BF7

**Short description**
See commits messages for descriptions of what got fixed.

**Definition of Done**

1.
2.


raiden-ts/src/channels/reducer.ts
```diff
@@ -157,23 +157,19 @@ function channelCloseSuccessReducer(
   return { ...state, channels: { ...state.channels, [key]: channel } };
 }
 
-function channelUpdateStateReducer(
+function channelSettleableStateReducer(
   state: RaidenState,
-  action: channelClose.request | channelSettle.request | channelSettleable,
```
palango, line 162: Mh, shouldn't we still set the channel state to 'closing' when the request is issued but revert this when the metamask tx is rejected?

Like this it would be possible to send/mediate payments while we're waiting for the tx to be mined - which is not what we want.

palango: Yes, but I think we need to somehow block transfers when we prepare the close transaction, because otherwise the state we transact will be outdated and wrong.

Can't we create some 'inactive' state, than can then be used for coop-settlement and this?

palango: I think this is the best way. How do we handle this for coop settle right now?

andrevmatos: Coop-settle doesn't incur `channel.state` change, the _locking_ is done through withdraw locked balance (available balances become zero, therefore no transfer can be sent or received, same as withdraw)

andrevmatos: Done :heavy_check_mark:

PR #2997, opened 2021-11-16
Update README.md

ezdac:
The workspace name has to be prefixed with `@raiden_network/` in order to work from root dir


andrevmatos (approved):
nice catch, thanks

PR #3008, opened 2021-11-24
Log invalid PFS

### LINKED ISSUE #3001: PFS's `ServiceRegistry.hasValidRegistration` is not verified
## Description

The node will only query the URL registered on-chain for a given PFS address but doesn't verify `ServiceRegistry.hasValidRegistration` for that address.
If that PFS did register in the past, but the registration expired, the contract will still return the URL.

In that case the node thinks that the PFS is functioning properly, but the transport server will not forward TO-DEVICE
messages to that PFS, since the transport server requires the service to have a valid registration in order to being whitelisted.

https://github.com/raiden-network/light-client/blob/b51dff44e492e757992ce72171ef76cab7621984/raiden-ts/src/services/utils.ts#L46-L53

## Acceptance criteria
- `hasValidRegistration` is polled on initialisation time
- eventually the expiry is memorised and the service disregarded on runtime at the expiry time.
- then a new PFS is used, or a warning emitted to the user if no other PFS could (should) be found

andrevmatos:
Fixes #3001

**Short description**
We add PFS's validity timestamp to `PFS`/`RaidenPFS` schema, test it and log a big warning in case the user tries to use an invalid one (either by specifying in config or one-shot in methods `options.pfs`). Since these are advanced options only, we only log and try to use them anyway (as we did before), trusting the user knows what they're doing. For the normal case, `auto` services list already take care of only having/showing valid services.
No changelog needed since this doesn't change user-facing behavior, just add a logging for an advanced edge case.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/services/utils.ts
+++ b/raiden-ts/src/services/utils.ts
@@ -119,22 +119,24 @@ export async function pfsInfo(
 
     const start = Date.now();
     const res = await fetch(url + '/api/v1/info', { mode: 'cors' });
+    const rtt = Date.now() - start;
     const text = await res.text();
     assert(res.ok, [ErrorCodes.PFS_ERROR_RESPONSE, { text }]);
     const info = decode(PathInfo, jsonParse(text));
 
-    assert(info.price_info.lte(pfsMaxFee), [
-      ErrorCodes.PFS_TOO_EXPENSIVE,
-      { price: info.price_info.toString() },
-    ]);
+    const { payment_address: address, price_info: price } = info;
+    assert(price.lte(pfsMaxFee), [ErrorCodes.PFS_TOO_EXPENSIVE, { price }]);
     pfsAddressCache_.set(url, Promise.resolve(info.payment_address));
+    const validTill =
+      (await serviceRegistryContract.callStatic.service_valid_till(address)).toNumber() * 1e3;
 
     return {
-      address: info.payment_address,
+      address,
       url,
-      rtt: Date.now() - start,
-      price: info.price_info,
+      rtt,
+      price,
       token: await serviceRegistryToken(serviceRegistryContract, provider.pollingInterval),
+      validTill,
     };
   } catch (err) {
     log.warn('Error fetching PFS info:', pfsAddrOrUrl, err);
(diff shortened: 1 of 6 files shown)


raiden-ts/src/services/utils.ts
```diff
@@ -119,22 +119,24 @@ export async function pfsInfo(
...
 
-    assert(info.price_info.lte(pfsMaxFee), [
-      ErrorCodes.PFS_TOO_EXPENSIVE,
-      { price: info.price_info.toString() },
-    ]);
+    const { payment_address: address, price_info: price } = info;
+    assert(price.lte(pfsMaxFee), [ErrorCodes.PFS_TOO_EXPENSIVE, { price }]);
     pfsAddressCache_.set(url, Promise.resolve(info.payment_address));
+    const validTill =
+      (await serviceRegistryContract.callStatic.service_valid_till(address)).toNumber() * 1e3;
```
ezdac, line 131: This does an RPC call to the ETH-node.
I don't fully understand what this code does:

https://github.com/raiden-network/light-client/blob/b51dff44e492e757992ce72171ef76cab7621984/raiden-ts/src/services/epics/helpers.ts#L211
Will the `pfsInfo` function execute on every path-request and thus make a JSON-RPC call to the eth node every time?
If not I have a follow up question:
- Do you update the `validTill` at some point like you do with the auto-discovered PFSes?
Because if the PFS calls `deposit` again to prolong the registration this would not be updated.

ezdac (approved):
Looks good to me!
I'm not too familiar with the rxjs functional approach so I added a question regarding observables.

raiden-ts/src/services/utils.ts
andrevmatos: This calls RPC every time a transfer is about to be made, because we also need to fetch up-to-date PFS info (in case they updated the price, etc), so there's no caching here.

ezdac: Yes, for the PFS-HTTP endpoint it makes sense - but technically you could easily spare at least the ethnode RPC-call here and rather update the validTill from emitted blockchain events upon `deposit` calls. This is just an optimisation - I'm just mentioning this since we saw quite some RPC calls to the ethnode in the python client that could be avoided and the ethnode is one of the bottlenecks of concern.

But in any case, please go ahead with merging this when you consider not worth it (right now).

andrevmatos: Good call. We happen to actually already monitor and persist the mapping of valid registered services in `state.services` (monitored from the event), so it was actually just a matter of checking if the service address is in the dict and fetching from the static contract call just as a fallback. Thanks for the headsup, we really need to look on those unneeded calls more closely.

andrevmatos: Done :heavy_check_mark:

PR #3023, opened 2021-12-24
Mainnet Release v2.0.1 Kṛttikā

weilbith:
# 💉 New Raiden Light Client SDK, dApp and CLI

> **INFO:** The Light Client SDK, CLI and dApp are all **work in progress** projects. All three projects have been released for **mainnet** and all code is available in the Light Client repository. As this release still has its limitations and is a **beta** release, it is crucial to read this readme including the security notes carefully before using the software.

This is a patch release to fix the update feature of the dApp. Users  of the officially hosted https://lightclient.raiden.network client should ensure to update from `v1.1.0` directly to `v2.0.1`. Please remember that this is still a major release jump and implies a switch to the new contracts. For more information, refer to the notes from the [v2.0.0 release](https://github.com/raiden-network/light-client/releases/tag/v2.0.0).

Raiden dApp
-------------------------------
### Fixed

- [#3021] Fix version handling to detect client updates

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "light-client",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "description": "Raiden Light Client monorepo",
   "author": "brainbot labs est.",
   "private": true,
--- a/raiden-cli/CHANGELOG.md
+++ b/raiden-cli/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+## [2.0.1] - 2021-12-24
+
 ## [2.0.0] - 2021-12-23
 ### Added
 - [#2949] Passthrough `/payments` parameters, including `paths`, which should receive pre-fetched route in the format `{ route: Address[]; estimated_fee: NumericString; address_metadata?: MetadataMap }[]`.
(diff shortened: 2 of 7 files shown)


weilbith:
@andrevmatos please only approve the PR. I'll double check it again and would like to be present when we merge and tag it so I can triple check it then when it becomes live.

andrevmatos (approved):
Lgtm, ty

weilbith:
I made the double check. Something is still broken.

weilbith:
I found the cause for what was distracting me. Lesson learned.
When the user **force** reloads the tab (`Shift` + `F5` or `Ctrl` + `R`) this will bypass the service worker and its cache (see W3C reference below). As result the most recent version from the server gets loaded without asking the user for confirmation. But as our service worker works, it will install the new service worker but still uses the old cache. This means when the user closes the tab and opens the client again, the browser will use the service worker again which will use its "old" cache to open the client.

This means the update feature got successfully fixed for this patch version. But it revealed a new issue we have to think about and fix. Because the problem could be that the user is actually using this force-reloaded new version to act on new contracts that gets persistent to the database. Or the SDK updated its database structure and migrated the old data into the new structure. Now this new version is only active temporarily. The next time the user loads the client it will have either broken the client or the user is confused as he is missing channels.

---

From the [W3C specification for service workers](https://www.w3.org/TR/service-workers/#navigator-service-worker-controller.1):
> navigator.serviceWorker.controller returns null if the request is a force refresh (shift+refresh).

weilbith:
I had a call with Kamal. As this will fix the current issue with having a version that is able to update to future versions again, we will release it as planned. The other issue still exists and needs to be tackled afterwards.

PR #3025, opened 2021-12-29
Feature/dapp prevent to load version directly from server

### LINKED ISSUE #3024: dApp - force reload update short circuit service worker
## Description

According to my comment [here](https://github.com/raiden-network/light-client/pull/3023#issuecomment-1001479369) (includes link to W3C specification) a force-reload of the light client can lead to serious issues. Quickly rephrasing, a force-reload will bypass the service worker and loads the assets directly from the server. This means the user is getting the most recent version available on the server, no matter which version was running before. The user probably does not recognize this in first place. There is no notification about an update or anything else. Anyway, as soon as the user closes the tab where he force-reloaded the client and then load the client later again, it will load the former "old" version again as usual. This is because our service worker will always serve the locally available version if the user does not manually trigger the update. Though the force-reload short circuit the service worker and its cache storage, it will keep the cache intact (which the "old" version). In result the user was using the new version for a temporally time frame.

There can be a couple of serious issues because of that. Two obvious ones are the contracts version and the database version. For the first case, the user will load a new version of the client that is linked to a new version of contracts. He will be confused because his old channels are not there anymore. But maybe he doesn't even recognize this and just blindly follows the quick pay or similar. Anyway, when he returns to the old version he also gets back his old channels. This leads to a lot of confusion and frustration, spending unnecessary many for gas costs etc. 
The second issue is about the database scheme version of the SDK to persist the client data. New versions of the client can include a new version of the database scheme. The SDK will automatically migrate the old data to the new scheme. But if the user gets back to the old version from his short trip to the new version, the SDK will fail to load the data from the new scheme it does not know about yet. The exact implications are to be cleared, though it is probably clear that the user will not be able to access his channels anymore.

In my opinion there are two strategies to handle such a situation: either ensure a force-reload performs a full mature update or prevent the update. Both approaches come with the challenge how to detect it. I can only repeat myself: what we do here is not a common thing and we have to fight against the current.

## Acceptance criteria
- decide on strategy to solve the problem
- figure out how to detect this issue in the application
- implement handling of issue
- establish end-to-end tests

## Tasks
- [ ]

weilbith:

Fixes #3024

**Short description**
Checkout the issue and the commit messages for details.

**Definition of Done**

1. Build productive version of the dApp for current version (`yarn build`)
2. Service build locally (e.g. `npx serve ./dist`)
3. Ensure the browser has no data on the now served location (in Chrome use dev-tools application, `Clear Site Data`)
4. Load current version of the local dApp and wait for it to install
5. Adapt the version to a higher number in `package.json` (e.g. `3.0.0`)
6. Rebuild the dApp (`yarn build`)
7. Force-reload the dApp in the browser (`Shift` + `F5` in Chrome)
8. Ensure that the dApp became temporally blocked (looks a bit squishy)
9. Ensure that the dApp has reloaded itself again
10. Ensure the dApp now shows a version update and the loaded version is the "old" one (at the bottom of the account screen)


raiden-dapp/src/store/user-settings/index.ts
```diff
@@ -1,23 +1,33 @@
...
+ * @param storage - object implementing the Storage interface to persist the
+ *                  state (defaults to `window.localStorage`)
+ * @returns plugin for the root store to persist data
+ */
+export function createUserSettingsPersistencePlugin(
+  storage?: Storage,
+): Plugin<RootStateWithUserSettings> {
+  const vuexPersistence = new VuexPersistence<RootStateWithUserSettings>({
+    key: 'raiden_dapp_settings',
+    storage: storage ?? window.localStorage,
```
andrevmatos, line 26: Nitpick/suggestion:
```suggestion
  storage: Storage = window.localStorage,
): Plugin<RootStateWithUserSettings> {
  const vuexPersistence = new VuexPersistence<RootStateWithUserSettings>({
    key: 'raiden_dapp_settings',
    storage,
```

andrevmatos (approved):
Some nitpicks/suggestions, but LGTM overall. The test suite is quite nice as well, but since this uses a quite involved and not very well documented environment (service workers, caches et al), we need to do some good testing by hand as well before releasing. Thank you very much for taking the time to do this whole improvement.

raiden-dapp/src/service-worker/assistant.ts
```diff
@@ -2,75 +2,161 @@
...
+      this.serviceWorkerContainer.controller?.postMessage(message);
+    }
+  }
+
+  private async unregisterAllServiceWorker(): Promise<void> {
+    const allServiceWokers = (await this.serviceWorkerContainer?.getRegistrations()) ?? [];
+
+    for (const serviceWorker of allServiceWokers) {
+      await serviceWorker.unregister();
     }
```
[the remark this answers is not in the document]
weilbith: I was thinking about something like this but was too lazy. 😆 
But because it is about performance here so the user has not much lagging it is important! 👍🏾

raiden-dapp/src/store/user-settings/index.ts
weilbith: ✔️

raiden-dapp/src/service-worker/assistant.ts
```diff
@@ -2,75 +2,161 @@
...
+    } else {
+      throw new Error(`Maleformed installation version: ${version}`);
+    }
+  }
+
+  private setAvailableVersion(version: string): void {
+    if (validateVersion(version)) {
+      this.store.commit('versionInformation/setAvailableVersion', version);
+    } else {
+      throw new Error(`Maleformed available version: ${version}`);
```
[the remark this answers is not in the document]
weilbith: ✔️

raiden-dapp/src/service-worker/assistant.ts
```diff
@@ -2,75 +2,161 @@
...
+  }
 
-  private setUpdateIsMandatory = (): void => {
-    this.store.commit('setUpdateIsMandatory');
-  };
+  private setInstalledVersion(version: string): void {
+    if (validateVersion(version)) {
+      this.store.commit('versionInformation/setInstalledVersion', version);
+    } else {
+      throw new Error(`Maleformed installation version: ${version}`);
```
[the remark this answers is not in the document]
weilbith: ✔️

raiden-dapp/src/service-worker/assistant.ts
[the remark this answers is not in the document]
weilbith: ✔️

weilbith:
I targeted all review comments with fixup commits. I'm confident about them, so I go ahead and resolve them. Then we have a state we can be subject of some final practical tests before we merge.

weilbith:
Sorry, for the linting mistake. Rooky mistake...

weilbith:
We made a more production realistic test. And we discovered an issue. I made a mistake an forgot that service workers must be able to understand messages of the old application and send messages to the old application **and** new one understands them. This is a complex issue.

weilbith:
Because I haven't worked with schema versioning for a while I read [this book](https://leanpub.com/esversioning/read) tonight to recap a little. The used approach is explained in much detail within the related commit message. Please check it out.
I made the tests that failed yesterday locally and forced a new service worker version to work with the old client (`v1.0.0`). It works as expected. It can detect a broken cache and also an update is possible to the new version. The same goes for the new version of both components.

raiden-dapp/src/service-worker/assistant.ts
```diff
@@ -57,37 +66,41 @@ export default class ServiceWorkerAssistant {
...
-        break;
-
-      case ServiceWorkerMessageIdentifier.INSTALLATION_ERROR:
-        this.reportInstallationError(payload.error);
-        break;
-
-      case ServiceWorkerMessageIdentifier.CACHE_IS_INVALID:
-        this.setUpdateIsMandatory();
-        break;
+    this.messageHandlers[message.type]?.(message.payload);
```
andrevmatos, line 78: You sure this will call the method with `this` set? The methods make use of `this`, and they're set in the mapping; I'm not sure this will preserve `this` in the context, maybe it'd be better to `.call` just to be on the safe side

andrevmatos: Oh, right, just noticed those are arrow functions; still, it'd be good to ensure which `this` is persisted when the arrow functions are defined in the body of classes

raiden-dapp/src/service-worker/messages.ts
```diff
@@ -1,17 +1,90 @@
...
   UPDATE = 'update',
   VERIFY_CACHE = 'verify_cache',
+  UNKNOWN = 'unknown', // Used for messages which could not be parsed.
+}
+
+export type ServiceWorkerMessagePayload = Record<string, unknown>;
+
+export type ServiceWorkerMessageHandler = (payload: ServiceWorkerMessagePayload) => void;
+
+export abstract class ServiceWorkerMessage {
```
andrevmatos, line 15: Wow, this is a quite involved approach. I'd have preferred a simpler one, we don't expect these messages to change that often or again need to break backwards compatibility, but if it works, let's keep it like that

andrevmatos (approved):
lgtm, let's test

raiden-dapp/src/service-worker/assistant.ts
weilbith: They are actually arrow functions in first place so I can say they are of that specific type...

weilbith: So what would you recommend to do? 🤔

andrevmatos: As you need to specify the function type explicitly in order to ensure it matches a given protocol, the arrow function is fine. Maybe just add the `this` type specifier in the `ServiceWorkerMessageHandler` type, and use `.call` to make explicit your intention/API here, even though they shouldn't matter much with arrow funcitons.

raiden-dapp/src/service-worker/messages.ts
weilbith: Yes it is involved. I just want to solve this problem once and for all now. If we go with a cheap solution it becomes harder and harder each time. Because I missed the signs (hello we talk about messaging here, typical CS problem) it already includes the exception now. It would have become worse over time. Back then I also thought there will not much change to it. But it happened. So the best I can do is to keep the code open to changes I don't know about yet and might think they never happen. But as it is about compatibility here to old versions we can't touch anymore it is a serious business we can't just do when it really becomes super urgent/necessary.
Therefore I really like to keep this. As long as you think it does what it should (wrapped mapping with a week schema).

raiden-dapp/src/service-worker/assistant.ts
weilbith: Do `.call(this, message.payload)`?

weilbith: Alright, I'll do. Let's make it bulletproof👍🏾

weilbith: I was so rude and amended the last commit and force pushed. I was too lazy to push it as fixup commit and then rebase again etc.
Done. ✔️

weilbith:
Andre made the suggestion that we can also protect all versions before the new release. We planned the design and I'm implementing it.

weilbith:
Our production close tests worked very well. We even made some last UI improvements. Waiting for CI now.

andrevmatos (approved):
lgtm, thank you for doing this in a so complete manner

PR #3027, opened 2021-12-30
Mainnet Release v2.1.0 Kṛttikā

weilbith:
# 🧱  New Raiden Light Client SDK, dApp and CLI

> **INFO:** The Light Client SDK, CLI and dApp are all **work in progress** projects. All three projects have been released for **mainnet** and all code is available in the Light Client repository. As this release still has its limitations and is a **beta** release, it is crucial to read this readme including the security notes carefully before using the software.

This new version adds a feature to the dApp that protects the user from loading temporary updates unintentionally. Besides the side effect of loading a different version temporary it is important that users always confirm updates and manually trigger the installation of a new version. Unintended updates can for example happen on a force-reload of of the web-browser. For more information, refer to the notes from the [v2.0.0 release](https://github.com/raiden-network/light-client/releases/tag/v2.0.0).

Raiden dApp
-------------------------------
### Added

- [#3024] Prevent the usage of uninstalled versions

PR #3028, opened 2021-12-31
dapp: fix loading for the first time with new version

weilbith:
Solves the issues detected in the end-to-end test of #3027 where the version got bumped over the critical threshold. See the commit messages for details of this bug.

### DIFF
--- a/raiden-dapp/src/store/version-information/getters.ts
+++ b/raiden-dapp/src/store/version-information/getters.ts
@@ -1,6 +1,8 @@
 import { compare as compareVersions, validate as validateVersion } from 'compare-versions';
 import type { GetterTree } from 'vuex';
 
+import { DISCLAIMER_STORAGE_KEY } from '@/store/constants';
+
 import type {
   RootStateWithVersionInformation,
   VersionInformationGetters,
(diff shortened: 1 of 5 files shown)


weilbith:
No CHANGELOG entry necessary for this.

raiden-dapp/src/store/version-information/getters.ts
```diff
@@ -26,21 +28,33 @@ function compareSemanticVersions(
...
+// This detection mechanism here isn't very nice. It is not possible to access
+// the data from the root store as it does not allow to distinguish the state.
+// Only when the user has accepted the disclaimer in the past, this data
+// exists. This also means this check here only works when evaluated before the
+// user accepts the disclaimer for the first time. In general it would be better
+// to have any other mechanism to detect the same information. Unfortunately
+// this must work with old versions of the client. This is the only and final
+// idea we came up with.
+function userLoadedApplicationForTheFirstTime(): boolean {
+  return localStorage.getItem(DISCLAIMER_STORAGE_KEY) === null;
```
andrevmatos, line 40: Maybe just check:
```suggestion
  return !window.localStorage.getItem(DISCLAIMER_STORAGE_KEY);
```
In case some implementation returns `undefined`? And also may be useful to access through `window` global? Just suggestions

andrevmatos (approved):
lgtm; just a small suggestion, but this seems fine, should fix e2e and also work on our practical tests

raiden-dapp/src/store/version-information/getters.ts
weilbith: According to the official W3C specification it will be always `null`:
https://html.spec.whatwg.org/multipage/webstorage.html#storage

weilbith: We use plain `localStorage` over `window.localStorage` at some other locations too. I'm very sorry, but at this late point I'm not motivated to change it only at this location just to then re-trigger the CI. I hope that is fine. Sorry.

andrevmatos: Yes, that's fine. Let's get this merged.

weilbith:
Important to mention: the end-to-end tests here are expected to succeed anyway. The issue of the linked PR is cause by the bumped version. But I tested it locally on top of `prepare_release`.

Issue #3053, opened 2022-02-16
dApp - adapt channel settling to new contracts for Arbitrum


weilbith:
## Description

There are two changes by the contracts for Arbitrum in regards of the settle timeout
  - the property has moved from individual channels to the token network
  - the value meaning has changed from block numbers to time in seconds
 
In result this affects the dApp in the following way:
  - the SDK does no more accept an individual settle timeout configuration value for new channels
  - the channel structure of the SDK does not include the settle timeout field anymore
  - the settle button of a closed channel has a countdown until it becomes settleable that needs to be adapted
  
## Acceptance criteria
- 

## Tasks
- [ ]

PR #3057, opened 2022-02-18
Feature/dapp adapt to arbitrum

### LINKED ISSUE #3051: dApp - adapt UDC withdrawal information to timestamps
## Description

We have a visual component that informs the user that a UDC withdrawal was planned and is waiting to become finally withdrawn. The included counter was based on block numbers. Due to this, the dApp is actually not even "compiling" anymore, as the new SDK version does not provide a block property for planned withdrawal events anymore.
First this needs to be adapted from a target block number to a target timestamp. The challenge is now how to adapt the visual part. Before, the new block event was causing the reactive update that result into a new "remaining block numbers" and thereby a countdown. As there is no such event for this use-case anymore as it is purely time based, we need to find an alternative.
One would be to use a countdown of the seconds. This is very similar to the old visuals, but not very nice. An alternative would be to render the actual date when the withdraw becomes possible.

## Acceptance criteria
- 

## Tasks
- [ ]

weilbith:

Fixes #3051 and #3053

**Short description**

I tackled both issues together, because only one of them would cause issues in the CI pipeline. Due to the underlying breaking changes of the SDK the dApp does not build properly without both changes together. As both issues are small enough, this should be fine.

Targets the `arbitrum` branch with the changes of the SDK.

**Definition of Done**

1. Build the dApp successfully
2. Close a channel and verify that the button changes to `settle` but is disabled and no block countdown
3. Plan a withdrawal form the UDC and verify that is displays the date when it becomes ready, verify after the date passes it shows green


weilbith:
I'll now rebase and force-push on top of the latest `arbitrum` changes.

weilbith:
Feedback to the failing CI:
  - the linting fails due to upstream issues in the `raiden-cli` which hasn't touched here (no fix)
  - both end-to-end test jobs are expected and will be resolved by merging #3064
  - the dApp unit tests fail due to a local time issue what needs to be fixed to run on any computer (I'll push a fix)

weilbith:
Alright, after a first failed attempt to fix the timing issue it works now. I'll integrate the fixup commit now, force-push and then we can merge soon. Please the above comment why some CI jobs are accepted to fail.

PR #3059, opened 2022-02-23
remove python client Bespin reference

weilbith:
First of all, Bespin is not the most recent version name for the Python client anymore. Nevertheless it is no more necessary to reference its implementation anymore here. Especially as the next major release of the contracts will be only supported by the Light Client.


taleldayekh (approved):
👍🏼

PR #3064, opened 2022-03-02
update e2e environment for new timestamp contracts

### LINKED ISSUE #3050: Update all components in the end-to-end Docker image in regards of the breaking contract changes
## Description

The contracts got updated with breaking changes to make them work on Roll-ups. This also requires all components like the services to adapt to it. So be able to test all workspaces of the LightClient, we need to update all references components in the end-to-end Docker image.
One problem here is that the breaking changes require to also get partner nodes that work with these new contracts. According to #3049 this will be the LC CLI. But the CLI will not be released before the SDK and everything got finally updated to work with the new contracts. To resolve this problem, it must be possible to get the local development version into the image when necessary.

## Acceptance criteria
- all referenced components link to versions compatible with the new contracts
- the image can be build with local components

## Tasks
- [ ]

weilbith:
Fixes #3050

Note that all versions point to custom release-candiate git tags that I created on all the components repositories. Especially the Light Client itself points to the `arbitrum` branch. We will need a final update of the image, pointing to official releases.
Anyway I have allowed myself to push [a release-candidate](https://hub.docker.com/layers/raidennetwork/lightclient-e2e-environment/v2.0.0rc1/images/sha256-9ac4eb107a07f8250d6d5ab4cd9dc1a29b47b132241536e7e62f00a565cf0a9c?context=explore) of the new image for convenience to use for #3034 (note that this requires to touch `e2e-environment/shared-script.sh#8`).


weilbith:
I also pushed a pre-release version of the new Docker image to the hub and referenced it in the CircleCI configuration. Let's see if we can reproduce the success from my local machine here.

weilbith:
I'll now rebase on latest `arbitrum`. The result should be a single commit.

weilbith:
The `lint` job is still expected to fail due to issues in the upstream branch. This will not be fixed here. All other jobs should pass here.

weilbith:
The `lint` job is expected to fail right now. A fix for this would be totally unrelated to the changes here. Therefore we will do this in the `arbitrum` upstream branch.

e2e-environment/Dockerfile
```diff
@@ -1,9 +1,9 @@
-ARG RAIDEN_VERSION="v2.1.0"
-ARG CONTRACTS_PACKAGE_VERSION="v0.40.0"
-ARG CONTRACTS_VERSION="0.40.0"
-ARG SERVICES_VERSION="v1.0.0"
+ARG RAIDEN_VERSION="arbitrum"
+ARG CONTRACTS_PACKAGE_VERSION="v1.0.0rc2"
```
netcriptus, line 2: There's an RC3 now, maybe it's a good idea to update

e2e-environment/Dockerfile
```diff
@@ -32,7 +35,8 @@ ARG CONTRACTS_VENV
 ARG CONTRACTS_PACKAGE_VERSION
 
 RUN python -m venv ${CONTRACTS_VENV}
-RUN . ${CONTRACTS_VENV}/bin/activate && pip install -U pip wheel && pip install raiden-contracts==${CONTRACTS_PACKAGE_VERSION}
+RUN ${CONTRACTS_VENV}/bin/pip install --upgrade pip wheel
+RUN ${CONTRACTS_VENV}/bin/pip install git+https://github.com/weilbith/raiden-contracts@${CONTRACTS_PACKAGE_VERSION}
```
netcriptus, line 39: Why are we getting it from your fork instead of the main repository?

netcriptus (commented):
Looking very good, I just need some clarification for one comment, and we might want to consider the new RC

e2e-environment/Dockerfile
weilbith: Is there? What does it include? The second release candidate includes all relevant changes that are not release procedure related, right?
Note that this will change anyway soon...

e2e-environment/Dockerfile
weilbith: At this point in time I did not want to spam the `origin` repos with such experimental tags. And as it will not stay like that anyway and will change, I thought that is fine.
Do you want me to change it? 🙃

netcriptus: As an effect, we'll have an `origin` repo pointing to a fork, which I honestly believe is worse. I'd prefer to have this pointing to a branch, but on a main repo.

e2e-environment/Dockerfile
netcriptus: It includes the version bump, but it's fine to move on with this one then.

e2e-environment/Dockerfile
weilbith: Sure. I'll do.

weilbith: ✔️

e2e-environment/Dockerfile
weilbith: ✔️

weilbith:
I addressed the review feedback. I also pointed to the new `raiden-contracts` tag after raiden-network/raiden-contracts#1614 . Finally I rebased on `arbitrum` which should make the `lint` CI job succeed too.

netcriptus (approved):

weilbith:
I now pushed the new image to DockerHub. This should be pulled by CircleCI and make the pipeline succeed.

PR #3069, opened 2022-03-05
Arbitrum: chose matrix server from PFS

andrevmatos:
Fixes #

**Short description**
This is an enhancement to some situations seen while testing with Arbitrum. When not explicitly specified (e.g. through cli's `--matrix-server` command line parameter), transport on `auto` mode will download the list of servers from RSB's github ([this list for testnets](https://raw.githubusercontent.com/raiden-network/raiden-service-bundle/master/known_servers/known_servers-development-v1.2.0.json)), and there's a high probability of those servers being incompatible with the current network (this is the case with arbitrum-rinkeby test PFS, which runs with its own matrix server).

The motivation of this change is that PFSs are either passed explicitly or picked from the chosen contracts (through `ServicesRegistry`), and therefore should always have a compatible matrix server, exposed in `/api/v1/info` endpoint. Therefore, preferring the server from the PFS will mostly ensure it's compatible with the current network and PFS, and only as a fallback we download the list from the `matrixServerLookup` URL.

Tests included.
This doesn't work only for Arbitrum, could work also for `master`, but since this touches test code which was already changed on `arbitrum` (#3034), it was easier to target that base, and most new features are already going to it anyway.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/services/utils.ts
+++ b/raiden-ts/src/services/utils.ts
@@ -7,9 +7,22 @@ import * as t from 'io-ts';
 import memoize from 'lodash/memoize';
 import uniqBy from 'lodash/uniqBy';
 import type { Observable } from 'rxjs';
-import { defer, EMPTY, firstValueFrom, from } from 'rxjs';
+import { defer, EMPTY, firstValueFrom, from, of } from 'rxjs';
 import { fromFetch } from 'rxjs/fetch';
-import { catchError, map, mergeMap, toArray } from 'rxjs/operators';
+import {
+  catchError,
+  concatMap,
+  delay,
+  first,
+  last,
+  map,
+  mergeAll,
+  mergeMap,
+  takeUntil,
+  tap,
+  throwIfEmpty,
+  toArray,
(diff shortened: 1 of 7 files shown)


raiden-ts/src/services/utils.ts
```diff
@@ -351,3 +365,54 @@ export async function signIOU(signer: Signer, iou: IOU): Promise<Signed<IOU>> {
...
+  pfsByAction: pathFind.request['payload']['pfs'],
+  deps: RaidenEpicDeps,
+): Observable<PFS> {
+  const { log, config$, latest$, init$ } = deps;
+  return config$.pipe(
+    first(),
+    mergeMap(({ pfsMode, additionalServices }) => {
+      if (pfsByAction) return of(pfsByAction);
+      else if (pfsMode === PfsMode.onlyAdditional) {
+        let firstError: Error;
```
weilbith, line 389: Thanks for using full variable names without abbreviations. 🤗

weilbith (approved):
Looks very useful. 👏🏾

PR #3070, opened 2022-03-07
Arbitrum: bump contracts to v1.0.0rc4 and remove incompatible deployments

andrevmatos:
Fixes # 

**Short description**
Title

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-ts/src/helpers.ts
+++ b/raiden-ts/src/helpers.ts
@@ -55,16 +55,8 @@ import {
   putRaidenState,
   replaceDatabase,
 } from './db/utils';
-import goerliDeploy from './deployment/deployment_goerli_unstable.json';
-import mainnetDeploy from './deployment/deployment_mainnet.json';
-import rinkebyDeploy from './deployment/deployment_rinkeby.json';
 import rinkebyArbitrumDeploy from './deployment/deployment_rinkeby-arbitrum.json';
-import ropstenDeploy from './deployment/deployment_ropsten.json';
-import goerliServicesDeploy from './deployment/deployment_services_goerli_unstable.json';
-import mainnetServicesDeploy from './deployment/deployment_services_mainnet.json';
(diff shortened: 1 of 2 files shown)


netcriptus (approved):

weilbith (changes requested):
Why can't you just try to import them when required and catch errors... Can't we somehow bundle the whole `data` directory with the SDK without always knowing what is inside?
I mean this way we always have to open a PR when we add or remove networks. Sound not very smart.

andrevmatos:
@weilbith in TS, there's 2 ways to import a JSON:
1. importing at compile time (`import _ from 'file.json';`) embeds it in bundled outputs, avoiding the requirement for the files to be available at the correct paths at runtime; downside is that we need to be explicit about which files to import, and they always need to be available at compile time (so we have to drop the non-available networks here)
2. using something like `fs.readFile*` to read and parse the JSON at runtime; can support dynamic files, but make static embedding harder (e.g. bundled builds).

But either way, I already support both solutions here: I try to keep static imports for known networks (in this branch, only `arbitrum-rinkeby`), to allow the embedding, but fallback to `fs.readFile` and throw only if deployment files for the current network aren't found in the expected paths. This should allow even dropping the files for custom/private networks after build and they'll be picked, as an additional feature over passing them explicitly when instantiating the SDK or relying only on `--user-deposit-contract-address` discovery (slower on first launch).

weilbith (approved):
Sorry, I missed the `default` case as it was folded away.
To be honest I don't understand why there is any case except the `default`. Doesn't make too much sense to me. Just more code, two things that do kinda the same, more responsibility, more that can break... 🤷🏾

andrevmatos:
The `default` case doesn't work for `build:bundle` case, and may create problems on webpack and other web-bundle solutions (if they don't make the JSONs in deployment folder properly available e.g. on a webserver), that's the reason we keep it for the known networks. Let's keep it like that and assess if any problems arise. It's not that much code, and checked at compile-time.

PR #3079, opened 2022-03-21
fix too quick channel interactions for quick-pay

### LINKED ISSUE #3078: dApp - quick pay delay after blockchain interaction
## Description

With the compatibility for Arbitrum, we also got zero confirmation blocks. In the scenario player runs we have seen that it can sometimes lead to issues when interacting with a channel that was just created/modified. Due to the zero confirmation blocks, the active node who modified the channels state immediately continues without a break. In result it leads to an unsynced partner node who has not yet seen the new channel/new deposit and will deny/error on the sent message. To prevent this issue, the scenarios received a couple of short delays after such a blockchain interaction.
For the "normal" user flow with open a channel manually and then continue with a transfer, this is not an issue for the dApp. But for the quick-pay feature we also observed this problem. Due to the automated nature of this feature, it will also immediately continue to act on the channel that just got opened or deposited to. The same solution approach as described above should be good enough. On non Arbitrum network the extra short delay isn't much of a factor. And for Arbitrum it is necessary to prevent such unsynchronized issues too often where the quick-pay transfer will completely fail.

## Acceptance criteria
- 

## Tasks
- [ ]

weilbith:
Fixes #3078

**Short description**
Please checkout the issue and commit message for details.

**Definition of Done**

(note that this targets a "flaky" issue)
 
1. In theory: do a couple of quick-pays which require to open a new channel and/or make a channel deposit
2. Verify that even after many of them it never failed...

### DIFF
--- a/raiden-dapp/src/components/channels/ChannelDepositAndTransferAction.vue
+++ b/raiden-dapp/src/components/channels/ChannelDepositAndTransferAction.vue
@@ -56,6 +56,9 @@ export default class ChannelDepositAndTransferAction extends Mixins(ActionMixin)
 
     await this.$raiden.deposit(tokenAddress, partnerAddress, depositTokenAmount);
 
+    // Sleep for short while to let the partner node see the channel deposit event.
+    await new Promise((resolve) => setTimeout(resolve, 3000));
+
     this.depositStep.completed = true;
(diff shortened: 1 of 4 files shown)


andrevmatos (approved):
Lgtm

PR #3091, opened 2022-03-25
dapp: adapt update message to express the risk

weilbith:
This is just a "hot fix". That means that the next update will include a breaking change in terms of new contracts. Users should not simply update without withdrawing their funds first. In future this will be replaced with a proper feature that is able to determine of which kind an update is to alternate the update message.


weilbith:
This change has been done together with @DinaBB 

There is no need for a CHANGELOG entry here.

taleldayekh (approved):
Thank you 😃

PR #3101, opened 2022-04-13
Backport deps updates, refactorings and fixes from master

### LINKED ISSUE #3100: LC uses online valid registered PFSs for presences even if they're operating on another network, causing offline presence responses
## Steps to Reproduce
1. stable/v2 SDK with PFS in `auto` mode (e.g. hosted instance at https://lightclient.raiden.network)
2. Go to goerli (which uses v0.40 `goerli_unstable` contracts)
3. It'll pick up valid (on registry deposit) PFSs, and when trying to check availability of a peer, use the first registered and online one, even if it's not operating on the same `TokenNetworkRegistry`
4. Any registered PFS  is offline or not replying properly, it's bypassed; but if the PFS is in fact online, it'll be used for presence; if it happens that this PFS is registered, online but working on a different network, it'll always reply that peers are offline

Notice that this error is caused by infrastructure misconfiguration, but the client must do better to identify this case and find valid services nonetheless.

## Expected Result
- Only a PFS which is registered, valid, online and operating on the same TNR as the node is picked for requests (including presence and route)

## Actual Result
- PFSs which are registered, valid and online, but not on the same TNR, may be picked up, and will prevent the node from getting consistent responses, instead getting always-offline-peer replies or no route found;

Current `stable` (latest `v2.1.0`) and `master` (arbitrum development branch) are affected

andrevmatos:
Fixes #3100 

**Short description**
In order to keep `stable` aka `v2` maintained, we need to backport the updates and refactorings (code quality) from the work done for Arbitrum into the stable branch. This includes the fix for the linked bug, which is also currently affecting Goerli. Keeping dependencies up to date is also important for security, in order to ensure we're not hit by fixed security flaws.

This doesn't pull the schema and contracts changes, which remain in `master` for arbitrum only, but keep the rest of the codebase somehow closer to master in order to facilitate any later bugfix and updates.

Most of these commits were already reviewed as part of Arbitrum work (sans conflict fixes). Otherwise, just the last 6 commits should require independent review.

**Definition of Done**

1.
2.


weilbith:
To merge this, we need to locally serve the current `gh-pages` version, install it, create some state etc. Then deploy locally this version and see if it works.

andrevmatos:
I've tested locally with old state (channels and udc deposits) in goerli, and this version worked fine; performed some direct transfers, everything working as expected.

weilbith (approved):

weilbith (approved):

PR #3110, opened 2022-04-28
Fix subpaths version loading and deployment

andrevmatos:
**Short description**
Ensure dApp instances deployed to subpaths (e.g. `/v2` and `/staging`) don't get deleted when a root instance is deployed (on tag), and also ensure those subpaths instances have service-worker assistant disabled so it doesn't affect the main cached version.

**Definition of Done**

1.
2.

### DIFF
--- a/raiden-dapp/.env.v2
+++ b/raiden-dapp/.env.v2
@@ -1,0 +1,6 @@
+VUE_APP_PUBLIC_PATH=/v2/
+VUE_APP_ALLOW_MAINNET=true
+VUE_APP_SERVICE_WORKER_DISABLED=true
+VUE_APP_IMPRINT=https://raiden.network/privacy.html
+VUE_APP_TERMS=https://github.com/raiden-network/light-client/blob/master/TERMS.md
+VUE_APP_CONFIGURATION_URL=${VUE_APP_PUBLIC_PATH}config.production.json
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,7 +119,7 @@ jobs:
       mode:
         description: Mode to configure the environment of the dApp build
         type: enum
-        enum: [production, staging, development, e2e, raiden-package]
+        enum: [production, staging, development, e2e, raiden-package, v2]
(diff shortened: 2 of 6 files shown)


raiden-dapp/.env.v2
```diff
@@ -0,0 +1,6 @@
+VUE_APP_PUBLIC_PATH=/v2/
+VUE_APP_ALLOW_MAINNET=true
+VUE_APP_SERVICE_WORKER_DISABLED=true
+VUE_APP_IMPRINT=https://raiden.network/privacy.html
+VUE_APP_TERMS=https://github.com/raiden-network/light-client/blob/master/TERMS.md
+VUE_APP_CONFIGURATION_URL=config.production.json
```
weilbith, line 6: That one need to point to `/v2/config.production.json` I think. I did that [here](https://github.com/raiden-network/light-client/blob/436a9530451fed1ddb816b3f2b5a30aee76a0af9/raiden-dapp/.env.production#L2) and I'm pretty certain that was necessary because we make a raw `fetch` call [here](https://github.com/raiden-network/light-client/blob/e8d0f259d6c20763e43186ef8ce49d81c6ef2d25/raiden-dapp/src/services/config-provider.ts#L35). So we should not load the configuration of Arbitrum One.

weilbith (changes requested):
Looks very good. I think you made some good alternative changes than I did in my commit. 
There is one "mistake" which definitely needs to be fixed. If I'm not completely off.

raiden-dapp/.env.v2
andrevmatos: This is a relative path. Shouldn't it download from BASE_URL?

andrevmatos: I tested here by hand, and indeed, both `config.production.json` and `./config.production.json` download from root.
I fixed it now.

weilbith: 👍🏾

weilbith (approved):
✔️

PR #3124, opened 2022-05-10
CLI endpoints and parameters to manage database backups

### LINKED ISSUE #3122: Provide state download endpoint and upload parameter for CLI
## Description
Allows users/scripts to backup internal CLI state in JSON format, actually the same used by dApp's state download functionality, and possibly relaod/upload it in another instance.

## Acceptance criteria
- User can download LC state as JSON from endpoint in CLI
- User can upload JSON with a command line parameter and file
- CLI doesn't shutdown while downloading, but (of course) only accepts upload when starting

## Tasks
- [ ]

andrevmatos:
Fixes #3122 

**Short description**
This adds the `/api/v1/state.json` endpoint to allow downloading state dump from a CLI node, and a new `--load-state <path.json>` parameter, to upload it back in a new instance / database path.
Additionally, we improve UX by not requiring SDK to be shut down for `Raiden.dumpDatabase`, and now it's also incremental by returning an `async` generator of rows (more useful on very big databases), instead of forcing collecting everything in an array beforehand.

**Definition of Done**

1. Lauch cli
2. `curl http://localhost:5001/api/v1/state.json -o cli_state_dump.json`
3. Stop cli
4. `rm -rv ./storage/raiden_*_<your_address>_*` to clear state
5. Launch cli with `--load-state cli_state_dump.json` and see it continue from state before

### DIFF
--- a/raiden-cli/src/routes/api.v1.ts
+++ b/raiden-cli/src/routes/api.v1.ts
@@ -84,5 +84,16 @@ export function makeApiV1Router(this: Cli): Router {
     response.json(RaidenConfig.encode(this.raiden.config));
   });
 
+  router.get('/state.json', async (_request: Request, response: Response) => {
+    response.writeHead(200, { 'Content-Type': 'application/json' });
+    response.write('[');
+    let count = 0;
+    for await (const row of this.raiden.dumpDatabase()) {
+      response.write((count++ ? ',\n' : '\n') + JSON.stringify(row));
+    }
+    response.write('\n]');
+    response.end();
+  });
+
   return router;
 }
--- a/raiden-dapp/src/components/account/backup-state/DownloadStateDialog.vue
+++ b/raiden-dapp/src/components/account/backup-state/DownloadStateDialog.vue
@@ -52,12 +52,17 @@ export default class DownloadStateDialog extends Mixins(NavigationMixin) {
 
     downloadLink.click();
     this.revokeDownloadURL(stateFileURL, downloadLink);
+    this.visible = false;
   }
 
   /* istanbul ignore next */
   async getStateFileURL(filename: string): Promise<string> {
-    const state = await this.$raiden.stopAndGetDatabaseDump();
-    const stateJSON = JSON.stringify(state);
+    let stateJSON = '[';
+    let count = 0;
+    for await (const row of this.$raiden.getDatabaseDump()) {
+      stateJSON += (count++ ? ',\n' : '\n') + JSON.stringify(row);
+    }
+    stateJSON += '\n]';
     const file = new File([stateJSON], filename, { type: 'application/json' });
     return URL.createObjectURL(file);
   }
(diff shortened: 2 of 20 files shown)


raiden-cli/src/routes/api.v1.ts
```diff
@@ -84,5 +84,16 @@ export function makeApiV1Router(this: Cli): Router {
     response.json(RaidenConfig.encode(this.raiden.config));
   });
 
+  router.get('/state.json', async (_request: Request, response: Response) => {
+    response.writeHead(200, { 'Content-Type': 'application/json' });
+    response.write('[');
+    let count = 0;
+    for await (const row of this.raiden.dumpDatabase()) {
+      response.write((count++ ? ',\n' : '\n') + JSON.stringify(row));
```
netcriptus, line 92: It looks like we're trying to do manually what `JSON.stringify()` would do for us. Can't we just load the result of `dumpDataBase()` in a variable (assuming it returns a list) and then just use `stringify` once?

raiden-dapp/src/components/account/backup-state/DownloadStateDialog.vue
```diff
@@ -52,12 +52,17 @@ export default class DownloadStateDialog extends Mixins(NavigationMixin) {
...
   }
 
   /* istanbul ignore next */
   async getStateFileURL(filename: string): Promise<string> {
-    const state = await this.$raiden.stopAndGetDatabaseDump();
-    const stateJSON = JSON.stringify(state);
+    let stateJSON = '[';
+    let count = 0;
+    for await (const row of this.$raiden.getDatabaseDump()) {
+      stateJSON += (count++ ? ',\n' : '\n') + JSON.stringify(row);
```
netcriptus, line 63: Same case as the comment above. Given that this is used more than once, even if delegating everything to `stringify` won't work, it maybe interesting to create a helper function to avoid the code repetition.

netcriptus (changes requested):

raiden-cli/src/routes/api.v1.ts
andrevmatos: Using `stringify` on the whole thing will store everything in memory. The idea here is exactly to _stream_ the contents (which can grow very big) one row at a time, iterating through the `AsyncIterable` and writing each item. This is specially useful here in cli, as its database is more probable to grow indefinitely.

raiden-dapp/src/components/account/backup-state/DownloadStateDialog.vue
andrevmatos: Unfortunately, I didn't find a simple solution to stream a programatically generated  `File` to a URL without collecting it in a string first, but at least the dApp isn't expected to contain that much entriesnas the cli. It's not very useful to try to split and reuse this code with cli, since this one collects in a string (`+=`) and the other stream to a writable object (`response.write`)

netcriptus:
Apart from the failing Dapp E2E tests, the changes look ok. I'll approve it once the build is green 👍

netcriptus (approved):
Approving for the code changes, just missing the green e2e tests. It's ok to merge as soon as that's fixed :)

PR #3132, opened 2022-05-31
WebUI

### LINKED ISSUE #3123: Add missing CLI endpoints for webUI, download and serve it from CLI web --web-ui parameter
## Description
LC CLI provides almost all endpoints from PC in a schema-compatible manner. Only a few config endpoints missing, which after implemented, should allow CLI to be a drop-in replacement for [Raiden webUI](https://github.com/raiden-network/webui), a nice user interface and admin panel for a raiden REST API-compatible client. Additionally, `--web-ui`/`--no-web-ui` command line options should control serving a pre-built version of the webUI in the respective endpoints.

This should give CLI users a better perspective of current node's state in a more ergonomic way than raw REST requests.

## Acceptance criteria
- Users can use the webUI with the CLI
- There's an easy way to serve the webUI from a CLI node

## Tasks
- [ ] Identify [possibly undocumented] endpoints needed by webUI
- [ ] Implement those endpoints with compatible schema
- [ ] Add build step/script to download and uncompress webUI
- [ ] Implement `--web-ui` to allow serving webUI files from `/`

andrevmatos:
Fixes #3123 

**Short description**
Adds [Raiden WebUI](https://github.com/raiden-network/webui) to CLI :tada: 
For compliance with schemas, features, behavior and endpoints expected by WebUI, some adjusments were needed:
- on SDK, a new method, `getTransfers`, allows efficient filtering and pagination when fetching transfers from database
- `/payments` and `pending_transfers` now use this method to serve properly paginated results to webUI
- `/config.*.json` serves a dynamic JSON to be used by webUI
- `/rpc` proxies ETH JSON RPC calls to SDK's provider 
- webUI can be downloaded and put inside `build` folder by `yarn build:webui` command, to be run after `yarn build`
- a new CLI option `--web-ui` enables serving the webUI from the `/ui` endpoint.

**Definition of Done**

1. After building the CLI, build the webui with `yarn build:webui`
2. Run the CLI with `--web-ui` parameter
3. Visit http://localhost:5001/ to see the webUI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,6 +142,7 @@ jobs:
           command: |
             yarn build
             yarn build:bundle
+            yarn build:webui
       - run:
           name: Compress CLI and bundle
           command: |
--- a/raiden-cli/CHANGELOG.md
+++ b/raiden-cli/CHANGELOG.md
@@ -3,8 +3,10 @@
 ## [Unreleased]
 ### Added
 - [#3122] `/api/v1/state.json` endpoint to allow downloading/backing up and `--load-state <path.json>` parameter to upload/rehydrate state/database in a fresh instance
+- [#3123] **Experimental**: `--web-ui` option to serve [Raiden WebUI](https://github.com/raiden-network/webui) from `/ui` route; requires `yarn build:webui`
 
 [#3122]: https://github.com/raiden-network/light-client/issues/3122
+[#3123]: https://github.com/raiden-network/light-client/issues/3123
 
 ## [3.0.0] - 2022-05-02
 ### Removed
--- a/raiden-cli/package.json
+++ b/raiden-cli/package.json
@@ -27,6 +27,7 @@
     "lint": "eslint --max-warnings 0 --ext '.ts' .",
     "build": "tsc --skipLibCheck",
     "build:bundle": "TS_NODE_PROJECT=./ignored ncc build -C ./src/index.ts -o ./bundle",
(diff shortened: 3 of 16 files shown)


karlb (approved):
I got it working.

The web UI won't tell you that another node is offline when you try to transfer (it is just a generic 500 error message). I'm quite sure the error message when running against the python client was more helpful, so there probably is an easy way change the server's response for that case to improve the situation.

andrevmatos:
Hi, @karlb , thank you very much for the review.
1. The SDK doesn't handle much PFS's errors, and instead prefer to just pass them through to caller, as they usually contain more details about what happened. The CLI then is responsible for catching this common error and translating it to the user, which it's now doing as a E409, same as python rest api. It seems there's no special handling for this kind of error in the webUI, but the 409 with `details` added now should look better and more descriptive in the `notificationService`. It's done now.
2. This is intended, the LC doesn't sync the same way as PC (i.e. all events of interest in each block[range]), and instead each epic polls independently its contracts for on-chain events, so we can't really tell how many blocks are left to scan, only when it's finished. Hopefully, after first sync, the SDK starts almost instantly, so it shouldn't be noticeable to users often.

I'll consider the points addressed, and merge this PR. If something is still not good, we can improve any point of concern in a followup PR. Thank you very much for going through this, this is a great addition for the LC, and should also extend webUI's life, which is also a great piece of software (good job @manuelwedler).

PR #3150, opened 2022-07-11
Revert "deps: Update dependency inquirer to v9"

andrevmatos:
Reverts #3140
Inquirer doesn't work well yet on the CJS CLI output, so let's stick with v8 for now.


compojoom (approved):
Looks fine and I can confirm that I can now start raiden without having to build the bundle first.

